PR #1, opened 2020-01-21
Feature/d3 asim 1867

spyrostz:


d3a_api_client/redis.py
```diff
@@ -0,0 +1,183 @@
...
+        self.redis_db.publish(self._command_topics[Commands.LIST_BIDS], json.dumps(""))
+        return self.wait_and_consume_command_response(Commands.LIST_BIDS)
+
+    def _on_register(self, msg):
+        message = json.loads(msg["data"])
+        if 'available_publish_channels' not in message or \
+                'available_subscribe_channels' not in message:
+            raise RedisAPIException(f'Registration to the market {self.market_id} failed.')
+
+        logging.info(f"Client was registered to market: {message}")
```
hannesdiedrich: I suggest to filter the message for the market here or make the prefix more generic (AFAIU, message does not only contain the market name, right?)

hannesdiedrich: Ah, I see now that you have the self. market_id handy here.

d3a_api_client/redis.py
```diff
@@ -0,0 +1,183 @@
...
+        return {
+            Commands.OFFER: f'{self._channel_prefix}/offer',
+            Commands.BID: f'{self._channel_prefix}/bid',
+            Commands.DELETE_OFFER: f'{self._channel_prefix}/delete_offer',
+            Commands.DELETE_BID: f'{self._channel_prefix}/delete_bid',
+            Commands.LIST_OFFERS: f'{self._channel_prefix}/offers',
+            Commands.LIST_BIDS: f'{self._channel_prefix}/bids'
+        }
+
+    def wait_and_consume_command_response(self, command_type):
```
hannesdiedrich: This could also be a private function to show the user the difference between "usable functions" and private ones. Do you agree?

spyrostz: Indeed, converting it to a private method. Thanks!

spyrostz:
> Nice! Thanks! Only minor comments. LGTM
> I could think of a wrapper class or module that only exposes the for the user interesting functions such as delete_bid, offer_energy, list_offers etc.
> Alternatively, we can add a detailed readme (as Faizan suggested, bit IMO not part of this story, but a separate one)

What I was thinking after creating this PR, was maybe a well-documented ABC interface. We are going to need an interface for the HTTP client anyways, but once this is implemented, it is going to be evident which methods are useful from an end-user perspective.  Adding it here and pinging you again to rereview.

PR #2, opened 2020-01-30
D3ASIM-1907: Added new client for devices. Added test that exhibits h…

spyrostz:
…ow this client is used.

### DIFF
--- a/tests/auto_offer_bid_on_device.py
+++ b/tests/auto_offer_bid_on_device.py
@@ -1,0 +1,38 @@
+"""
+Test file for the device client. Depends on d3a test setup file strategy_tests.external_devices
+"""
+from time import sleep
+from d3a_api_client.redis import RedisDeviceClient
+
+
+class AutoOfferBidOnMarket(RedisDeviceClient):
+
+    def on_market_cycle(self, market_info):
+        """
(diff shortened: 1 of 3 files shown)


tests/auto_offer_bid_on_device.py
```diff
@@ -0,0 +1,38 @@
...
+        """
+        print(f"New market information {market_info}")
+        if "available_energy_kWh" in market_info and market_info["available_energy_kWh"] > 0.0:
+            offer = self.offer_energy(market_info["available_energy_kWh"], 0.1)
+            print(f"Offer placed on the new market: {offer}")
+        print(market_info["energy_requirement_kWh"])
+        if market_info["energy_requirement_kWh"] > 0.0:
+            bid = self.bid_energy(market_info["energy_requirement_kWh"], 100)
+            print(f"Bid placed on the new market: {bid}")
+        print(f"Aggregated device/market statistics: {self.list_stats()}")
```
faizan2590: I assume these prints are not intentional, please remove before merging.

spyrostz: These actually are intentional, but on second thought I can replace them with logging.debug before merging. Thanks

faizan2590: replacing with logs would be better

PR #3, opened 2020-02-05
Feature/tests 4

spyrostz:


integration_tests/test_load_connection.py
```diff
@@ -0,0 +1,55 @@
...
+                bid_info = json.loads(bid["bid"])
+                assert bid_info["price"] == 33 * market_info["energy_requirement_kWh"]
+                assert bid_info["energy"] == market_info["energy_requirement_kWh"]
+
+            stats = self.list_stats()
+            traded_slots = stats["market_stats"]["energy_trade_profile"]["bought_energy"]["load"]["accumulated"].values()
+            assert isclose(stats["device_stats"]["bills"]["bought"], sum(traded_slots))
+            assert all(t in [0.05, 0.0] for t in traded_slots)
+
+            if market_info["start_time"][-5:] == "23:45":
```
hannesdiedrich: This will be prettier once we use unixtime, too, right? 
Just a minor thought, and I should have mentioned that already when you implemented the market_info: How about using unixtime for the API already, maybe the researchers/Datascientiest would be happy about that, too.  Or what do you think?

integration_tests/steps/device.py
```diff
@@ -0,0 +1,57 @@
...
+
+@then('the external client does not report errors')
+def step_impl(context):
+    assert context.load.errors == 0
+
+
+@then('the energy bills of the load report the required energy was bought by the load')
+def step_impl(context):
+    print(context.load.latest_stats["device_stats"]["bills"])
+    print()
```
hannesdiedrich: These prints are probably remnant from testing right?

spyrostz: Will do, thanks!

integration_tests/test_load_connection.py
spyrostz: The purpose of this is to actually detect that the simulation has finished, because there is no such event in d3a. Therefore a timestamp would not help me determine if the simulation has finished or not. However I agree with your suggestion and I will add a separate field for the market timestamp.

integration_tests/steps/device.py
```diff
@@ -0,0 +1,57 @@
...
+    # Connects one client to the load device
+    context.load = AutoBidOnLoadDevice('load', autoregister=True, redis_url='redis://localhost:6379/')
+
+
+@when('the external client container is started with test_pv_connection')
+def step_impl(context):
+    # Wait for d3a to activate all areas
+    sleep(5)
+    # Connects one client to the load device
+    context.load = AutoOfferOnPVDevice('pv', autoregister=True, redis_url='redis://localhost:6379/')
```
faizan2590: May be offline, if you can show me where exactly is `the external client container is started` ?

spyrostz: The word container is indeed misleading here. I am removing this to change the step name to `the external client is started with test_pv_connection`.

spyrostz:
> A bit out of context but when I tried to do `fab sync` it complained that [zipp](https://github.com/gridsingularity/d3a-api-client/blob/master/requirements/base.txt#L28) version is different in `d3a-api-client` and in `d3a-interface`

Don't worry about that, it is the same incompatibility that we encounter in d3a too. Should work okay since only the minor zipp version is changing across repos

PR #9, opened 2020-02-20
Feature/d3 asim 1991

spyrostz:


integration_tests/steps/device.py
```diff
@@ -82,5 +83,5 @@ def step_impl(context):
 
 @then('the energy bills of the load report the required energy was bought by the load')
 def step_impl(context):
-    assert context.device.latest_stats["device_stats"]["bills"]["bought"] == (24 * 4 - 2) * 0.05
+    assert isclose(context.device.latest_stats["device_stats"]["bills"]["bought"], 22 * 0.2)
```
hannesdiedrich: Hmm, I reverted the slot length to 15 mins after our offline discussion. Which one should we chouse?

integration_tests/test_pv_connection.py
```diff
@@ -25,26 +25,26 @@ def on_market_cycle(self, market_info):
...
                 assert listed_offer["price"] == listed_offer["price"]
                 assert listed_offer["energy"] == listed_offer["energy"]
-                # Try to delete the bid
+                # Try to delete the offer
                 delete_resp = self.delete_offer(offer_info["id"])
                 assert delete_resp["deleted_offer"] == offer_info["id"]
-                # Validate that the bid was deleted from the market
-                empty_listing = self.list_bids()
+                # Validate that the offer was deleted from the market
+                empty_listing = self.list_offers()
```
hannesdiedrich: What? We didnt see this in multiple PRs? Thanks for fixing!

integration_tests/build_test_containers.sh
```diff
@@ -2,7 +2,8 @@
 
 if [[ "$(docker images -q d3a-tests 2> /dev/null)" == "" ]]; then
   cd tests/ && \
-  git clone https://github.com/gridsingularity/d3a.git && \
+  cp -rf ~/dev/d3a . && \
+#  git clone https://github.com/gridsingularity/d3a.git && \
```
faizan2590: I guess this needs to be reverted before merging?

integration_tests/test_pv_connection.py
```diff
@@ -25,26 +25,26 @@ def on_market_cycle(self, market_info):
...
+                # Place the offer with a price that will be acceptable for trading
                 offer = self.offer_energy(market_info["available_energy_kWh"], 10 * market_info["available_energy_kWh"])
                 offer_info = json.loads(offer["offer"])
                 assert offer_info["price"] == 10 * market_info["available_energy_kWh"]
                 assert offer_info["energy"] == market_info["available_energy_kWh"]
 
             stats = self.list_stats()
             traded_slots = stats["market_stats"]["energy_trade_profile"]["sold_energy"]["pv"]["accumulated"].values()
-            assert isclose(stats["device_stats"]["bills"]["bought"], sum(traded_slots))
+            assert isclose(stats["device_stats"]["bills"]["sold"], sum(traded_slots), rel_tol=0.002)
```
faizan2590: Do you think at some point we should also introduce global constant for `rel_tol` too as its value is different over all our repos. ?

integration_tests/steps/device.py
spyrostz: Well, since I tried these locally with 60mins and the tests passed, I am inclined to say to keep the 60 mins slot length, in order to reduce the runtime of the simulation a little.

integration_tests/test_pv_connection.py
spyrostz: I really have no clue

integration_tests/build_test_containers.sh
spyrostz: Reverting, thanks

integration_tests/test_pv_connection.py
spyrostz: Maybe, but this is not the case here. The reason why the tolerance has such a big value (0.002 is rather big IMO) is because the results that we read from d3a are in a 3 decimal point precision. Therefore, there might be a deviation of 0.001 due to rounding. That's why the tolerance has to be at least 0.002 in order for all the checks to pass.

PR #10, opened 2020-02-21
Feature/d3 asim 1979

hannesdiedrich:


integration_tests/steps/device.py
```diff
@@ -3,21 +3,22 @@
...
            '--net integtestnet gsyd3a/d3a:redis-staging')
 
 
 @given('d3a container is started using setup file {setup_file}')
 def step_impl(context, setup_file):
     sleep(3)
-    system(f'docker run -d --env REDIS_URL=redis://redis.container:6379/ --net integtestnet d3a-tests '
-           f'  -l DEBUG run -t 1s -s 60m --setup {setup_file}')
+    system(f'docker run -d --name d3a-tests --env REDIS_URL=redis://redis.container:6379/ --net integtestnet '
+           f' d3a-tests -l INFO run -t 1s -s 15m --setup {setup_file} --no-export')
```
faizan2590: better make slot_length as a variable, as some test as using 60m market_slot_length too

d3a_api_client/redis.py
```diff
@@ -9,7 +9,7 @@
 
 
 root_logger = logging.getLogger()
-root_logger.setLevel(logging.ERROR)
+root_logger.setLevel(logging.INFO)
```
spyrostz: I think it is better to revert it by default to ERROR, and to let the user decide if they want to use a more informative log level

d3a_api_client/redis_market.py
```diff
@@ -0,0 +1,52 @@
+import logging
```
spyrostz: As a general comment in this file, you could potentially inherit from RedisDeviceClient and override/replace methods in order to achieve the same result as here and to be able to reuse code at the same time. Despite that, not important. Feel free to skip it

d3a_api_client/redis.py
hannesdiedrich: Yes, you are right. forgot to change it back

integration_tests/build_test_containers.sh
```diff
@@ -1,11 +1,22 @@
 #!/usr/bin/env bash
 
-if [[ "$(docker images -q d3a-tests 2> /dev/null)" == "" ]]; then
+if [[ "$OSTYPE" == "linux-gnu" ]]; then
```
spyrostz: Hmm, I do not necessarily disagree here, but wouldn't this cause problems if we decide to run this on Travis ? (because it is a Linux OS but does not use sudo - it is usually disabled in our jobs)

integration_tests/build_test_containers.sh
```diff
@@ -1,11 +1,22 @@
...
+    docker_command="sudo docker"
+else
+    docker_command="docker"
+fi;
+
+D3A_IMAGE_TAG="d3a-tests"
+
+if [[ "$(${docker_command} images -q ${D3A_IMAGE_TAG} 2> /dev/null)" == "" ]]; then
+  echo "Building d3a image ..." && \
+  rm -rf tests/d3a && \
```
spyrostz: Why do you try to delete the cloned repo again ? Does not hurt of course, since rm -rf does not raise an error if the file/directory is not there, but still it could be avoided IMO

hannesdiedrich: hach, yes, you are right. Will revert it then. Is a remnant of me trying to make it run on my linux machine.

integration_tests/build_test_containers.sh
hannesdiedrich: hach, yes, you are right. Will revert it then. Is a remnant of me trying to make it run on my linux machine.

integration_tests/test_ess_bid_connection.py
```diff
@@ -21,6 +21,7 @@ def on_market_cycle(self, market_info):
         try:
             assert "energy_to_buy_dict" in market_info
             energy = market_info["energy_to_buy_dict"]
+            print("energy", market_info["energy_to_buy_dict"])
```
spyrostz: Just a reminder to remove before merge.

integration_tests/build_test_containers.sh
hannesdiedrich: If you are testing multiple times, the an Error (or rather a warning) is shown: "d3a folder already exists". Now that you are point it out, i dont know what git clone does with existing folders. So, i still vote for keeping it.

d3a_api_client/redis_market.py
hannesdiedrich: I definitely thought about it, but i dont want to expose functions like register and unregister to the RedisMarketClient. But i could override them with pass. What do you think?

spyrostz: That would be the way yes, to create methods that have an empty body (only a pass statement) in order to mask the base class implementation. Still, if you feel it is not worth it leave it as is

hannesdiedrich: I dont like the idea really, because i would have to skip the super().__init__() , too. 
So, omitting inheriting .

PR #11, opened 2020-02-26
D3ASIM-1943: Initial implementation of the api client that connects t…

spyrostz:
…hrough d3a-web. Still pending, SSL and extensive testing.

### DIFF
--- a/d3a_api_client/rest_device.py
+++ b/d3a_api_client/rest_device.py
@@ -74,7 +74,7 @@ class RestDeviceClient(APIClientInterface):
         return True
 
     def _get_request(self, endpoint_suffix):
-        resp = requests.post(
+        resp = requests.get(
             f"{self._url_prefix}/{endpoint_suffix}/",
             headers={"Content-Type": "application/json",
                      "Authorization": f"JWT {self.jwt_token}"})
--- a/d3a_api_client/rest_device.py
+++ b/d3a_api_client/rest_device.py
@@ -1,0 +1,164 @@
+from d3a_api_client import APIClientInterface
+import os
+import logging
+import requests
+import json
+import ssl
+from functools import wraps
+from concurrent.futures.thread import ThreadPoolExecutor
+from d3a_api_client.websocket_device import WebsocketMessageDispatcher, WebsocketThread
+
+
+root_logger = logging.getLogger()
+root_logger.setLevel(logging.INFO)
+
+
+def logging_decorator(command_name):
+    def decorator(f):
+        @wraps(f)
+        def wrapped(self, *args, **kwargs):
+            logging.debug(f'Sending command {command_name} to device.')
+            return_value = f(self, *args, **kwargs)
+            logging.debug(f'Command {command_name} responded with: {return_value}.')
+            return return_value
+        return wrapped
+    return decorator
+
+
+class RestDeviceClient(APIClientInterface):
+
+    def __init__(self, simulation_id, device_id, domain_name,
+                 websockets_domain_name, is_ssl, autoregister=False):
+        self.simulation_id = simulation_id
+        self.device_id = device_id
+        self.domain_name = domain_name
+        self.retrieve_key_from_server()
+
+        if is_ssl:
+            self.pem_cert = ssl.get_server_certificate(domain_name)
+        self.dispatcher = WebsocketMessageDispatcher(self)
+        self.websocket_thread = WebsocketThread(simulation_id, device_id, self.jwt_token,
+                                                websockets_domain_name, self.dispatcher)
+        self.websocket_thread.start()
+        self.callback_thread = ThreadPoolExecutor(max_workers=5)
+        self.registered = False
+        if autoregister:
+            print(f"Trying to register")
+            self.register()
+
+    def retrieve_key_from_server(self):
+        resp = requests.post(
+            f"{self.domain_name}/api-token-auth/",
+            data=json.dumps({"username": os.environ["API_CLIENT_USERNAME"],
+                             "password": os.environ["API_CLIENT_PASSWORD"]}),
+            headers={"Content-Type": "application/json"})
+        if resp.status_code != 200:
+            logging.error(f"Request for token authentication failed with status code {resp.status_code}."
+                          f"Response body: {resp.text}")
+            return
+        self.jwt_token = json.loads(resp.text)["token"]
+
+    @property
+    def _url_prefix(self):
+        return f'{self.domain_name}/external-connection/api/{self.simulation_id}/{self.device_id}/'
+
+    def post_request(self, endpoint_suffix, data):
+        resp = requests.post(
+            f"{self._url_prefix}/{endpoint_suffix}/",
+            data=json.dumps(data),
+            headers={"Content-Type": "application/json",
+                     "Authorization": f"JWT {self.jwt_token}"})
+        if resp.status_code != 200:
+            logging.error(f"Request {endpoint_suffix} failed with status code {resp.status_code}."
+                          f"Response body: {resp.text}")
+            return False
+        return True
+
+    def get_request(self, endpoint_suffix):
+        resp = requests.post(
+            f"{self._url_prefix}/{endpoint_suffix}/",
+            headers={"Content-Type": "application/json",
+                     "Authorization": f"JWT {self.jwt_token}"})
+        if resp.status_code != 200:
+            logging.error(f"Request {endpoint_suffix} failed with status code {resp.status_code}."
+                          f"Response body: {resp.text}")
+            return False
+        return True
+
+    @logging_decorator('register')
+    def register(self, is_blocking=True):
+        if self.post_request('register', {}):
+            return_value = self.dispatcher.wait_for_command_response('register')
+            self.registered = True
+            return return_value
+
+    @logging_decorator('unregister')
+    def unregister(self, is_blocking):
+        if self.post_request('unregister', {}):
+            return_value = self.dispatcher.wait_for_command_response('unregister')
+            self.registered = False
+            return return_value
+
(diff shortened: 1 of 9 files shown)


d3a_api_client/rest_device.py
```diff
@@ -0,0 +1,164 @@
...
+            logging.error(f"Request {endpoint_suffix} failed with status code {resp.status_code}."
+                          f"Response body: {resp.text}")
+            return False
+        return True
+
+    def get_request(self, endpoint_suffix):
+        resp = requests.post(
+            f"{self._url_prefix}/{endpoint_suffix}/",
+            headers={"Content-Type": "application/json",
+                     "Authorization": f"JWT {self.jwt_token}"})
```
hannesdiedrich: In the API, you are requesting a header that contains `Authorization` not  `authorization `. Is it case sensitive?

d3a_api_client/rest_device.py
```diff
@@ -0,0 +1,164 @@
...
+        if is_ssl:
+            self.pem_cert = ssl.get_server_certificate(domain_name)
+        self.dispatcher = WebsocketMessageDispatcher(self)
+        self.websocket_thread = WebsocketThread(simulation_id, device_id, self.jwt_token,
+                                                websockets_domain_name, self.dispatcher)
+        self.websocket_thread.start()
+        self.callback_thread = ThreadPoolExecutor(max_workers=5)
+        self.registered = False
+        if autoregister:
+            print(f"Trying to register")
```
hannesdiedrich: Please exchange with a logging message. Probably even in the function register().

hannesdiedrich: AH, no, i just see your loggoing decorator! Nice! Just delete this line then.

d3a_api_client/rest_device.py
```diff
@@ -0,0 +1,164 @@
...
+            logging.error(f"Request {endpoint_suffix} failed with status code {resp.status_code}."
+                          f"Response body: {resp.text}")
+            return False
+        return True
+
+    @logging_decorator('register')
+    def register(self, is_blocking=True):
+        if self.post_request('register', {}):
+            return_value = self.dispatcher.wait_for_command_response('register')
+            self.registered = True
```
hannesdiedrich: Cant the user can also be restricted to register? In what case self.registered should not be set to True, right?

d3a_api_client/websocket_device.py
```diff
@@ -0,0 +1,80 @@
...
+import traceback
+import websockets
+from d3a_interface.utils import wait_until_timeout_blocking
+
+
+class RestWebsocketAPIException(Exception):
+    pass
+
+
+class WebsocketMessageDispatcher:
```
hannesdiedrich: As discussed offline, the websocket in the API is used for sending data, and in my perception, on this end, it really receives the data, so IMO it should be renamed to WebsocketMessageReceiver. Or did i get it wrong?

d3a_api_client/websocket_device.py
```diff
@@ -0,0 +1,80 @@
...
+        response = next(c
+                        for c in self.command_response_buffer
+                        if "command" in c and c["command"] == command_name)
+        self.command_response_buffer.remove(response)
+        return response
+
+
+async def websocket_coroutine(websocket_uri, websocket_headers, message_dispatcher):
+    async with websockets.connect(websocket_uri, extra_headers=websocket_headers) as websocket:
+        while True:
```
hannesdiedrich: I wonder why you need a `while True` here shouldnt the async def and assync with deal with reiterating?

integration_tests/build_test_containers.sh
```diff
@@ -6,7 +6,8 @@ if [[ "$(docker images -q ${D3A_IMAGE_TAG} 2> /dev/null)" == "" ]]; then
   echo "Building d3a image ..." && \
   rm -rf tests/d3a && \
   cd tests/ && \
-  git clone https://github.com/gridsingularity/d3a.git && \
+  cp -rf ~/dev/d3a . && \
+#  git clone https://github.com/gridsingularity/d3a.git && \
```
hannesdiedrich: Please revert when merging (and thanks for the tip! No need for pushing the d3a changes to github anymore :-) )

d3a_api_client/websocket_device.py
```diff
@@ -0,0 +1,80 @@
...
+        super().__init__(*args, **kwargs)
+
+    def run(self):
+        event_loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(event_loop)
+        websockets_uri = f"{self.domain_name}/{self.sim_id}/{self.area_uuid}/"
+        event_loop.run_until_complete(
+            websocket_coroutine(websockets_uri, self.websocket_headers, self.message_dispatcher)
+        )
+        event_loop.run_forever()
```
hannesdiedrich: Only briefly looking into asyncio.run_forever() shows me, that you need to receive a stop() command. Is this the case when the user keyboard-interrups? 
Without this, the eventloop will also successfully be stopped when the websocket_coroutine future is finished, right?

d3a_api_client/rest_device.py
spyrostz: It is not case sensitive, but just in case it is sent in the future in mixed-case, I'll be adding a conversion to lowercase in the d3a-web PR. Thanks!

d3a_api_client/rest_device.py
spyrostz: Thanks, deleting that

d3a_api_client/rest_device.py
spyrostz: Indeed, I'll be checking the output of register command to fill this value, thanks

d3a_api_client/websocket_device.py
spyrostz: Indeed you are right, naming you suggest is much better. Fixing this

d3a_api_client/websocket_device.py
spyrostz: All `await` does is to wait asynchronously until the command returns. Apart from that, not async nor await have any iteration logic. What has iteration logic is the event loop, but cannot be used this way to my knowledge (by default the event loop is processing events that are market as async, and cannot force a function to be triggered multiple times)

integration_tests/build_test_containers.sh
spyrostz: Will do thanks!

d3a_api_client/websocket_device.py
spyrostz: You are right, this is a remnant from a past implementation. The `run_forever` call should not be needed here, therefore removing this. Thanks!

d3a_api_client/websocket_device.py
hannesdiedrich: ok, thanks for the explanation!

spyrostz: More on that on the workshop next week!

tests/test_auto_offer_on_device_rest.py
```diff
@@ -21,7 +21,7 @@ def on_market_cycle(self, market_info):
         logging.debug(f"New market information {market_info}")
         if "available_energy_kWh" in market_info["device_info"] and market_info["device_info"]["available_energy_kWh"] > 0.0:
             print(market_info["device_info"]["available_energy_kWh"])
```
hannesdiedrich: You can remove this print, too

d3a_api_client/websocket_device.py
hannesdiedrich: @spyrostz please remove this, havent seen this change in your latest commit. Thanks!

d3a_api_client/rest_device.py
```diff
@@ -0,0 +1,163 @@
...
+            logging.error(f"Request for token authentication failed with status code {resp.status_code}."
+                          f"Response body: {resp.text}")
+            return
+        self.jwt_token = json.loads(resp.text)["token"]
+
+    @property
+    def _url_prefix(self):
+        return f'{self.domain_name}/external-connection/api/{self.simulation_id}/{self.device_id}/'
+
+    def post_request(self, endpoint_suffix, data):
```
faizan2590: I think it could be changed to a private method?

spyrostz: Changing it, thanks

d3a_api_client/rest_device.py
```diff
@@ -0,0 +1,163 @@
+from d3a_api_client import APIClientInterface
+import os
+import logging
+import requests
```
faizan2590: Please add it in project requirements. Thanks!

spyrostz: It's already there https://github.com/gridsingularity/d3a-api-client/blob/feature/D3ASIM-1943/requirements/base.in#L4

faizan2590: strange, dont know why my PyCharm is still complaining

d3a_api_client/websocket_device.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+class WebsocketMessageReceiver:
+    def __init__(self, rest_client):
+        self.client = rest_client
+        self.command_response_buffer = []
+
+    def received_message(self, message):
+        if "event" in message:
+            if message["event"] == "market":
```
faizan2590: Shouldn't we introduce an Enum at this point as we will exactly have only three type of events ?

spyrostz: Well, this is a websocket message, and we receive it from the network connection, therefore in order to expose an Enum here it has to be populated in d3a. Therefore d3a-interface has to be involved in order to keep the enum definition and be reused across d3a and d3a-api-client. Not negative to introduce it, but not for now. Maybe after the deploy.

PR #14, opened 2020-03-04
D3ASIM-2068: Created market rest api client, to be able to connect to…

spyrostz:
… a market area.


d3a_api_client/rest_market.py
```diff
@@ -0,0 +1,39 @@
...
+        self.websocket_thread.start()
+        self.callback_thread = ThreadPoolExecutor(max_workers=5)
+
+    @property
+    def _url_prefix(self):
+        return f'{self.domain_name}/external-connection/api/{self.simulation_id}/{self.device_id}'
+
+    @logging_decorator('market_stats')
+    def list_stats(self):
+        if self._post_request('market_stats', {}):
```
hannesdiedrich: I think you have to incorporate the `_url_prefix` into the entpoint here, right?

spyrostz: Nice catch! Will do, thanks

d3a_api_client/utils.py
```diff
@@ -0,0 +1,42 @@
...
+def retrieve_jwt_key_from_server(domain_name):
+    resp = requests.post(
+        f"{domain_name}/api-token-auth/",
+        data=json.dumps({"username": os.environ["API_CLIENT_USERNAME"],
+                         "password": os.environ["API_CLIENT_PASSWORD"]}),
+        headers={"Content-Type": "application/json"})
+    if resp.status_code != 200:
+        logging.error(f"Request for token authentication failed with status code {resp.status_code}."
+                      f"Response body: {resp.text}")
+        return
```
faizan2590:
```suggestion
        return None
```
?

d3a_api_client/rest_market.py
```diff
@@ -0,0 +1,45 @@
...
+        self.simulation_id = simulation_id
+        self.device_id = device_id
+        self.domain_name = domain_name
+        self.jwt_token = retrieve_jwt_key_from_server(domain_name)
+
+        self.dispatcher = WebsocketMessageReceiver(self)
+        self.websocket_thread = WebsocketThread(simulation_id, device_id, self.jwt_token,
+                                                websockets_domain_name, self.dispatcher)
+        self.websocket_thread.start()
+        self.callback_thread = ThreadPoolExecutor(max_workers=5)
```
faizan2590: May be an overkill but would it makes sense to set max_workers via constants ?

d3a_api_client/utils.py
spyrostz: It is exactly the same as now actually, return value of a function that just returns is None

d3a_api_client/rest_market.py
spyrostz: Kind of, I'll do it though since you mentioned it

PR #15, opened 2020-03-09
Added utility function for retrieving the UUID of a device from the d…

spyrostz:
…evice name, mainly targeted for end-to-end integration tests.


d3a_api_client/utils.py
```diff
@@ -41,3 +41,31 @@ def get_request(endpoint, data, jwt_token):
...
+    headers = {'Authorization': f'JWT {jwt_key}', 'Content-Type': 'application/json'}
+
+    query = 'query { readConfiguration(uuid: "{' + collab_id + \
+            '}") { scenarioData { representation { serialized } } } }'
+
+    endpoint = HTTPEndpoint(url, headers)
+    data = endpoint(query=query)
+    return get_area_uuid_from_area_name(
+        json.loads(data["data"]["readConfiguration"]["scenarioData"]["representation"]["serialized"]), area_name
+    )
```
faizan2590: It could be the case where no area_uuid is catched with provided `area_name` parameter, if Im right then an additional error message should be thrown IMO ?

spyrostz: ATM None is returned from the function at the case you mention. That way the user of the function knows that no area was found implicitly, by checking the return value and finding out that it is None. Therefore functionally it works as it is now. However, if you prefer an exception instead of the None return value please write it here and I'll adapt it.

faizan2590: The reason Im insisting it a bit more is since it is client end that may need to know to debug what went wrong. Even a small message like `No area exist with this name inside Collaboration.` would be nice to have.

spyrostz: Persuaded. Adding the exception, thanks!

PR #17, opened 2020-03-12
Feature/d3 asim 2084

hannesdiedrich:


integration_tests/test_pv_connection.py
```diff
@@ -36,7 +36,7 @@ def on_market_cycle(self, market_info):
                 assert listed_offer["energy"] == listed_offer["energy"]
                 # Try to delete the offer
                 delete_resp = self.delete_offer(offer_info["id"])
-                assert delete_resp["deleted_offer"] == offer_info["id"]
+                assert delete_resp["deleted_offers"] == [offer_info["id"]]
```
faizan2590: Very minor but its better to follow a uniform naming convention, for instance here it is `deleted_offers`  BUT for bids [here](https://github.com/gridsingularity/d3a-api-client/pull/17/files#diff-b154c1ac9b2fffcdf06e03c6c5944d7aR37) it is `bids_deleted`

integration_tests/test_load_connection.py
```diff
@@ -34,7 +34,7 @@ def on_market_cycle(self, market_info):
                 assert listed_bid["energy"] == bid_info["energy"]
                 # Try to delete the bid
                 delete_resp = self.delete_bid(bid_info["id"])
-                assert delete_resp["bid_deleted"] == bid_info["id"]
+                assert delete_resp["bids_deleted"] == [bid_info["id"]]
```
spyrostz: I do not have an objection here, but I think it would be nice for Colin to know about this change, so that he adapts his test scripts to this change.

integration_tests/test_pv_connection.py
spyrostz: Same here, Colin should be notified for any change in the endpoint response structure

integration_tests/test_load_connection.py
hannesdiedrich: ok! Will do, especially after incorporating @faizan2590 request, which i agree on.

PR #22, opened 2020-03-19
Feature/d3 asim 2115

hannesdiedrich:


d3a_api_client/redis_device.py
```diff
@@ -11,11 +11,15 @@ def __init__(self, device_id, autoregister=True, redis_url='redis://localhost:63
 
     def _on_register(self, msg):
         message = json.loads(msg["data"])
+        if not self._is_trans_id_matching(message):
+            return
```
spyrostz: I think it is better if you raise an exception here. That way the user will be notified that the registration failed, and the client will exit.

d3a_api_client/redis_device.py
```diff
@@ -11,11 +11,15 @@ def __init__(self, device_id, autoregister=True, redis_url='redis://localhost:63
...
         logging.info(f"Client was registered to the device: {message}")
         self.is_active = True
 
     def _on_unregister(self, msg):
         message = json.loads(msg["data"])
+        if not self._is_trans_id_matching(message):
```
spyrostz: Same here

d3a_api_client/redis_device.py
hannesdiedrich: You are right! There is only ONE register request in this channel. Will add it.

hannesdiedrich: But this applies also to the rest client, right? Will add it there too.

spyrostz: Applies to the rest client indeed. Thanks

d3a_api_client/redis.py
```diff
@@ -264,6 +261,13 @@ def executor_function():
             self.on_finish(message)
         self.executor.submit(executor_function)
 
+    def _check_buffer_message_matching_command_and_id(self, message):
+        transaction_id = message["transaction_id"]
+        if not any(command == "register" and "transaction_id" in data and
+                   data["transaction_id"] == transaction_id
+                   for command, data in self._blocking_command_responses.items()):
+            raise Exception("There is no matching command response in _blocking_command_responses.")
```
faizan2590:
```suggestion
            raise RedisAPIException("There is no matching command response in _blocking_command_responses.")
```

d3a_api_client/redis.py
```diff
@@ -9,7 +10,7 @@
 from d3a_api_client.constants import MAX_WORKER_THREADS
 
 root_logger = logging.getLogger()
-root_logger.setLevel(logging.ERROR)
+root_logger.setLevel(logging.INFO)
```
faizan2590: Any particular reason to change the log level?

d3a_api_client/redis.py
```diff
@@ -264,6 +261,13 @@ def executor_function():
             self.on_finish(message)
         self.executor.submit(executor_function)
 
+    def _check_buffer_message_matching_command_and_id(self, message):
+        transaction_id = message["transaction_id"]
```
faizan2590: Can you also check and raise an appropriate exception if `transaction_id` is None or if `transaction_id` key not part of message ?

hannesdiedrich: yes! Good idea! Thanks!

d3a_api_client/redis.py
hannesdiedrich: No, this is a remnant of testing. Will change it back.

PR #30, opened 2020-04-20
D3ASIM-1956: Adding implementation to post offer/bid with desired rate.

faizan2590:


faizan2590:
> It still surpasses my understanding why this is needed, but LGTM

@hannesdiedrich Well as far as I have understood from the description of the ticket that @candrews42  wrote is it is somehow a matter of convenience to the users to place offers/bids based on `rate` rather than `price` and the client on its own takes care of `rate to price conversion`

PR #38, opened 2020-06-12
TESTS-24: add researcher_id flag in order to allow multiple researche…

leajski:
…r to connect to different devices

### DIFF
--- a/tests/test_auto_offer_on_device_rest.py
+++ b/tests/test_auto_offer_on_device_rest.py
@@ -3,10 +3,24 @@ Test file for the device client. Depends on d3a test setup file strategy_tests.e
 """
 import logging
 import sys
+import argparse
+import numpy as np
 from time import sleep
+from random import randint
 from d3a_api_client.rest_device import RestDeviceClient
 from d3a_api_client.utils import get_area_uuid_from_area_name_and_collaboration_id
 
+#parse arguments
+parser = argparse.ArgumentParser()
+parser.add_argument("--simulation_id", type=str, help="Simulation uuid")
+parser.add_argument("--load", type=str, help="Load device")
+parser.add_argument("--pv", type=str, help="PV device")
+parser.add_argument("--researcher_id", type=float, help="Researcher id")
+args = parser.parse_args()
+
+number_of_areas = 10
+number_of_researcher = 2
+connected_devices = []
 
 class AutoOfferBidOnMarket(RestDeviceClient):
 
@@ -24,9 +38,10 @@ class AutoOfferBidOnMarket(RestDeviceClient):
         """
         if self.registered is False or self.is_finished is True:
             return
+        #sleep(randint(1,10)/100.)
         logging.debug(f"New market information {market_info}")
         if "available_energy_kWh" in market_info["device_info"] and market_info["device_info"]["available_energy_kWh"] > 0.0:
-            offer = self.offer_energy_rate(market_info["device_info"]["available_energy_kWh"], 16)
+            offer = self.offer_energy_rate(market_info["device_info"]["available_energy_kWh"] / 2, 16)
             logging.debug(f"Offer placed on the new market: {offer}")
             assert len(self.list_offers()) == 1
         if "energy_requirement_kWh" in market_info["device_info"] and market_info["device_info"]["energy_requirement_kWh"] > 0.0:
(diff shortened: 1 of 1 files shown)


tests/test_auto_offer_on_device_rest.py
```diff
@@ -43,24 +58,33 @@ def on_trade(self, trade_info):
...
+#the list contains lists of the devices numbers each researcher will connect to
+areas_split = [i.tolist() for i in np.array_split(np.array(areas),number_of_researcher)]
+
+if args.researcher_id <= number_of_researcher:
+    #create a list with the names of the Load the researcher connects to 
+    load_names = [args.load if i == 0 else f'{args.load} {i+1}' for i in areas_split[int(args.researcher_id-1)]]
+    #create a list with the names of the PV the researcher connects to 
+    pv_names = [args.pv if i == 0 else f'{args.pv} {i+1}' for i in areas_split[int(args.researcher_id-1)]]
+else:
+    raise ValueError(f'The researcher id is out of range. The number of researcher is {number_of_researcher}.')
```
hannesdiedrich, line 73: I think i understand what you are trying to do here but IMO this could be done simpler.
How about not having consecutive area ids/names but just add the researcher id in the area_name
for example something like this:
`pv_names = [f"area_{researcher_id}_{area_number}" for researcher_id in range(len(number_of_researcher)) for area_number in number_of_areas ]`
Or did i get it wrong?
It would be good if you could omit using numpy because it is a large library, takes long time to install (needs to be done every time when the integrationtests are running on travis). Especially it you only need it for list/array mutations.
Thanks!

tests/test_auto_offer_on_device_rest.py
```diff
@@ -24,9 +38,10 @@ def on_market_cycle(self, market_info):
         """
         if self.registered is False or self.is_finished is True:
             return
+        #sleep(randint(1,10)/100.)
```
hannesdiedrich, line 41: If you dont use it, please remove this line and also the import at the top.

tests/test_auto_offer_on_device_rest.py
```diff
@@ -3,10 +3,24 @@
...
 import sys
+import argparse
+import numpy as np
 from time import sleep
+from random import randint
 from d3a_api_client.rest_device import RestDeviceClient
 from d3a_api_client.utils import get_area_uuid_from_area_name_and_collaboration_id
 
+#parse arguments
+parser = argparse.ArgumentParser()
```
hannesdiedrich, line 14: Thanks for this!

tests/test_auto_offer_on_device_rest.py
```diff
@@ -3,10 +3,24 @@
...
+parser.add_argument("--simulation_id", type=str, help="Simulation uuid")
+parser.add_argument("--load", type=str, help="Load device")
```
hannesdiedrich, line 16: Please be more descriptive here. It is really a Load ID,  right?

tests/test_auto_offer_on_device_rest.py
leajski: I actually split it that way because the API gets the device uuid with the device name. So the name refers to the one given by default when I create the area on the JS test: 'Load, Load 2 ...'. If I use the researcher_id in the names, it implies I have to rename in the test all devices when they are created (correct me if I'm wrong).
Therefore, I don't know what it is better: split in the API script based on the default name, or rename one by one the devices in the test ?

leajski: Also, does the number of researcher should be "hard-coded" ? For example, if we decide to have the test running with always 2 researchers then I don't need numpy and I can directly attribute the first 5 devices names to the first researcher and the 5 next to the other (or rename the devices with researcher_id 1 or 2 in the collaboration). 
I used numpy in order to split the devices accordingly to the number of researcher, I am wondering is this necessary or not ? (I also guess I could do it without numpy)

tests/test_auto_offer_on_device_rest.py
leajski: Sure, I will do it

tests/test_auto_offer_on_device_rest.py
leajski: Actually the argument that needs to be passed is 'Load' or 'PV'. Maybe a boolean flag would be better ? Or considering here that the devices are always Load and PV, should it be "hard-coded" ?

tests/test_auto_offer_on_device_rest.py
```diff
@@ -43,24 +53,23 @@ def on_trade(self, trade_info):
     def on_finish(self, finish_info):
         self.is_finished = True
 
-number_of_areas = 10
-load_names = [str(sys.argv[2]) if i == 0 else str(sys.argv[2])+' '+str(i+1) for i in range(number_of_areas)]
-pv_names = [str(sys.argv[3]) if i == 0 else str(sys.argv[3])+' '+str(i+1) for i in range(number_of_areas)]
-connected_devices = []
-
-for i in range(number_of_areas):
+for i in range(len(args.load_names)):
```
hannesdiedrich, line 56: How about this simpler loop over all members of  the load_names list
```suggestion
for load_name in args.load_names:
```
In general, one-latter variables (here `i`)  are not very helpful for the code readability. So please try to gice verbose names. If you really need this as a counting variable, you could have named it as `load_number` or something like this. But with my suggestion you dont need it.
Thanks!

tests/test_auto_offer_on_device_rest.py
```diff
@@ -43,24 +53,23 @@ def on_trade(self, trade_info):
...
+        simulation_id= args.simulation_id, 
+        device_id= get_area_uuid_from_area_name_and_collaboration_id(
+            args.simulation_id, args.load_names[i], 
+            'https://d3aweb-dev.gridsingularity.com'),
         domain_name='https://d3aweb-dev.gridsingularity.com',
         websockets_domain_name='wss://d3aweb-dev.gridsingularity.com/external-ws',
         autoregister=True)
     connected_devices.append(load)
     # Connects a second client to the pv device
     pv = AutoOfferBidOnMarket(
```
hannesdiedrich, line 68: Here, too. If you follow the following suggestion, then you can even parse lists that have different lengths for pv and lodas.
```suggestion
for pv in args.pv_names:
    pv = AutoOfferBidOnMarket(
```

tests/test_auto_offer_on_device_rest.py
leajski: You're right, I'm changing it

tests/test_auto_offer_on_device_rest.py
leajski: Yes, changing it as well, thanks

PR #39, opened 2020-06-24
Feature/d3 asim 2401

spyrostz:


d3a_api_client/aggregator.py
```diff
@@ -0,0 +1,114 @@
...
+        self.device_uuid_list = []
+        self.aggregator_uuid = None
+        user_aggrs = self.list_aggregators()
+        for a in user_aggrs:
+            if a["name"] == self.aggregator_name:
+                self.aggregator_uuid = a["uuid"]
+        if self.aggregator_uuid is None:
+            aggr = self._create_aggregator()
+            self.aggregator_uuid = aggr["uuid"]
+        self.start_websocket_connection()
```
hannesdiedrich, line 58: Minor but i would like it this to be refactored into its own function called "connect_to_simulation" or something like this. But skip it if you dont agree.

tests/test_aggregator.py
```diff
@@ -0,0 +1,73 @@
+import logging
```
hannesdiedrich, line 1: Could you please leave a TODO here, that this needs to be use in an integrationtest? Thanks!

spyrostz: Sure, leaving it now

d3a_api_client/aggregator.py
spyrostz: Will do, thanks

PR #40, opened 2020-06-30
Feature/d3 asim 2537: Enabling aggregator to handled/tested in an standalone scenario

faizan2590:


d3a_api_client/aggregator.py
```diff
@@ -39,10 +39,11 @@ def received_message(self, message):
 class Aggregator(RestDeviceClient):
 
     def __init__(self, simulation_id, domain_name, aggregator_name,
-                 websockets_domain_name, accept_all_devices=True):
+                 websockets_domain_name, autoregister=False, accept_all_devices=True):
```
hannesdiedrich, line 42: the aggregator does not register. This is why Spyros left this keyword out

d3a_api_client/redis_device.py
```diff
@@ -32,10 +39,41 @@ def _subscribe_to_response_channels(self):
         channel_subs[f'{self._channel_prefix}/events/tick'] = self._on_tick
         channel_subs[f'{self._channel_prefix}/events/trade'] = self._on_trade
         channel_subs[f'{self._channel_prefix}/events/finish'] = self._on_finish
+        channel_subs["crud_aggregator_response"] = self._aggregator_response_callback
```
hannesdiedrich, line 42: Can you tell me the reason for calling it `crud_aggregator`? I looked it up and it means "dirt" is this really what you meant?

d3a_api_client/redis_client_base.py
```diff
@@ -71,8 +72,8 @@ def register(self, is_blocking=False):
         logging.info(f"Trying to register to {self.area_id} as client {self.client_id}")
         if self.is_active:
             raise RedisAPIException(f'API is already registered to the market.')
-
-        data = {"name": self.client_id, "transaction_id": str(uuid.uuid4())}
+        self.device_uuid = str(uuid.uuid4())
+        data = {"name": self.client_id, "transaction_id": self.device_uuid}
```
hannesdiedrich, line 76: the device uuid is not the same as transaction_id.

d3a_api_client/aggregator.py
faizan2590: Remnant of testing, will remove it

d3a_api_client/redis_device.py
faizan2590: CRUD - Create, Read, Update & Delete is a typical short-form in CS. Since the purpose of this channel is add/delete aggregator as well as selection of aggregator locally, so I added it as a prefix, though Im open to better naming suggestion

d3a_api_client/redis_client_base.py
faizan2590: `self.device_uuid` is needed for aggregator-device mapping, so why not use `transaction_id ` at the time of instantiating since it already unique.

d3a_api_client/redis_device.py
hannesdiedrich: Why not simply "aggregator"?

d3a_api_client/redis_client_base.py
hannesdiedrich: I dont agree. It is not a big effort to create anther uuid for the transaction_id. I predict that this will lead to confusion in the future. So please revert. Thanks!

tests/test_ess_bid_redis_aggregator.py
```diff
@@ -0,0 +1,66 @@
...
+        super().__init__(*args, **kwargs)
+        self.is_finished = False
+
+    def on_market_cycle(self, market_info):
+        logging.info(f"market_info: {market_info}")
+        # if self.is_finished is True:
+        #     return
+        # if "content" not in market_info:
+        #     return
+        #
```
hannesdiedrich, line 19: please remove, also in the other files. thanks

d3a_api_client/redis_client_base.py
```diff
@@ -44,12 +44,13 @@ def __init__(self, area_id, client_id, autoregister=True, redis_url='redis://loc
...
         self.area_id = area_id
         self.client_id = client_id
+        self.device_uuid = None
         self.is_active = False
         self._blocking_command_responses = {}
         self._subscribe_to_response_channels()
         self.executor = ThreadPoolExecutor(max_workers=MAX_WORKER_THREADS)
         if autoregister:
-            self.register(is_blocking=False)
+            self.register(is_blocking=True)
```
hannesdiedrich, line 53: I wonder why this was set to False before. Is this change braking the "normal" redis device client?
No need to check before the deploy. But would be great afterwards, thanks!

d3a_api_client/redis_device.py
```diff
@@ -1,15 +1,23 @@
...
+from d3a_interface.utils import wait_until_timeout_blocking
+
+
+class RedisAPIException(Exception):
+    pass
 
 
 class RedisDeviceClient(RedisClient):
     def __init__(self, device_id, autoregister=True, redis_url='redis://localhost:6379'):
         self.device_id = device_id
```
hannesdiedrich, line 14: We should change this to "device_name" or "device_slug" as it can be confusing with `self.device_uuid`
Also something for after the deploy

d3a_api_client/redis_aggregator.py
```diff
@@ -0,0 +1,207 @@
...
+root_logger.setLevel(logging.INFO)
+
+
+class RedisAPIException(Exception):
+    pass
+
+
+class RedisAggregator:
+
+    def __init__(self, aggregator_name, autoregister=False,
```
hannesdiedrich, line 19: Pleas remove the `autoregister` parameter totally as it gives a false impression. The aggregator does not register. Thanks!

tests/test_redis_aggregator.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def on_batch_response(self, market_stats):
+        logging.info(f"AGGREGATORS_BATCH_RESPONSE: {market_stats}")
+
+
+aggregator = AutoAggregator(
+    aggregator_name="faizan_aggregator",
+    autoregister=True)
+
+sleep(5)
```
hannesdiedrich, line 63: Can you please explain why you need all the sleeps in this test file?

d3a_api_client/redis_client_base.py
faizan2590: This is in response to your comment [here](https://github.com/gridsingularity/d3a-api-client/pull/40#discussion_r449414220), now I had to explicitly wait for response to get device_uuid from d3a which is necessary for mapping aggregator with device based on their uuids.

tests/test_redis_aggregator.py
faizan2590: Remnants of testing, removing it now!

PR #41, opened 2020-07-06
D3ASIM-2538: Enhancing testing scripts for aggregators to list market…

faizan2590:
… stats of selected devices.

### DIFF
--- a/tests/test_create_aggregator.py
+++ b/tests/test_create_aggregator.py
@@ -1,7 +1,10 @@
+from pendulum import today
+import logging
 from d3a_api_client.aggregator import Aggregator
 from d3a_api_client.rest_device import RestDeviceClient
 from d3a_api_client.utils import get_area_uuid_from_area_name_and_collaboration_id
-import logging
+from d3a_interface.constants_limits import DATE_TIME_FORMAT
+from d3a_api_client.rest_market import RestMarketClient
 
 
 class TestAggregator(Aggregator):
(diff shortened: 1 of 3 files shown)


tests/test_redis_aggregator.py
```diff
@@ -74,5 +75,11 @@ def on_batch_response(self, market_stats):
 selected = pv.select_aggregator(aggregator.aggregator_uuid)
 logging.info(f"SELECTED: {selected}")
 
+
+redis_market = RedisMarketClient('house-2')
+market_slot_string_2 = today().add(minutes=60).format(DATE_TIME_FORMAT)
```
hannesdiedrich, line 80: why `_2` ?

tests/test_create_aggregator.py
```diff
@@ -106,6 +109,14 @@ def on_finish(self, finish_info):
 load2.select_aggregator(aggr.aggregator_uuid)
 pv1.select_aggregator(aggr.aggregator_uuid)
 
+area_uuid = get_area_uuid_from_area_name_and_collaboration_id(
+    simulation_id, "House", domain_name)
+
+rest_market = RestMarketClient(simulation_id, area_uuid, domain_name, websocket_domain_name)
+market_slot_string_2 = today().add(minutes=60).format(DATE_TIME_FORMAT)
+
 from time import sleep
```
hannesdiedrich, line 118: please move to top (saw it before, but forgot to comment). Thanks

tests/test_redis_aggregator.py
faizan2590: Just quick variable naming and copy/pasting. Will update with better name.

tests/test_redis_aggregator.py
```diff
@@ -74,5 +75,11 @@ def on_batch_response(self, market_stats):
...
+
 while not aggregator.is_finished:
-    sleep(0.5)
+    list_market_stats_results = redis_market.list_market_stats([market_slot_string_2])
```
hannesdiedrich, line 83: As discussed offline, this is just as reference for the aggregator user. Please move it to the on_market_cycle event and leave the while loop clean (and also revert the update time to 0.5 seconds). Thanks!

PR #42, opened 2020-07-07
D3ASIM-2562: Scheduling auto jwt_token refresh before it expires.

faizan2590:


d3a_api_client/rest_device.py
```diff
@@ -21,13 +22,20 @@ def __init__(self, simulation_id, device_id, domain_name,
         self.websockets_domain_name = websockets_domain_name
         self.aggregator_prefix = get_aggregator_prefix(domain_name, simulation_id)
         self.active_aggregator = None
+        self.lock = Lock()
+        self.jwt_token_refresh = RepeatingTimer(3500, self.refresh_jwt_token, [domain_name])
```
hannesdiedrich, line 26: isnt the timeout of the tocken set in a variable somewhere? Cant we use this to get the "just before expiring" time?

faizan2590: Its set in other repo too but also hard coded there too, so I will take you advise and replace it with a common constant

PR #45, opened 2020-07-17
Adding script to get area name uuid mapping from a configuration.

faizan2590:


d3a_api_client/utils.py
```diff
@@ -126,6 +126,36 @@ def get_area_uuid_from_area_name_and_collaboration_id(collab_id, area_name, doma
     return area_uuid
 
 
+def _name_uuid_mapping_wrapper(serialized_scenario, mapping={}):
```
hannesdiedrich, line 129: Sorry, but this is not a wrapper, right? This is the function that recursively updates the mapping. So, IMO this naming is misleading, I propose: `_get_name_uuid_mapping` Thanks!

d3a_api_client/utils.py
```diff
@@ -126,6 +126,36 @@ def get_area_uuid_from_area_name_and_collaboration_id(collab_id, area_name, doma
...
+    headers = {'Authorization': f'JWT {jwt_key}', 'Content-Type': 'application/json'}
+
+    query = 'query { readConfiguration(uuid: "{' + collab_id + \
+            '}") { scenarioData { latest { serialized } } } }'
+
+    endpoint = HTTPEndpoint(url, headers)
+    data = endpoint(query=query)
+    area_name_uuid_map = _name_uuid_mapping_wrapper(
+        json.loads(data["data"]["readConfiguration"]["scenarioData"]["latest"]["serialized"])
+    )
```
hannesdiedrich, line 155: Can you please put this in a try except block and return a user friendly error message in case the query does not return a dict that has this structure? Thanks!

tests/extract_area_uuid_from_area_name.py
```diff
@@ -0,0 +1,20 @@
...
+# set simulation parameters
+simulation_id = str(sys.argv[3])
+domain_name = 'https://d3aweb-dev.gridsingularity.com'
+
+mapping = get_area_uuid_and_name_mapping_from_simulation_id(
+    simulation_id, domain_name
+)
+
+with open(os.path.join(os.getcwd(), 'area_name_uuid_map.json'), "w") as mapped:
+    mapped.write(json.dumps(mapping, indent=2))
```
hannesdiedrich, line 20: json can dump directly to files:
```suggestion
with open(os.path.join(os.getcwd(), 'area_name_uuid_map.json'), "w") as outfile:
    json.dump(mapping, outfile, indent=2)
```

hannesdiedrich (approved):
One last comment, otherwise LGTM

spyrostz (approved):
LGTM

PR #47, opened 2020-07-23
D3ASIM-2636: Adding Enabling pubsub object sharing to limit number of…

faizan2590:
… threads.

### DIFF
--- a/d3a_api_client/redis_client_base.py
+++ b/d3a_api_client/redis_client_base.py
@@ -77,9 +77,9 @@ class RedisClient(APIClientInterface):
         if self.is_active:
             raise RedisAPIException(f'API is already registered to the market.')
         data = {"name": self.client_id, "transaction_id": str(uuid.uuid4())}
-        self.redis_db.publish(f'{self.area_id}/register_participant', json.dumps(data))
         with self.lock:
             self._blocking_command_responses["register"] = data
+        self.redis_db.publish(f'{self.area_id}/register_participant', json.dumps(data))
 
         if is_blocking:
             try:
--- a/d3a_api_client/redis_client_base.py
+++ b/d3a_api_client/redis_client_base.py
@@ -7,6 +7,7 @@ from redis import StrictRedis
 from d3a_interface.utils import wait_until_timeout_blocking
 from d3a_api_client import APIClientInterface
 from concurrent.futures.thread import ThreadPoolExecutor
+from threading import Lock
 from d3a_api_client.constants import MAX_WORKER_THREADS
 
 root_logger = logging.getLogger()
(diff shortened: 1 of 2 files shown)


d3a_api_client/redis_device.py
```diff
@@ -75,6 +77,14 @@ def select_aggregator(self, aggregator_uuid, is_blocking=True):
...
             except AssertionError:
                 raise RedisAPIException(f'API has timed out.')
+        self.pubsub.unsubscribe(
+            [f'{self.area_id}/response/register_participant',
+             f'{self.area_id}/response/unregister_participant',
+             f'{self._channel_prefix}/events/market',
+             f'{self._channel_prefix}/events/tick',
+             f'{self._channel_prefix}/events/trade',
+             f'{self._channel_prefix}/events/finish']
+        )
```
hannesdiedrich, line 87: Can you please explain why this is needed? Maybe the client wants to see some events even though it as selected an aggregator!?

d3a_api_client/redis_client_base.py
```diff
@@ -66,15 +69,17 @@ def _subscribe_to_response_channels(self):
...
+        if pubsub is None:
+            self.pubsub.run_in_thread(daemon=True)
 
     def register(self, is_blocking=False):
         logging.info(f"Trying to register to {self.area_id} as client {self.client_id}")
         if self.is_active:
             raise RedisAPIException(f'API is already registered to the market.')
         data = {"name": self.client_id, "transaction_id": str(uuid.uuid4())}
+        with self.lock:
+            self._blocking_command_responses["register"] = data
```
hannesdiedrich, line 81: Why are you locking a `assigning to a dict` command? It is already blocking.

d3a_api_client/redis_device.py
hannesdiedrich: Discussed offline. Please add a flag to toggle this unsubscription. Thanks

d3a_api_client/redis_client_base.py
hannesdiedrich: As discussed offline, please remove the lock and change the order of assigning to _blocking_command_responses and publishing on other places too. Thanks

d3a_api_client/redis_client_base.py
```diff
@@ -37,22 +38,24 @@ class Commands(Enum):
 
 
 class RedisClient(APIClientInterface):
-    def __init__(self, area_id, client_id, autoregister=True, redis_url='redis://localhost:6379'):
+    def __init__(self, area_id, client_id, autoregister=True, redis_url='redis://localhost:6379',
+                 pubsub=None):
```
hannesdiedrich, line 42: Please_rename this to maybe:
`pubsub_thread` or `redis_thread` Thanks!

d3a_api_client/redis_device.py
```diff
@@ -54,7 +56,7 @@ def _aggregator_response_callback(self, message):
     def _check_transaction_id_cached_out(self, transaction_id):
         return transaction_id in self._transaction_id_buffer
 
-    def select_aggregator(self, aggregator_uuid, is_blocking=True):
+    def select_aggregator(self, aggregator_uuid, is_blocking=True, should_unsubscribe=True):
```
hannesdiedrich, line 59: Minor, but IMO this flag naming is misleading. I would propose: 'unsubscribe_from_device_events'.

PR #48, opened 2020-08-20
Feature/d3 asim 2677

hannesdiedrich:


d3a_api_client/utils.py
```diff
@@ -50,21 +52,17 @@ def post_request(endpoint, data, jwt_token):
...
-                      f"Response body: {resp.text} {resp.reason}")
-        return False
-    return True
+    return json.loads(resp.text) if request_response_returns_http_200(endpoint, resp) else None
 
 
 def blocking_post_request(endpoint, data, jwt_token):
     data["transaction_id"] = str(uuid.uuid4())
-    response = requests.post(
+    resp = requests.post(
```
faizan2590, line 60: Any reason to rename variable from `response ` to `resp ` ?

d3a_api_client/rest_market.py
```diff
@@ -47,14 +49,20 @@ def _on_market_cycle(self, message):
...
 
     def _on_finish(self, message):
         logging.debug(f"Simulation finished. Information: {message}")
 
         def executor_function():
-            self.on_finish(message)
+            try:
+                self.on_finish(message)
+            except Exception as e:
+                logging.error(f"on_finish raised exception: {e}. Traceback: {traceback.format_exc()}")
```
faizan2590, line 65: Would be better IMO to have traceback starting from new line

faizan2590 (commented):
One general comment is to extend the logging of exceptions for RedisClient too.

d3a_api_client/utils.py
hannesdiedrich: To be consistent with the other functions in this module, too.

d3a_api_client/rest_market.py
hannesdiedrich: I agree, will add a newline.

hannesdiedrich:
> One general comment is to extend the logging of exceptions for RedisClient too.

Ok, will do so.

d3a_api_client/rest_device.py
```diff
@@ -139,28 +143,40 @@ def _on_market_cycle(self, message):
...
 
     def _on_finish(self, message):
         logging.debug(f"Simulation finished. Information: {message}")
 
         def executor_function():
-            self.on_finish(message)
+            try:
+                self.on_finish(message)
+            except Exception as e:
+                logging.error(f"on_finish raised exception: {e}. Traceback: {traceback.format_exc()}")
```
spyrostz, line 179: In order to make it more straightforward to detect which device failed in multi-device tests, it would be great to also have the device_id in the log too.

spyrostz (approved):
2 comments, LGTM otherwise

d3a_api_client/aggregator.py
```diff
@@ -71,7 +71,10 @@ def start_websocket_connection(self):
 
     @logging_decorator('create_aggregator')
     def list_aggregators(self):
-        return blocking_get_request(f'{self.aggregator_prefix}list-aggregators/', {}, self.jwt_token)
+        list_of_aggregators = blocking_get_request(f'{self.aggregator_prefix}list-aggregators/', {}, self.jwt_token)
+        if list_of_aggregators is None:
+            raise Exception(f"No aggregators found on {self.aggregator_prefix}")
```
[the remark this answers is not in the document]
hannesdiedrich: I agree, but the problem is that if the user runs the locally, the user aggregator scrips fail later (when selecting a aggregator) and there is no feedback. So, I suggest for this case to not raise an exception, but to log.error and return an empty list. OK?

d3a_api_client/rest_device.py
hannesdiedrich: Good idea! Will add it. Thanks

d3a_api_client/aggregator.py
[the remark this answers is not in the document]
spyrostz: Agreed! Thanks

PR #50, opened 2020-09-07
D3ASIM-2708: Having the same grid_fee_implementation for redis and re…

faizan2590:
…st client.


d3a_api_client/redis_market.py
```diff
@@ -58,7 +59,7 @@ def list_market_stats(self, market_slot_list):
         self.redis_db.publish(f"{self.area_id}/market_stats", json.dumps({"market_slots": market_slot_list}))
         return self._wait_and_consume_command_response("market_stats")
 
-    def change_grid_fees_const(self, fee_cents_kwh):
+    def grid_fees(self, fee_cents_kwh):
```
hannesdiedrich, line 62: Was this a requirement from Colin? IMO `grid_fees` does not tell mw what the function does.

d3a_api_client/redis_market.py
```diff
@@ -11,7 +12,7 @@
 
 class RedisMarketClient:
     def __init__(self, area_id, redis_url='redis://localhost:6379'):
-        self.area_id = area_id
+        self.area_id = slugify(area_id, to_lower=True)
```
hannesdiedrich, line 15: Can you please check whether this is also needed in `RedisClient`?
Also, why not changing this on the d3a side and skip slugifying the name here?

hannesdiedrich (commented):
Left some comments, happy to discuss them.

d3a_api_client/redis_market.py
hannesdiedrich: Sorry, just read the story description. But IMO `get_grid_fees` would be more descriptive.

d3a_api_client/redis_market.py
faizan2590: I checked and this is not needed for `RedisClient ` as devices don't slugify their channel prefix whereas area market do, you are right that it could be done on d3a side but it would be a rather time consuming refactoring (may be not so important either) that is to adapt not just `d3a` but also `d3a-web` too, so I took the the easier approach to simply slugify on d3a-api-client side.

d3a_api_client/redis_market.py
faizan2590: Yes your suggestion makes sense but then we had to rename it in `RestMarket` too here(https://github.com/gridsingularity/d3a-api-client/blob/master/d3a_api_client/rest_market.py#L37) which I don't want to disturb that is possibly already been in use by some external data scientists

hannesdiedrich: Ahh, it is actually SETTING the grid fees. Ok. 
Yeah, keep it as `grid_fees`.

d3a_api_client/redis_market.py
hannesdiedrich: Actually, I don't see a point in the d3a-web where the channel is sluggified. But yeah, you are right. It is too much work for this side task. But IMO we should be consistent on all parts. So If you agree, please add a refactoring story that either removes the sluggification for the channel names or add it to all (market AND device) .

hannesdiedrich (approved):
LGTM

PR #59, opened 2020-10-22
D3ASIM-2925: Added test scripts for hardware api

hannesdiedrich:


d3a_api_client/rest_device.py
```diff
@@ -80,6 +80,20 @@ def unselect_aggregator(self, aggregator_uuid):
...
 
+    @logging_decorator('pv_power_forecast')
+    def set_pv_power_forecast(self, pv_power_forecast):
+        transaction_id, posted = self._post_request('set_pv_power_forecast',
+                                                    {"pv_power_forecast": pv_power_forecast})
+        if posted:
+            return self.dispatcher.wait_for_command_response('set_pv_power_forecast', transaction_id)
+
+    @logging_decorator('set_load_power_forecast')
+    def set_load_power_forecast(self, pv_power_forecast):
```
spyrostz, line 91: Minor, but argument name should be `load_power_forecast`. Also a unit suffix would help in both functions' arguments.

spyrostz (approved):
LGTM, with one reservation about whether there will be a need for aggregator functions. For now it should not be needed though, therefore LGTM

tests/test_hardware_api_trading_bot.py
```diff
@@ -0,0 +1,69 @@
+"""
```
faizan2590, line 1: As a general comment, please use `set_pv_power_forecast ` and `set_load_power_forecast ` under the method `on_market_cycle `as an example for user to send its agent's forecasted energy consumption/production for the next market.

hannesdiedrich: @faizan2590 I added a separate Script for that: https://github.com/gridsingularity/d3a-api-client/pull/59/files#diff-9e7b1acf31c4255a76caade7f40642b13f1b09f02d5dae95b0a598cb75383e51
(to simulate that the trading bot will not send the forecast in real live)

tests/test_sending_power_forecast.py
```diff
@@ -0,0 +1,64 @@
...
+        """
+        root_logger.debug(f"New market information {market_info}")
+        self.forecast += 50
+        if "available_energy_kWh" in market_info["device_info"]:
+            root_logger.info(f"self.set_pv_power_forecast({self.forecast})")
+            response = self.set_pv_power_forecast(self.forecast)
+
+        if "energy_requirement_kWh" in market_info["device_info"]:
+            root_logger.info(f"self.set_load_power_forecast({self.forecast})")
+            response = self.set_load_power_forecast(self.forecast)
```
faizan2590, line 29: I have a general comment, this seems like a hacky way, meaning if the strategy contains `available_energy_kWh ` so send the forecasted pv power or if the strategy contains `energy_requirement_kWh ` send the forecasted load power. Im not saying it won't work but it gives a false image to user that while removing
`if "available_energy_kWh" in market_info["device_info"]:` and `if "energy_requirement_kWh" in market_info["device_info"]:` can send the load forecast as pv and the other way around.

TLDR: Better to have separate scripts for pv forecast and load forecast.

hannesdiedrich: @faizan2590 Yes, of course, I agree with your comment, but this is really only a script for testing the hardware api. 
I really made it for you. The user (or here: the hardware) should know that it can only send the forecast of a load with `set_load_power_forecast`.

faizan2590 (approved):
LGTM

PR #61, opened 2020-11-02
Feature/d3 asim 2927

spyrostz:


mqtt_subscriber/mqtt_redis_job.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def connect_to_running_canary_networks():
+    cns = list_running_canary_networks_and_devices_with_live_data(environ["API_CLIENT_DOMAIN_NAME"])
+
+    if not cns:
+        return
+
+    redis_conn = StrictRedis.from_url(environ.get('REDIS_URL', 'redis://localhost'), retry_on_timeout=True)
+    q = Queue(name="canary_mqtt", connection=redis_conn)
```
faizan2590, line 56: Minor but more than one alphabet variable would look nice like `queue`

mqtt_subscriber/mqtt_redis_job.py
```diff
@@ -0,0 +1,72 @@
...
+from redis import StrictRedis
+from rq import Connection, Worker, get_current_job, Queue
+from rq.decorators import job
+from mqtt_subscriber.oli_broker import MQTTConnection
+from d3a_api_client.utils import get_area_uuid_from_area_name_and_collaboration_id, \
+    list_running_canary_networks_and_devices_with_live_data
+from d3a_api_client.rest_device import RestDeviceClient
+
+
+allowed_device_names = ["DOSE/OLI_42"]
```
faizan2590, line 14: Since this is a public repo, wouldn't be nice to expose devices names via env variable in general concerning user privacy, though in this case its totally fine as interested parties wants to market it as PoC.

mqtt_subscriber/oli_broker.py
```diff
@@ -0,0 +1,47 @@
...
+        self.api_clients = api_client_dict
+        self.client = mqtt.Client()
+        self.client.on_connect = self.on_connect
+        self.client.on_message = self.on_message
+
+    def on_connect(self, client, userdata, flags, rc):
+        logging.debug(f"Connected with result code {str(rc)}")
+
+        for device_name in self.api_clients:
+            client.subscribe(f"{device_name}/activeEnergy/*")
```
faizan2590, line 23: So, mqtt uses `.subscribe` for both exact and pattern subscription?

mqtt_subscriber/oli_broker.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def on_message(self, client, userdata, msg):
+        logging.debug(f"{msg.topic} {str(msg.payload)}")
```
faizan2590, line 26: `topic` in mqtt are synonym of `channel` in redis?

mqtt_subscriber/mqtt_redis_job.py
```diff
@@ -0,0 +1,72 @@
...
+    job.save_meta()
+
+    device_args = {
+        "simulation_id": configuration_id,
+        "domain_name": environ["API_CLIENT_DOMAIN_NAME"],
+        "websockets_domain_name": environ["API_CLIENT_WEBSOCKET_DOMAIN_NAME"],
+        "autoregister": True
+    }
+
+    api_clients = {}
```
faizan2590, line 31:
```suggestion
    mqtt_rest_api_clients_mapping = {}
```
Bit verbose but also quick to understand its purpose.

d3a_api_client/utils.py
```diff
@@ -169,3 +169,35 @@ def wrapped(self, *args, **kwargs):
...
+            forecastStreamAreaMapping
+          }
+        }
+      }
+    }
+    '''
+
+    endpoint = HTTPEndpoint(url, headers)
+    data = endpoint(query=query)
+    print(data)
```
faizan2590, line 198: Reminder to remove before merging.

faizan2590 (approved):
Some minor comments, other than that LGTM.

mqtt_subscriber/mqtt_redis_job.py
spyrostz: Fixing this

mqtt_subscriber/mqtt_redis_job.py
spyrostz: Indeed it is not nice in the long run, however for the foreseeable future we do not expect to use this code for other purposes than the hackathon. If we do so in the long run, we will have to refactor this not via an env variable (because the name list can grow), but via an endpoint, in order to maintain the device list in the database.  Definitely not for now though

mqtt_subscriber/oli_broker.py
spyrostz: Exactly, uses the same function to subscribe and psubscribe

mqtt_subscriber/oli_broker.py
spyrostz: Exactly

mqtt_subscriber/mqtt_redis_job.py
spyrostz: Indeed, will rename this

d3a_api_client/utils.py
spyrostz: Will do so, thanks

mqtt_subscriber/oli_broker.py
```diff
@@ -0,0 +1,47 @@
...
+
+        try:
+            device_name = msg.topic.split("/activeEnergy/")[0]
+            energy = msg.payload["value"]
+            power = energy * (MINUTES_IN_HOUR / MEASUREMENT_PERIOD_MINUTES)
```
hannesdiedrich, line 31: Is the MEASUREMENT_PERIOD_MINUTES fixed on the Oli side? 
I am asking because they could publish "forecasts" more often if they like, there is no limit here. Right?

spyrostz: Yes, that's what they promised us and that's why they do not send us the measurement period even though this exists in their message structure. They only send us the energy values and the timestamp in the messages.

hannesdiedrich (approved):
LGTM

PR #63, opened 2020-11-07
D3ASIM-2927: Fixed parsing of the live data device info from the MQTT…

spyrostz:
… subscriber. Do not read live data for Canary Networks that are not running. Fixed decoding in received MQTT data.

### DIFF
--- a/d3a_api_client/utils.py
+++ b/d3a_api_client/utils.py
@@ -141,7 +141,6 @@ def get_area_uuid_and_name_mapping_from_simulation_id(collab_id, domain_name):
     jwt_key = retrieve_jwt_key_from_server(domain_name)
     from sgqlc.endpoint.http import HTTPEndpoint
 
-
     url = f"{domain_name}/graphql/"
     headers = {'Authorization': f'JWT {jwt_key}', 'Content-Type': 'application/json'}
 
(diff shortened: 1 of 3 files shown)


faizan2590 (approved):
LGTM

d3a_api_client/utils.py
```diff
@@ -197,6 +197,7 @@ def list_running_canary_networks_and_devices_with_live_data(domain_name):
     data = endpoint(query=query)
 
     return {
-        cn["uuid"]: json.loads(cn["scenarioData"]["forecastStreamAreaMapping"])
+        cn["uuid"]: cn["scenarioData"]["forecastStreamAreaMapping"]
         for cn in data["data"]["listCanaryNetworks"]["configurations"]
-    }
+        if cn["resultsStatus"] == "running"
```
hannesdiedrich, line 202: `else None` is not needed here?

hannesdiedrich (approved):
LGTM

d3a_api_client/utils.py
spyrostz: No, if the Canary Network is not running then we simply do not connect to it when the MQTT subscriber starts. When the CN starts the MQTT subscriber will be notified via the Redis queue.

PR #64, opened 2020-11-09
D3ASIM-2927: Reverted from the rq-based process queue to a normal Pyt…

spyrostz:
…hon process which polls to get updated CN/device values.


mqtt_subscriber/oli_broker.py
```diff
@@ -3,40 +3,64 @@
...
+
+
+class MQTTBrokerConnectionError(Exception):
+    pass
 
 
 MQTT_PORT = 1883
 MINUTES_IN_HOUR = 60
 MEASUREMENT_PERIOD_MINUTES = 15
+RELOAD_CN_DEVICE_LIST_TIMEOUT_SECONDS = 60 * 15
```
faizan2590, line 18:
```suggestion
RELOAD_CN_DEVICE_LIST_TIMEOUT_SECONDS = 60 * MEASUREMENT_PERIOD_MINUTES
```
Seems quicker to understand the math

faizan2590 (approved):
lgtm

mqtt_subscriber/oli_broker.py
```diff
@@ -3,40 +3,64 @@
...
             payload = json.loads(msg.payload.decode("utf-8"))
 
-            device_name = msg.topic.split("/activeEnergy/")[0]
             energy = payload["value"]
             power = energy * (MINUTES_IN_HOUR / MEASUREMENT_PERIOD_MINUTES)
-            self.api_clients[device_name].set_power_forecast(power)
+
+            # Transmit power values to CN
+            for api_args in self.topic_api_client_dict[msg.topic]:
+                RestDeviceClient(**api_args).set_power_forecast(power, do_not_wait=True)
```
hannesdiedrich, line 52: So you are re-initiating the RestDevices every market slot?

mqtt_subscriber/mqtt_redis_job.py
```diff
@@ -1,98 +0,0 @@
-import json
```
hannesdiedrich, line 1: minor, but this file can be removed now i guess

hannesdiedrich (approved):
One question, LGTM

mqtt_subscriber/oli_broker.py
spyrostz: Yes, for simplicity's sake I thought it was better to at first, completely ignore the response from the d3a via the websocket and just send the request. That way we do not have to maintain long-running connections that are prone to crash due to e.g. a network error.

PR #68, opened 2020-11-18
D3ASIM-3028: Added retry support for the websocket connection for the…

spyrostz:
… api client. Added test coverage.

### DIFF
--- a/d3a_api_client/websocket_device.py
+++ b/d3a_api_client/websocket_device.py
@@ -4,7 +4,13 @@ import threading
 import traceback
 import websockets
 import json
+from time import time
 from d3a_interface.utils import wait_until_timeout_blocking
+
+
+WEBSOCKET_MAX_CONNECTION_RETRIES = 5
+WEBSOCKET_WAIT_BEFORE_RETRY_SECONDS = 5
+WEBSOCKET_ERROR_THRESHOLD_SECONDS = 30
 
 
 class WebsocketMessageReceiver:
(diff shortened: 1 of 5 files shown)


d3a_api_client/websocket_device.py
```diff
@@ -55,6 +61,21 @@ async def websocket_coroutine(websocket_uri, websocket_headers, message_dispatch
...
 
+async def retry_coroutine(websocket_uri, websocket_headers, message_dispatcher, retry_count=0):
+    ws_connect_time = time()
+    try:
+        await websocket_coroutine(websocket_uri, websocket_headers, message_dispatcher)
+    except Exception as e:
+        logging.warning(f"Connection failed, trying to reconnect.")
+        ws_error_time = time()
+        if ws_error_time - ws_connect_time > WEBSOCKET_ERROR_THRESHOLD_SECONDS:
+            retry_count = 0
```
hannesdiedrich, line 72: Can you please explain what you try to achieve here? 
Also, i dont see the retry_timeout increased somewhere. In the unittest you are mocking it with a global variable `coro_execution_counter` so it indeed gets increased, but not here. Or don't i see the forrest behind all the trees ? ;-)

hannesdiedrich (commented):
One comment. Will review after addressed.

d3a_api_client/websocket_device.py
hannesdiedrich: Alright, i missed the last line actually. 
But I have another question: if the `connection manages to be established`, no Exception would be raised in `websocket_coroutine` and this part of the code would not be run, right?

spyrostz: No exception will be raised indeed if the connection never crashes (in which case we do not need the retry), therefore all will go well. If, after a couple of hours though, the connection crashes again, an exception will be raised and the retry counter needs to start from 0 and not from the last value of the retry_count. In order to differentiate if the connect has succeeded for some time or not, the easiest way is to use a timeout for that (WEBSOCKET_ERROR_THRESHOLD_SECONDS) in our case.

hannesdiedrich: AH, ok, makes sense, thanks for being patient with me!

hannesdiedrich (approved):
LGTM

d3a_api_client/websocket_device.py
spyrostz: No worries, thanks!

faizan2590 (approved):
LGTM

PR #70, opened 2020-12-03
TESTS-52: Test file for the device clients

Ravineesh:
Two test files for the device client which will connect to a collab for placing offers and bids in the market.

### DIFF
--- a/tests/e2e_test_auto_offer_on_device_rest.py
+++ b/tests/e2e_test_auto_offer_on_device_rest.py
@@ -1,0 +1,141 @@
+"""
+Test file for the device client (Load, PV, and Storage). Depends on d3a test setup file strategy_tests.external_devices
+"""
+import os
+import json
+import logging
+import sys
+import argparse
+from time import sleep
+from random import randint
+from d3a_api_client.rest_device import RestDeviceClient
+from d3a_api_client.utils import get_area_uuid_from_area_name_and_collaboration_id
+# parse arguments
+parser = argparse.ArgumentParser()
+parser.add_argument("--simulation_id", type=str, help="Simulation uuid")
+parser.add_argument("--load_names", nargs='+')
+parser.add_argument("--pv_names", nargs='+')
+parser.add_argument("--storage_names", nargs='+')
+args = parser.parse_args()
+connected_devices = []
+
+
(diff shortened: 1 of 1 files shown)


hannesdiedrich:
I have some general comments:
- AFAIS, you dont need one script for "2 devices" and one for "3 devices" because for not trading for the storage, you just dont parse a list of storages to the 3-device-script, so, IMO, you can get rid of the 2-device script
- in addition to that, the filename should not contain the number of devices, but it would be good if the scripts used in the end to end tests are market, e.g. with prepending: `e2e_`

tests/test_auto_offer_on_3_device_rest.py
```diff
@@ -0,0 +1,146 @@
...
+        logging.info("***********Market Info***************")
+        logging.info(market_info["device_info"])
+        logging.info("*************************************")
+        
+        if "available_energy_kWh" in market_info["device_info"] and market_info["device_info"]["available_energy_kWh"] > 0.0:
+            offered_energy = market_info["device_info"]["available_energy_kWh"]
+            offered_rate = 28*offered_energy
+            offer_response = self.offer_energy_rate(offered_energy, offered_rate)
+            expected_price = offered_rate*offered_energy
+            logging.info("*****************Offer Placed in Market by PV***************")
```
hannesdiedrich, line 49: In order to make this script applicable for multiple devices, I suggest to print the name of the device here and in the other places for better debugging. Reducing the number of `logger.info` calls will also make the debugging easier because there are less logging messages then.

hannesdiedrich (commented):
Some general comments, but apart from them, LGTM (if the trading script works :-) )

tests/test_auto_offer_on_3_device_rest.py
Ravineesh: by device name, you mean which are passed as the argument ..right?

hannesdiedrich: I thought the device name is sent in the market_info. But it is not unfortunately.
You can distinguish the devices between each other with the `area_uuid`, which should be available in the market_info. So I wold be in favor of printing the area_uuid when printing the offer/bid placed message.

faizan2590 (approved):
LGTM if trading script works.

PR #74, opened 2020-12-10
D3ASIM-3092:Upgrade python version to 3.8 + upgraded packages

Hassan754:
# To Note
Should not be merged before merging the d3a-interface upgrade PR


spyrostz (approved):
LGTM, however one comment, I would recommend if it is not needed to remove the `dump.rdb` file from the PR, thanks!

hannesdiedrich (approved):
LGTM

PR #76, opened 2020-12-23
TESTS-52: Changing testing url from dev to staging

Ravineesh:
Changing testing url from dev to staging for time being.

### DIFF
--- a/tests/e2e_test_auto_offer_on_device_rest.py
+++ b/tests/e2e_test_auto_offer_on_device_rest.py
@@ -18,8 +18,8 @@ parser.add_argument("--storage_names", nargs='+', default=[])
 args = parser.parse_args()
 connected_devices = []
 
-domain_name = 'https://d3aweb-dev.gridsingularity.com'
-websockets_domain_name = 'wss://d3aweb-dev.gridsingularity.com/external-ws'
+domain_name = 'https://d3aweb-staging.gridsingularity.com'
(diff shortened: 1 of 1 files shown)


Hassan754:
Why do you need to change the testing url in the first case

Ravineesh:
> Why do you need to change the testing url in the first case



> Why do you need to change the testing url in the first case

It is required for time being to pass one of the test cases related to collaboration where clients will use the trade using trading script.

faizan2590 (approved):
LGTM

Hassan754 (approved):

tests/e2e_test_auto_offer_on_device_rest.py
```diff
@@ -18,8 +18,8 @@
 args = parser.parse_args()
 connected_devices = []
 
-domain_name = 'https://d3aweb-dev.gridsingularity.com'
-websockets_domain_name = 'wss://d3aweb-dev.gridsingularity.com/external-ws'
+domain_name = 'https://d3aweb-staging.gridsingularity.com'
+websockets_domain_name = 'wss://d3aweb-staging.gridsingularity.com/external-ws'
```
hannesdiedrich, line 22: @Ravineesh Sorry, just saw that. I would have suggested to add these to variables as a argument to the script to easily be able to switch between domains from the d3a-e2e-ui-test side.

Ravineesh: Ok thanks. I'll change it when I'll switch it back to dev url.

PR #80, opened 2021-01-21
feature/D3ASIM-3213: Abstract batch commands API to python functions

Hassan754:
### Status
`Ready`
The batch commands experience is replaced in the Rest Aggregator device only and implemented in the rest one.

### Notes
Added python function wrapper to the batch commands interface of aggregatorsfollowing the `Command Design Pattern`

### Implementation Notes
Instead of calling batch commands and passing a dict with commands, we made the following:

1-Created a ClientCommand class which is a wrapper for the commands 
This class contains the methods of the client, like the `def offer_energy`, `def bid_energy` and so on
To use this functionality, the end user will call add_to_batch_commands then the respective method in it, 
The end user can then call the `execute_batch` function which will return a dict with the same format needed to be passed to `d3a-web` i.e:
`aggregator.add_to_batch_commands.offer_energy(**kwargs); dict = aggregator.execute_batch()`


d3a_api_client/redis_client_base.py
```diff
@@ -313,4 +307,12 @@ def on_trade(self, trade_info):
...
 
     def on_finish(self, finish_info):
-        pass
\ No newline at end of file
+        pass
+
+    def add_to_batch(self, area_uuid: str, ):
+        command = ClientCommand(area_uuid=area_uuid)
+        self.commands_buffer.append(command)
+        return command
```
Hassan754, line 315: This makes adding to commands_buffer like this 
`device.add_to_batch(area_uuid).offer(energy, price)`
or 
`device.add_to_batch(area_uuid).delete_offer(offer_id)`

spyrostz: Why do you need the area_uuid explicitly ? This should be part of the class already (self.area_id should give you that exact information)

Hassan754: I could not understand the reason for explicitly get it from the event object instead of from self in [here](https://github.com/gridsingularity/d3a-api-client/blob/feature/D3ASIM-3213/tests/test_create_aggregator.py#L35)
In tests the area_uuid is being passed explicitly

spyrostz: I also find this approach a bit more difficult for the end user than it should be. Instead of exposing a function that returns a ClientCommand object, which in turn can be used to call offer, delete_offer etc. , I think it would be much easier to either have dedicated methods on the RedisClient (similar to what we have already with the bid_energy, offer_energy), or, if you want to keep the existing structure, to expose `add_to_batch` as a property/class member and not as a function.

Hassan754: The second option is what I had in mind , imo it would make the code cleaner 
anw concerning the `area_uuid` point , can I safely use the `self.area_id` ? and just to note it looks like this variable refers to the name of an area not to its ID

d3a_api_client/commands.py
```diff
@@ -0,0 +1,68 @@
...
+
+    def device_info(self):
+        self.action = Commands.DEVICE_INFO
+
+    def execute(self) -> dict:
+        if self.area_uuid:
+            return {self.area_uuid: {"type": command_enum_to_command_name(self.action), **self.callback_args}}
+
+
+class ClientCommandList(Command, list):
```
spyrostz, line 55: There is no need for multiple inheritance and extending the list here. You can have a private member, namely `_pending_commands`, that will hold all pending commands that have not been sent yet. 
Generally speaking, I usually favor composition over inheritance: https://en.wikipedia.org/wiki/Composition_over_inheritance

d3a_api_client/redis_client_base.py
spyrostz: OK, now I see the point of confusion. The batch_command is only part of the aggregator api, and not part of the device api. An aggregator is an entity that is able to control multiple devices at once. That's why you actually need the area_uuid argument, because the user indeed needs to define in which area should the command be sent to. 
Taking this into account, I think that the best solution would be to add it as an extra argument to each method. That IMO would make the api easier to use.

d3a_api_client/commands.py
Hassan754: I have this already implemented as `commands_buffer` variable in the RedisClient class , I'll make it private and make it a list instead of a ClientCommandList object , the only issue here is where to put the batch commands execute logic , I can make it a static function inside the ClientCommand class if suitable

d3a_api_client/aggregator.py
```diff
@@ -107,16 +108,11 @@ def _all_uuids_in_selected_device_uuid_list(self, uuid_list):
...
 
-    def batch_command(self, batch_command_dict):
-        """
-        batch_dict : dict where keys are device_uuids and values list of commands
-        e.g.: batch_dict = {
-                        "dev_uuid1": [{"energy": 10, "rate": 30, "type": "offer"}, {"energy": 9, "rate": 12, "type": "bid"}],
-                        "dev_uuid2": [{"energy": 20, "rate": 60, "type": "bid"}, {"type": "list_market_stats"}]
-                        }
-        """
+    def batch_command(self):
```
Hassan754, line 111: I think I should rename this method to something more meaningful

d3a_api_client/redis_client_base.py
```diff
@@ -313,4 +307,16 @@ def on_trade(self, trade_info):
         pass
 
     def on_finish(self, finish_info):
-        pass
\ No newline at end of file
+        pass
+
+    def add_to_batch(self, ):
```
Hassan754, line 312: I would have to name this too , I am not sure if this makes sense `device.add_to_batch.offer()`

d3a_api_client/commands.py
spyrostz: Instead of keeping the container on the RedisClient class, why not storing it in the ClientCommand class directly ? That way you would store and control the functionality of the commands in the same class, and also you would not have to make the `execute()` method static.

Hassan754: @spyrostz Done , can you check now please

d3a_api_client/redis_client_base.py
hannesdiedrich: I think this is ok, but how about this: 
`device.add_to_batch_commands.offer()`

hannesdiedrich: @Hassan754 i dont see this function in the code diff. did you change this?

hannesdiedrich: ah, ok, you changed somethin later . OK

Hassan754: Works for me, done.

d3a_api_client/aggregator.py
hannesdiedrich: Maybe `execute_batch_commands` or `send_batch_commands`

d3a_api_client/commands.py
```diff
@@ -0,0 +1,72 @@
...
+class ClientCommandBuffer:
+
+    def __init__(self, ):
+        self._commands_buffer = []
+
+    def offer_energy(self, area_uuid, energy, price):
+        area_uuid = area_uuid
+        callback_args = {energy, price}
+        action = Commands.OFFER
+        self.execute(area_uuid, action, callback_args)
```
hannesdiedrich, line 13: Sorry, i just saw that. IMO these are not callbac_args, but "normal" arguments for the command, so I propose renaming them to `args` or similar.
Or can you tell my you have chosen this naming?

hannesdiedrich (approved):
One last renaming comment. LGTM 
In general we should start adding integrationtests again. Or shall we "outsource" it to the testing department after the e2e tests are matured @spyrostz?

d3a_api_client/commands.py
Hassan754: Yep I created this PR with a different approach and these were callback args which were not called immediately but rather later, but this was changed.
I will rename it, Thanks !

spyrostz:
> One last renaming comment. LGTM
> In general we should start adding integrationtests again. Or shall we "outsource" it to the testing department after the e2e tests are matured @spyrostz?

I think we should add these ourselves, especially the first time a feature is implemented. IMO this is a very good example where a Redis integration test should be added to the API client.

d3a_api_client/commands.py
```diff
@@ -0,0 +1,72 @@
...
+        args = {}
+        self.execute(area_uuid, action, args)
+
+    def device_info(self, area_uuid):
+        area_uuid = area_uuid
+        action = Commands.DEVICE_INFO
+        args = {}
+        self.execute(area_uuid, action, args)
+
+    def execute(self, area_uuid, action, args):
```
spyrostz, line 57: Should be renamed to `_add_to_buffer` or `_store`. This method is used in order to save the pending commands to the buffer, not to execute them

d3a_api_client/commands.py
```diff
@@ -0,0 +1,72 @@
...
+class ClientCommandBuffer:
+
+    def __init__(self, ):
+        self._commands_buffer = []
+
+    def offer_energy(self, area_uuid, energy, price):
+        area_uuid = area_uuid
+        args = {energy, price}
+        action = Commands.OFFER
+        self.execute(area_uuid, action, args)
```
spyrostz, line 13: The 3 separate variables are redundant here. The method will be easy to read even if you use the arguments directly: `self.execute(area_uuid, Commands.OFFER, {energy, price})`.

d3a_api_client/commands.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def offer_energy_rate(self, area_uuid, energy, rate):
+        area_uuid = area_uuid
+        args = {energy, rate}
```
spyrostz, line 17: IIRC these arguments should not be a set but a dict. You need to construct the argument dict, for instance https://github.com/gridsingularity/d3a-api-client/blob/master/d3a_api_client/rest_device.py#L87

spyrostz (commented):
Some comments, and a general comment that was discussed with @hannesdiedrich that am integration test would be great to have here

d3a_api_client/commands.py
spyrostz: That looks good, thanks! I have left some other comments for renaming the methods to correctly indicate their intent.

tests/test_create_aggregator.py
```diff
@@ -27,27 +27,20 @@ def on_market_cycle(self, market_info):
...
-
         for device_event in market_info["content"]:
             if "available_energy_kWh" in device_event["device_info"] and \
                     device_event["device_info"]["available_energy_kWh"] > 0.0:
-                batch_commands[device_event["area_uuid"]] = [
-                    {"type": "offer",
-                     "price": 1,
-                     "energy": device_event["device_info"]["available_energy_kWh"] / 2},
-                    {"type": "list_offers"}]
+                self.add_to_batch_commands.offer_energy(device_event["area_uuid"], price=1,
```
Hassan754, line 33: @Ravineesh , I adapted this test to use the new interface.

d3a_api_client/commands.py
```diff
@@ -0,0 +1,69 @@
...
+    def __init__(self, ):
+        self._commands_buffer = []
+
+    @property
+    def buffer_length(self):
+        return len(self._commands_buffer)
+
+    def offer_energy(self, area_uuid, energy, price):
+        self._add_to_buffer(area_uuid, Commands.OFFER, {"energy": energy, "price": price})
+        return self
```
spyrostz, line 15: In order to avoid repeating `return self` in all these methods, you could `return self` at the end of `_add_to_buffer` method, and `return self._add_to_buffer()`.

spyrostz (approved):
One comment, LGTM

hannesdiedrich (dismissed):
LGTM

hannesdiedrich (approved):
LGTM

spyrostz (approved):
LGTM

PR #82, opened 2021-01-25
feature/D3ASIM-3204: Enabled cli options in the Api-client entry point

Hassan754:
### PR Details
Added an entry point for the api client (In setup) targeting the `cli.main` which adds support to passing extra options to the command
Entry point is now `d3a-api-client run --setup MODULE_NAME ...
Added the feature in README
Moved the Tests folder inside the app directory


cli.py
```diff
@@ -0,0 +1,64 @@
...
+    root_logger.addHandler(handler)
+
+
+@main.command()
+@click.option('--setup', 'setup_module_name', default="default_2a",
+              help="Simulation setup module use]")
+def run(setup_module_name, **kwargs):
+
+    try:
+        exec(open(f"./{setup_module_name}").read())
```
Hassan754, line 61: I'm not sure this is how the entry point should run the desired file, open for any other solutions

faizan2590: I think better approach would be to use current path as default in addition to adding a CLI parameter referring to path where user's script is.

hannesdiedrich: I partly agree with Faizan, IMO setting the right path to the setup file should be left to the user.

cli.py
```diff
@@ -0,0 +1,64 @@
...
+            datefmt="%H:%M:%S"
+        )
+    )
+    root_logger = logging.getLogger()
+    root_logger.setLevel(log_level)
+    root_logger.addHandler(handler)
+
+
+@main.command()
+@click.option('--setup', 'setup_module_name', default="default_2a",
```
hannesdiedrich, line 56: I was really confused by the name `setup_module_name`, because we use this in the d3a for the setup for running the simulation. But AFAIU, you are expecting the downloaded collaboration file here, right?

hannesdiedrich: So, I suggest to rename this to `setup_file` or similar

cli.py
spyrostz: Even better, not a path but a Python module. d3a --setup parameter does not accept filesystem paths, but Python modules, which makes it far easier to import the desired setup file.

cli.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+log = getLogger(__name__)
+
+
+@click.group(name='d3a-api-client', cls=DefaultGroup, default='run', default_if_no_args=True,
+             context_settings={'max_content_width': 120})
+@click.option('-l', '--log-level', type=Choice(list(logging._nameToLevel.keys())), default='INFO',
+              show_default=True, help="Log level")
+def main(log_level):
```
hannesdiedrich, line 40: Please add this new feature to README.md. Thanks!

hannesdiedrich (commented):
Left some comments, will review again once addressed.

cli.py
```diff
@@ -0,0 +1,64 @@
...
+        )
+    )
+    root_logger = logging.getLogger()
+    root_logger.setLevel(log_level)
+    root_logger.addHandler(handler)
+
+
+@main.command()
+@click.option('--setup', 'setup_module_name', default="default_2a",
+              help="Simulation setup module use]")
```
faizan2590, line 57:
```suggestion
              help="setup module of client script]")
```

cli.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+_setup_modules = available_client_setups
+
+
+@main.command()
+@click.option('--setup', 'module_name', default="auto_offer_bid_on_device",
+              help="Setup module of client script. Available modules: [{}]".format(
+                  ', '.join(_setup_modules)))
+def run(module_name, **kwargs):
```
faizan2590, line 64: Clients scripts that are scripts that are inside this repo are working via `d3a-api-client run --setup auto_offer_bid_on_device` thanks for that but can you guide me how to run the client scripts that are outside of this repo?

faizan2590 (commented):
Tested locally via cli, looks good though left one more comment and will approve once addressed/answered.

cli.py
Hassan754: @faizan2590 this was in response to [Spyros comment](https://github.com/gridsingularity/d3a-api-client/pull/82#discussion_r564336361), I can surely add this functionality if it is intended @spyrostz

spyrostz: @faizan2590 @Hassan754 I personally for sure wanted the ability to import it as a Python package, because the research team is accustomed to use it like this (because they are used to the d3a behavior). In order to support client scripts that are not part of this repo, this can be done using an extra command line argument (like --setup-file-path) that will navigate to the path of the setup file. Not a fan of this method, due to the fact that one cannot present a list of potential setup files (like the other method is able to), however @faizan2590 if you insist in order to add flexibility to the user then let's add it with an additional argument.

README.md
```diff
@@ -203,3 +204,7 @@ device_client.set_energy_forecast(<pv_energy_forecast_Wh>)
 ```
 
 ---
+## Interacting with the CLI
+You can run one of the setup scripts residing in the `d3a_api_client.tests` folder from the terminal.
```
hannesdiedrich, line 208:
```suggestion
You can run one of the setup scripts residing in the `d3a_api_client/setups` folder from the terminal.
```

README.md
```diff
@@ -203,3 +204,7 @@ device_client.set_energy_forecast(<pv_energy_forecast_Wh>)
...
+In order to run a module you can use the following command
+`d3a-api-client run --setup {module_name}`
```
hannesdiedrich, line 210: @Ravineesh Please be aware of this change of how to run the api-client (once it is pushed). Thanks!

cli.py
faizan2590: Screen Shot 2021-01-27 at 13 35 20
What I meant was that just like `d3a` user can set the base path where all its respective client script could be found and can be imported by `d3a-api-client` like a python module

hannesdiedrich (approved):
One last minor comment. Once addressed, LGTM

cli.py
spyrostz: Got it now, and I agree, this is indeed helpful

d3a_api_client/utils.py
```diff
@@ -220,3 +223,17 @@ def execute_function_util(function: callable, function_name):
...
+
+def iterate_over_all_client_setup():
+    module_list = []
+    client_modules_path = setups.__path__
+    for loader, module_name, is_pkg in walk_packages(client_modules_path):
+        if is_pkg:
+            loader.find_module(module_name).load_module(module_name)
+        else:
+            module_list.append(module_name)
+    return module_list
```
spyrostz, line 236: How about instead of copying this function, to move it from d3a repository to d3a-interface ? That way you will be able to reuse it in both d3a and d3a-api-client repositories. I think you also need to add one extra function argument for the modules path, but it is fairly straightforward to do so

README.md
Ravineesh: This is for running script locally, right? 

For Cypress (inside docker): 
image
where $imageName = d3a_api_client
Do I have to change this also ?

Hassan754: @Ravineesh Yes, I will keep you updated when I push the changes so You can adapt and test accordingly

hannesdiedrich: ```where $imageName = d3a_api_client```
No, as the images is build each time you start the tests, no need to change the image. Only the way of starting the api script.

Hassan754: I have changed the predefined setups folder @Ravineesh

Ravineesh: ok, Thanks

faizan2590 (approved):

spyrostz (approved):
LGTM, just make sure that the `tests` directory has been completely removed (reason I am asking about it is because there is a `tests/__init__.py` file at the end of this diff, which I think should not be there since the directory has been renamed to `setups`)

Hassan754:
> LGTM, just make sure that the `tests` directory has been completely removed (reason I am asking about it is because there is a `tests/__init__.py` file at the end of this diff, which I think should not be there since the directory has been renamed to `setups`)

@spyrostz this is intended, as when running integration tests, the d3a image will be pulled inside it.

spyrostz:
> > LGTM, just make sure that the `tests` directory has been completely removed (reason I am asking about it is because there is a `tests/__init__.py` file at the end of this diff, which I think should not be there since the directory has been renamed to `setups`)
> 
> @spyrostz this is intended, as when running integration tests, the d3a image will be pulled inside it.

Thanks for clarifying that!

PR #84, opened 2021-01-26
refactor_TESTS_52

Ravineesh:
Adding the domain_name and websockets_domain_name as an argument. So, the trading script will not be dependent on particular testing enviroment


Hassan754:
@spyrostz Aren't we going to add these arguments from the cli in the current story ?

Hassan754 (approved):

hannesdiedrich (approved):
LGTM

hannesdiedrich:
> @spyrostz Aren't we going to add these arguments from the cli in the current story ?

AFAIU, this is for enabling fast switching between `dev` and `staging` for running the E2E tests on  by just setting an env variable.

PR #89, opened 2021-01-29
D3ASIM-3220: Add replace_existing flag on bids and offers

fievelk:
- Add `replace_existing` flag on the bid/offer methods
- reduce duration of integration tests (approximately by 1/2)

## Related PRs

- https://github.com/gridsingularity/d3a/pull/1038


d3a_api_client/redis_client_base.py
```diff
@@ -177,18 +178,27 @@ def offer_energy_rate(self, energy, rate):
...
+
+        response = self._publish_and_wait(
+            Commands.BID,
+            {"energy": energy, "price": rate * energy, 'replace_existing': replace_existing})
+
         log_bid_offer_confirmation(response)
+
         return response
 
     @registered_connection
```
hannesdiedrich, line 204: How about the `offer_energy` and `offer_energy_rate` ?

hannesdiedrich (commented):
Left one comment. Once addressed and once an integrationtest targetting the `replace_existing` functionality is added, LGTM

d3a_api_client/redis_client_base.py
fievelk: I had planned to create a separate PR for `offer_energy` and `offer_energy_rate` (as I am doing on d3a). Would you rather have them together? In that case, this PR would need to wait before getting merged, until both the "bids" PR and the future "offers" PR will be merged on d3a.

hannesdiedrich: yes, please add both, as stated in the tickets title.
Also, as stated in d3a, you can test with the corresponding d3a branch without having to push it to master first.
Just adapt the dockerfile for the integrationtests. If you need more help, let me know. Thanks!

d3a_api_client/rest_device.py
```diff
@@ -83,34 +84,42 @@ def set_energy_forecast(self, pv_energy_forecast_Wh, do_not_wait=False):
...
             log_bid_offer_confirmation(response)
             return response
 
     @logging_decorator('offer')
-    def offer_energy_rate(self, energy, rate):
+    def offer_energy_rate(self, energy: float, rate: float, replace_existing: bool = True):
         transaction_id, posted = self._post_request(
-            'offer', {"energy": energy, "price": rate * energy})
+            'offer', {
+                "energy": energy, "price": rate * energy, 'replace_existing': replace_existing})
```
hannesdiedrich, line 100: could you please use this opportunity to homogenise the string quotes to `'`? Thanks!

fievelk: Done!

integration_tests/test_load_connection.py
```diff
@@ -38,11 +39,13 @@ def on_market_cycle(self, market_info):
                 # Validate that the bid was deleted from the market
                 empty_listing = self.list_bids()
                 assert not any(b for b in empty_listing["bid_list"] if b["id"] == bid_info["id"])
+
                 # Place the bid with a rate that will be acceptable for trading
-                bid = self.bid_energy_rate(energy_requirement, 33)
+                bid = self.bid_energy_rate(energy_requirement, 33, replace_existing=False)
```
hannesdiedrich, line 44: Would be great to also test `replace_existing = True` in this or an additional integrationtest. 
Or do you think the unittests on the d3a are sufficient?

d3a_api_client/utils.py
```diff
@@ -6,11 +6,13 @@
...
 import uuid
 from functools import wraps
-from tabulate import tabulate
+
 from d3a_interface.utils import key_in_dict_and_not_none, get_area_name_uuid_mapping,RepeatingTimer
 from d3a_interface.constants_limits import JWT_TOKEN_EXPIRY_IN_SECS
 from d3a_api_client.constants import DEFAULT_DOMAIN_NAME, DEFAULT_WEBSOCKET_DOMAIN
 
+from tabulate import tabulate
+
```
hannesdiedrich, line 15: Please revert to follow PEP8

hannesdiedrich (approved):
Some last minor comments, LGTM once addressed.

integration_tests/test_load_connection.py
fievelk: I'm actually having a difficult time wrapping my head around d3a-api-client integration tests for some reasons. 

- I tried to test `replace_existing=True` by checking the number of bids available in `self.list_bids()`, but I have the feeling that the speed of the simulation is hindering this test. The list was almost always empty and I couldn't find an easy way to stop the process and debug it. Adding print statements to debug takes too long because of the running times, unfortunately.
- What the d3a-api-client integration tests are doing is running a simulation and applying a new strategy on it, using assertions to check that nothing breaks. In my opinion, this is not very stable nor scalable in the long term. I tried to create a new integration test and it would increase the running time of the test suite by 25 minutes, so I avoided it. The possibility of stopping a test after a few market slots isn't currently implemented I think. But most of all, the synchronization between the d3a simulation and the client requests can only be obtained by reducing the simulation speed, and it's not exactly a deterministic process. That's why I suggested we could try to implement a way to advance the simulation one slot at a time so that we could check its current state and make requests during that slot. We had a conversation on Slack a week ago where I suggested something like:
```python
d3_simulation.run_next_cycle()  # This could be a subprocess command for d3a
do_something()
assert something_about_t1  

d3_simulation.run_next_cycle()
do_something()
assert something_else_about_t2
```
I think @spyrostz wanted to create a new task on Jira for this, but I didn't check yet.

hannesdiedrich: @fievelk I generally agree, most of the time it is not needed that the simulation run a whole day (default duration=24h) for testing a single issue. How about changing the duration parameter in your "new" step function here: https://github.com/gridsingularity/d3a-api-client/pull/89/files#diff-bc44b0a5290e14efb9ff70fa0fbf4099cfda89eed3f5541781144eaa4bb070ea 
I also did something similar in my PR (and will adapt to your version once you pushed): https://github.com/gridsingularity/d3a-api-client/pull/97/files#diff-bc44b0a5290e14efb9ff70fa0fbf4099cfda89eed3f5541781144eaa4bb070ea
Afterwards we can go the proposed alternative way in the frame of the `TESTS-75` ticket.

For the empty bid_list I have s suspicion: They get traded right away. So you could suppress this by setting the price for the bid really low. This way these bids will stay in the market longer to be displayed in the list_bids. 

Let me know if you need more help.

d3a_api_client/utils.py
fievelk: Yeah, I don't know why I thought it was in the wrong place :D Fixed that and the other imports' sorting.

integration_tests/test_load_connection.py
fievelk: I tried changing the `--duration` parameter, but it had no effect. I think it is because of the two following chunks:

https://github.com/gridsingularity/d3a-api-client/blob/d9db6002378564f6d18d58ad9f1cbe9cfdcc0f20/integration_tests/steps/device.py#L92-L100

https://github.com/gridsingularity/d3a-api-client/blob/83c31965bb17f1a2033c608f129732c26ac545ab/integration_tests/test_load_connection.py#L59-L61

If I got it right, the `"finished"` flag will be set only when the market time reaches `23:00`. So I'm not sure the `duration` flag (and the actual duration of the simulation) is considered. But it seems odd, so I might be wrong.

hannesdiedrich: Ah yes, this is bad anyway, the `self.status = "finished" ` should only be changed when `on_finished` is triggered. So you need to add `on_finished` to the `AutoBidOnLoadDevice`  and `AutoBidOnPVDevice` and get rid of the last block.

integration_tests/test_aggregator_batch_commands.py
```diff
@@ -28,31 +30,146 @@ def _setup(self):
...
+                    ).offer_energy(
+                        area_uuid=device_event['area_uuid'],
+                        price=1,
+                        energy=device_event['device_info']['available_energy_kWh'] / 4,
+                        replace_existing=True
+                    ).offer_energy(
+                        area_uuid=device_event['area_uuid'],
+                        price=1,
+                        energy=device_event['device_info']['available_energy_kWh'] / 4,
+                        replace_existing=False
```
faizan2590, line 61: Rather than replacing bids with identical price and energy is to have different and validate in integration test that the last one was traded

fievelk: This is what I was trying to do [here](https://github.com/gridsingularity/d3a-api-client/pull/89/files/a5afed4a333d0c68d3a112fe4eef2e9b91f12370#diff-69601bf521aa6db62641cf28c676950ab0bd31b13024830a70d0c37b1e0df234R115-R117). But you might be right, manually checking the values could actually be more useful.

hannesdiedrich (dismissed):
Thanks a lot! LGTM

d3a_api_client/utils.py
```diff
@@ -1,15 +1,18 @@
...
+import traceback
 import uuid
 from functools import wraps
+
 from tabulate import tabulate
-from d3a_interface.utils import key_in_dict_and_not_none, get_area_name_uuid_mapping,RepeatingTimer
+
 from d3a_interface.constants_limits import JWT_TOKEN_EXPIRY_IN_SECS
 from d3a_api_client.constants import DEFAULT_DOMAIN_NAME, DEFAULT_WEBSOCKET_DOMAIN
+from d3a_interface.utils import get_area_name_uuid_mapping, key_in_dict_and_not_none, \
```
spyrostz, line 14: Minor, but better to move the d3a_interface imports above the d3a_api_client imports. In general it is much more preferrable to have the local imports last.

spyrostz (dismissed):
LGTM

integration_tests/test_aggregator_batch_commands.py
fievelk: Done!

d3a_api_client/utils.py
fievelk: True, I wasn't considering `d3a_interface` as a 3rd party, but that's actually what it is wrt this repo 👍

faizan2590 (dismissed):
Before merging this PR, please adapt one of the existing client scripts for users to use as a reference(potentially using [test_redis_aggregator](https://github.com/gridsingularity/d3a-api-client/blob/master/d3a_api_client/setups/test_redis_aggregator.py) and adding a `replacing_existing` flag on bids/offers at `on_tick`)

Hassan754 (dismissed):

fievelk:
> Before merging this PR, please adapt one of the existing client scripts for users to use as a reference(potentially using test_redis_aggregator and adding a replacing_existing flag on bids/offers at on_tick)

I will work on this. Please note that the `replace_existing` flag is `True` by default, so every existing script will be using it.

d3a_api_client/commands.py
```diff
@@ -13,20 +13,40 @@ def __init__(self, ):
...
 
-    def bid_energy(self, area_uuid, energy, price):
-        return self._add_to_buffer(area_uuid, Commands.BID, {"energy": energy, "price": price})
+    def bid_energy(
+            self, area_uuid: str, energy: float, price: float, replace_existing: bool = True):
+
+        return self._add_to_buffer(
+            area_uuid,
+            Commands.BID,
+            {'energy': energy, 'price': price, 'replace_existing': replace_existing})
```
hannesdiedrich, line 41: Just a minor question. Is this refactoring following a best practice layout?
When I would have implemented it, adding the `replace_existing` parameter would have added one line to the `bid_energy`. Now the function has 7 lines and IMO the readability did not improve IMO.
Happy to learn how to do it correctly.

d3a_api_client/redis_client_base.py
```diff
@@ -164,31 +165,46 @@ def _publish_and_wait(self, command_type, data):
...
-    def bid_energy(self, energy, price):
+    def bid_energy(self, energy: float, price: float, replace_existing: bool = True):
         logging.debug(f"{self.area_id}Client tries to place a bid for {energy} kWh at {price} cents.")
-        response = self._publish_and_wait(Commands.BID, {"energy": energy, "price": price})
+
+        response = self._publish_and_wait(
+            Commands.BID,
+            {'energy': energy, 'price': price, 'replace_existing': replace_existing})
         log_bid_offer_confirmation(response)
+
```
hannesdiedrich, line 196: Also here, the additional two blank lines. IMO they dont improve the readability, only the length of the file.

integration_tests/steps/device.py
```diff
@@ -16,20 +16,18 @@ def step_impl(context):
...
-
-
-@given('d3a container started {setup_file} with duration set to {hours} hours')
-def step_impl(context, setup_file, hours):
-    sleep(3)
-    system(f'docker run -d --name d3a-tests --env REDIS_URL=redis://redis.container:6379/ '
-           f'--net integtestnet d3a-tests -l INFO run -t 1s -s 60m -d {hours}h --setup {setup_file} '
-           f'--no-export --seed 0 --enable-external-connection')
+           f'--net integtestnet d3a-tests -l INFO run --setup {setup_file} '
+           f'--no-export --seed 0 --enable-external-connection {d3a_options}')
```
hannesdiedrich, line 30: Great! Thanks a lot!

hannesdiedrich (approved):
LGTM

d3a_api_client/commands.py
fievelk: Do you have an example of how you would have implemented it?
Adding type hints already makes the signature longer than 100 characters. Because of that, I added a new line after `def bid_energy(`.
The same applies to the parameters for `self._add_to_buffer`: they were longer than 100 characters, so I had to break them somehow. Since the last parameter (the dict) takes almost one line, I decided to allocate one single line to each parameter because I thought it looked clearer. 
There are lots of ways to break continuation lines though; I usually use [hanging indents](https://www.python.org/dev/peps/pep-0008/#indentation) when I can, and let use my personal taste when I'm in doubt.

hannesdiedrich: Having read the link to the corresponding paragraph in PEP8 you posted, I agree with your way of implementing (my proposal would have been one of the solutions under the `# Wrong:` part :-D)

d3a_api_client/redis_client_base.py
fievelk: I tend to break code into logically-related parts (see: [PEP 8 - Blank Lines](https://www.python.org/dev/peps/pep-0008/#blank-lines))
> Use blank lines in functions, sparingly, to indicate logical sections.

Of course, the _sparingly_ can be interpreted :D

In this case, I've "seen" three parts:
- initial part for logging (before anything is done)
- the actual logic of the method and its logging
- the `return` statement (I generally add a blank line before return statements, unless they're one-liners).
 
This helps me to clearly identify the different parts of a method and to visually understand when a chunk of code has "done its part". I understand your concerns about the length of the file, but I prefer the risk of being unnecessarily verbose to the risk of being overly cryptic (not in this case, but as a general rule). It's very subjective though.

On a separate note: if the file has too many lines, most of the time it's because of a design issue (for example, two classes are needed instead of one).

hannesdiedrich: you are of course right, especially with the latter comment.
If we would have time i would have proposed to add a refactoring story for this for each repo we have, but we dont, so yeah, will keep your explanations in mind from now on. Thanks a lot for bearing with me!

fievelk: Of course, there's a time/quality balance to keep in mind, and code quality can/should sometimes be sacrificed depending on the time availability (e.g. deadlines or low resources) :)

faizan2590 (approved):
LGTM

PR #93, opened 2021-02-03
D3ASIM-3213: Added Market batch commands/ Adapted setups

Hassan754:
Added the market Batch commands functions + Adapted the setups using them

### DIFF
--- a/d3a_api_client/setups/test_ess_bid_redis_aggregator.py
+++ b/d3a_api_client/setups/test_ess_bid_redis_aggregator.py
@@ -9,6 +9,7 @@ class AutoAggregator(RedisAggregator):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.is_finished = False
+        self.is_buffer_empty = True
 
     def on_market_cycle(self, market_info):
         logging.info(f"market_info: {market_info}")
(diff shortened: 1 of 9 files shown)


d3a_api_client/setups/test_ess_bid_redis_aggregator.py
```diff
@@ -19,17 +20,15 @@ def on_market_cycle(self, market_info):
...
-            logging.info(f"Batch command placed on the new market")
+                self.add_to_batch_commands.bid_energy(area_uuid=device_event["area_uuid"],
+                                                      price=31 * buy_energy, energy=buy_energy).\
+                    list_bids(area_uuid=device_event["area_uuid"])
+                self.is_buffer_empty = False
+
+        if not self.is_buffer_empty:
+            response = self.execute_batch_commands(batch_commands)
+            self.is_buffer_empty = True
+            logging.info(f"Batch command placed on the new market: {response}")
```
hannesdiedrich, line 31: Can you please explain the reason for this (self.is_buffer_empty)?

Hassan754: This is just to check if the batch commands buffer is empty, anyway if it is empty and we called the execute_batch_commands ,it will not execute anything so no harm in both ways

fievelk: The `RedisAggregator` has a `_client_command_buffer` attribute of type `ClientCommandBuffer`. This class comes with a `buffer_length` property: what would you think of using that to check whether the buffer is empty or not?

https://github.com/gridsingularity/d3a-api-client/blob/e8c51844decf376c93a037f7aea636a27a71a2da/d3a_api_client/commands.py#L9-L11

Maybe you could even define it as a property shortcut on the `RedisAggregator`.

spyrostz (dismissed):
LGTM

d3a_api_client/setups/test_ess_bid_redis_aggregator.py
hannesdiedrich: This is the user facing side, so I suggest to remove the check if it is not needed, Otherwise there will be questions.

hannesdiedrich (approved):
LGTM

faizan2590 (approved):
LGTM

PR #94, opened 2021-02-03
D3ASIM-3218: Added command line argument for json file to be imported.

faizan2590:


cli.py
```diff
@@ -68,18 +69,26 @@ def main(log_level):
...
+def run(setup_module_name, username, password, domain_name, web_socket, json_file_name,
+        run_on_redis, **kwargs):
     if username is not None:
         os.environ["API_CLIENT_USERNAME"] = username
     if password is not None:
         os.environ["API_CLIENT_PASSWORD"] = password
     os.environ["API_CLIENT_DOMAIN_NAME"] = domain_name
     os.environ["API_CLIENT_WEBSOCKET_DOMAIN_NAME"] = web_socket
     os.environ["API_CLIENT_RUN_ON_REDIS"] = "true" if run_on_redis else "false"
+    os.environ["JSON_FILE_PATH"] = os.path.join(modules_path[0], json_file_name)
```
fievelk, line 84: This means that the path of the JSON file will always be relative to `modules_path[0]` (that in general is the `setups/` folder). Is this the intended behavior?

cli.py
```diff
@@ -68,18 +69,26 @@ def main(log_level):
...
 
     try:
-        importlib.import_module(f"d3a_api_client.setups.{setup_module_name}")
+        if SETUP_FILE_PATH:
+            sys.path.append(SETUP_FILE_PATH)
+            importlib.import_module(f"{setup_module_name}")
+        else:
+            importlib.import_module(f"d3a_api_client.setups.{setup_module_name}")
```
hannesdiedrich, line 91: I dont see the SETUP_FILE_PATH set somewhere neither a way the user could set it. Can you help me please?

d3a_api_client/setups/test_create_aggregator.py
```diff
@@ -53,8 +56,11 @@ def on_finish(self, finish_info):
...
 
-import os
-simulation_id = os.environ["API_CLIENT_SIMULATION_ID"]
+with open(os.environ['JSON_FILE_PATH']) as json_file:
+    config_info = json.load(json_file)
+
+
+simulation_id = config_info['uuid']
 domain_name = os.environ["API_CLIENT_DOMAIN_NAME"]
 websocket_domain_name = os.environ["API_CLIENT_WEBSOCKET_DOMAIN_NAME"]
```
hannesdiedrich, line 65: I think the goal is here to make the life easier for the user, so i suggest to handle this inside the `Aggregator` class and make the simulation_id, domain_name, websocket_domain_name optional parameters. If there are not set, the JSON file is read.

hannesdiedrich (commented):
Left two comments, will review again once addressed.

cli.py
```diff
@@ -68,18 +69,26 @@ def main(log_level):
               type=str, help="D3A domain name")
 @click.option('-w', '--web-socket', default=DEFAULT_WEBSOCKET_DOMAIN,
               type=str, help="D3A websocket URL")
+@click.option('-j', '--json-file-name', type=str,
```
fievelk, line 72: I suggest using a name that would give information about the content of the file, rather than its format. Something like `--simulation-info` (or anything better, I can't think of anything else at the moment).

cli.py
faizan2590: For now it could be set via user and found in [constants](https://github.com/gridsingularity/d3a-api-client/blob/master/d3a_api_client/constants.py#L7)

fievelk: > I dont see the SETUP_FILE_PATH set somewhere neither a way the user could set it. Can you help me please?

In situations like this I like to use namespace imports rather than direct imports:
```python
# Direct import
from d3a_api_client.constants import SETUP_FILE_PATH

# Namespace import
from d3a_api_client import constants
```

So the call would become `constants.SETUP_FILE_PATH`, which makes it immediately obvious that the object comes from a different module (and that it is a constant, given the name of that module).

faizan2590: The purpose of having a CLI tool in my perspective is to relief user for the need to go inside `api-client` library code and  had to overwrite any constant settings.  I will add a flag for api-client to read client's scripts other than the default setup path.

d3a_api_client/setups/test_create_aggregator.py
faizan2590: I see the user would be reading this json file containing relevant simulation-info as they themselves have to accordingly get their script connected for their approved area

cli.py
faizan2590: Will adapt it to make it absolute as well as relative to user's current working dir.

d3a_api_client/setups/test_create_aggregator.py
hannesdiedrich: Sorry, I thought this is the file that the user downloads from the d3a-web.
And it was designed so the user does not have to set all these parameters him/herself but only to read from this JSON file. So AFAIU, you need to read more parameters from there.

cli.py
```diff
@@ -68,18 +71,35 @@ def main(log_level):
...
+def run(base_setup_path, setup_module_name, username, password, domain_name, web_socket,
+        simulation_info, run_on_redis, **kwargs):
     if username is not None:
         os.environ["API_CLIENT_USERNAME"] = username
     if password is not None:
         os.environ["API_CLIENT_PASSWORD"] = password
     os.environ["API_CLIENT_DOMAIN_NAME"] = domain_name
     os.environ["API_CLIENT_WEBSOCKET_DOMAIN_NAME"] = web_socket
     os.environ["API_CLIENT_RUN_ON_REDIS"] = "true" if run_on_redis else "false"
+    if os.path.isabs(simulation_info):
```
fievelk, line 87: A very small note on this: if you want, you could use the [`pathlib`](https://docs.python.org/3.8/library/pathlib.html) module to replace `os.path`. I think it is a bit cleaner and I have the feeling that's the direction in which the python community has been going for a while. (For example: [Why you should be using pathlib](https://treyhunner.com/2018/12/why-you-should-be-using-pathlib)).
Not a big deal anyway.

cli.py
```diff
@@ -68,18 +71,35 @@ def main(log_level):
...
     try:
-        importlib.import_module(f"d3a_api_client.setups.{setup_module_name}")
+        if base_setup_path is None:
+            importlib.import_module(f"d3a_api_client.setups.{setup_module_name}")
+        elif base_setup_path and os.path.isabs(base_setup_path):
+            sys.path.append(base_setup_path)
+            importlib.import_module(setup_module_name)
+        else:
+            setup_file_path = os.path.join(os.getcwd(), base_setup_path)
+            # print(f"pathhh: {pathhh}")
```
fievelk, line 100: I suppose this can be safely removed 😄

cli.py
```diff
@@ -59,6 +60,8 @@ def main(log_level):
 
 
 @main.command()
+@click.option('-b', 'base_setup_path', default=None, type=str,
+              help="Accept absolute or relative path for client script")
```
fievelk, line 64: Since you used this for the `--simulation-info`:
> Simulation File info (accept absolute and relative path)
it would be good to also have something like
> Client script (accepts absolute and relative path)

Also, I think it would be nice to add a description of the `base_setup_path` and `simulation_info` usage in the README, since the difference is not immediately clear from the cli arguments and help strings.

cli.py
```diff
@@ -68,18 +71,35 @@ def main(log_level):
...
+        if base_setup_path is None:
+            importlib.import_module(f"d3a_api_client.setups.{setup_module_name}")
+        elif base_setup_path and os.path.isabs(base_setup_path):
+            sys.path.append(base_setup_path)
+            importlib.import_module(setup_module_name)
+        else:
+            setup_file_path = os.path.join(os.getcwd(), base_setup_path)
+            # print(f"pathhh: {pathhh}")
+            sys.path.append(setup_file_path)
+            importlib.import_module(setup_module_name)
```
fievelk, line 102: It could happen that the relative path for `setup_file_path` doesn't lead anywhere (it doesn't exist). In that case, a `ModuleNotFoundError` would be raised.
It would be good to catch this and show an explicative error to the user to tell them what went wrong.

d3a_api_client/setups/test_create_aggregator.py
fievelk: I also understood from the task description that this information (like the websocket name and domain name) would be provided inside of the JSON file. So that information should probably overwrite the one contained in the `os.environ` dictionary.

fievelk (commented):
Left some comments. Most of them are minor, but there's still [something to check regarding env variables](https://github.com/gridsingularity/d3a-api-client/pull/94#discussion_r572943307).

d3a_api_client/setups/test_create_aggregator.py
faizan2590: Yes, it is even now possible. I will leave it in this as an example.

cli.py
faizan2590: Thanks for the suggestion, will add it.

cli.py
faizan2590: Sure, will do once finalised.

cli.py
faizan2590: @spyrostz @hannesdiedrich Would you guys prefers pathlib. Im indifferent actually

cli.py
faizan2590: https://github.com/gridsingularity/d3a-api-client/pull/94/files#diff-7648f5f4cce4bf01d02de43f8c5f6f2130b787baece0cf6338210496d8bdfcdcR106-R107

d3a_api_client/setups/test_create_aggregator.py
```diff
@@ -28,13 +29,13 @@ def on_market_cycle(self, market_info):
             return
 
         for device_event in market_info["content"]:
-            if "available_energy_kWh" in device_event["device_info"] and \
+            if key_in_dict_and_not_none(device_event, "available_energy_kWh") and \
```
Hassan754, line 32: It would be nice if you can separate the checks into private functions like [This one](https://github.com/gridsingularity/d3a-api-client/blob/80a074c5e3543639a69afa925f7c9a3d7797d154/integration_tests/test_aggregator_batch_commands.py#L198)

d3a_api_client/constants.py
```diff
@@ -6,5 +6,5 @@
 
 SETUP_FILE_PATH = None
 
-DEFAULT_DOMAIN_NAME = 'https://d3aweb.gridsingularity.com'
-DEFAULT_WEBSOCKET_DOMAIN = 'wss://d3aweb.gridsingularity.com/external-ws'
+DEFAULT_DOMAIN_NAME = 'http://localhost:8000'
```
Hassan754, line 9: Unless this is necessary, please revert these before merging

d3a_api_client/rest_device.py
```diff
@@ -15,20 +18,28 @@
...
+            with open(os.environ['JSON_FILE_PATH']) as json_file:
+                simulation_info = json.load(json_file)
+                self.simulation_id = simulation_info['uuid']
+                self.domain_name = simulation_info['domain_name']
+                self.websockets_domain_name = simulation_info['web_socket_domain_name']
+        else:
+            self.simulation_id = simulation_id
+            self.domain_name = domain_name
+            self.websockets_domain_name = websockets_domain_name
+
```
Hassan754, line 36: If possible can you move this logic into its own private function to keep the __init__ clean

d3a_api_client/setups/test_create_aggregator.py
```diff
@@ -53,62 +54,51 @@ def on_finish(self, finish_info):
...
+if os.environ['JSON_FILE_PATH'] is not None:
+    with open(os.environ['JSON_FILE_PATH']) as json_file:
+        simulation_info = json.load(json_file)
+        simulation_id = simulation_info['uuid']
+        domain_name = simulation_info['domain_name']
+        websockets_domain_name = simulation_info['web_socket_domain_name']
+else:
+    simulation_id = os.environ["API_CLIENT_SIMULATION_ID"]
+    domain_name = os.environ["API_CLIENT_DOMAIN_NAME"]
+    websockets_domain_name = os.environ["API_CLIENT_WEBSOCKET_DOMAIN_NAME"]
```
Hassan754, line 66: I see this is used again, maybe we can create a function that sets these,  a function that returns tuple maybe
```python
simulation_id, domain_name, websockets_domain_name = some_function()
```

Hassan754 (commented):
Some enhancements suggestions, LGTM

d3a_api_client/setups/test_create_aggregator.py
Hassan754: If the number of variables may scale to more than 3, we should use some other way

d3a_api_client/redis_aggregator.py
```diff
@@ -44,13 +44,13 @@ def _connect_to_simulation(self, is_blocking=True):
...
         event_channel = f'external-aggregator/*/{self.aggregator_uuid}/events/all'
-        channel_dict = {"aggregator_response": self._aggregator_response_callback,
-                        event_channel: self._events_callback_dict,
+        channel_dict = {event_channel: self._events_callback_dict,
                         f"external-aggregator/*/{self.aggregator_uuid}/response/batch_commands":
                             self._batch_response,
                         }
 
         self.pubsub.psubscribe(**channel_dict)
+        self.pubsub.subscribe(**{"aggregator_response": self._aggregator_response_callback})
```
hannesdiedrich, line 53: Any reason for that? ´psubscribe´ can also handle non-patterns.

faizan2590: I'm not sure why I wasn'T getting response callback while being part of `psubscribe` while debugging issue reported [here](https://gridsingularity.atlassian.net/browse/D3ASIM-3203?focusedCommentId=21970)

cli.py
```diff
@@ -68,20 +71,54 @@ def main(log_level):
...
+
+
+def set_json_file_env(base_setup_path, simulation_info, run_on_redis):
+    if run_on_redis is True:
+        os.environ["JSON_FILE_PATH"] = ""
+        return
+    if simulation_info is None:
+        if base_setup_path is None:
+            json_file = [l for l in os.listdir(setups.__path__[0]) if l.endswith('.json')]
+            os.environ["JSON_FILE_PATH"] = os.path.join(setups.__path__[0], json_file[0])
```
hannesdiedrich, line 117: So if i dont provide a simulation_info_file, the first `.json` file in the setup file path is taken?
Very risky IMO. There could be other json files that dont follow the simulation_info_file pattern.
IMO in this case, nothing should be done and all settings should be read from the environment.

cli.py
```diff
@@ -68,20 +71,54 @@ def main(log_level):
...
+        else:
+            json_file = [l for l in os.listdir(base_setup_path) if l.endswith('.json')]
+            os.environ["JSON_FILE_PATH"] = os.path.join(base_setup_path, json_file[0])
```
hannesdiedrich, line 120: So if i dont provide a simulation_info_file, the first .json file in the setup file path is taken?
Very risky IMO. There could be other json files that dont follow the simulation_info_file pattern.
IMO in this case, nothing should be done and all settings should be read from the environment.

d3a_api_client/setups/test_create_aggregator.py
faizan2590: No need, there is already a `key_in_dict_and_not_none_and_greater_than_zero` in `d3a-interface`, will re-use it

hannesdiedrich (commented):
Left some comments, will review again

d3a_api_client/setups/test_create_aggregator.py
```diff
@@ -53,62 +54,51 @@ def on_finish(self, finish_info):
...
-    simulation_id, "House", domain_name)
+# area_uuid = get_area_uuid_from_area_name_and_collaboration_id(
+#     simulation_id, "House", domain_name)
 
-rest_market = RestMarketClient(simulation_id, area_uuid, domain_name, websocket_domain_name)
-market_slot_string = today().add(minutes=60).format(DATE_TIME_FORMAT)
-last_market_stats = rest_market.last_market_stats()
+# rest_market = RestMarketClient(simulation_id, area_uuid, domain_name, websocket_domain_name)
+# market_slot_string = today().add(minutes=60).format(DATE_TIME_FORMAT)
+# last_market_stats = rest_market.last_market_stats()
```
fievelk, line 101: Leftover comments :)

cli.py
```diff
@@ -68,20 +71,54 @@ def main(log_level):
...
+            sys.path.append(setup_file_path)
+            importlib.import_module(setup_module_name)
 
     except D3AException as ex:
         raise click.BadOptionUsage(ex.args[0])
     except ModuleNotFoundError as ex:
         log.error("Could not find the specified module")
+
+
+def set_json_file_env(base_setup_path, simulation_info, run_on_redis):
```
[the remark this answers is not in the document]
faizan2590: would rename it to `set_simulation_file_env `

Hassan754 (approved):
LGTM

d3a_api_client/setups/test_create_aggregator.py
```diff
@@ -53,60 +60,40 @@ def on_finish(self, finish_info):
...
-device_args["device_id"] = load2_uuid
-
-load2 = RestDeviceClient(
-    **device_args
-)
-
-pv1_uuid = get_area_uuid_from_area_name_and_collaboration_id(
-    device_args["simulation_id"], "PV", device_args["domain_name"])
+pv1_uuid = get_area_uuid_from_area_name_and_collaboration_id(simulation_id, "PV", domain_name)
 device_args["device_id"] = pv1_uuid
```
hannesdiedrich, line 85: This step will be obsolete once D3ASIM-3209 is implemented, right?

d3a_api_client/redis_aggregator.py
hannesdiedrich: Hmm, are you sure that these are connected?

cli.py
```diff
@@ -68,20 +71,50 @@ def main(log_level):
...
 @click.option('-w', '--web-socket', default=DEFAULT_WEBSOCKET_DOMAIN,
               type=str, help="D3A websocket URL")
-@click.option('--run-on-redis', is_flag=True, default=False, help="Start the client using the Redis API")
-def run(setup_module_name, username, password, domain_name, web_socket, run_on_redis, **kwargs):
+@click.option('-i', '--simulation-info', type=str,
+              help="Simulation File info (accept absolute and relative path)")
+@click.option('--run-on-redis', is_flag=True, default=False,
+              help="Start the client using the Redis API")
+def run(base_setup_path, setup_module_name, username, password, domain_name, web_socket,
+        simulation_info, run_on_redis, **kwargs):
```
hannesdiedrich, line 79: Minor, but in order to show the reader that this is a file_path_str, how abour renaming this inputvariable (and consequently the clit parameter) to somwthing like:
`simulation_info_path` or even `simulation_info_file_path?`

d3a_api_client/utils.py
```diff
@@ -268,3 +272,18 @@ def log_bid_offer_confirmation(message):
                          f"{round(energy, 2)} kWh at {price} cts/kWh")
     except Exception as e:
         logging.error(f"Logging bid/offer info failed.{e}")
+
+
+def get_simulation_info(simulation_id=simulation_id_from_env, domain_name=domain_name_from_env,
```
hannesdiedrich, line 277: Please add a small unittest for this function. Thanks!

hannesdiedrich (commented):
Two comments, LGTM once resolved

cli.py
faizan2590: It could be the filename as well as path.

d3a_api_client/setups/test_create_aggregator.py
faizan2590: As far as I see not really as its done from the device owner to get uuid of its device which is being used to `select_aggregator` whereas 3209 is for an aggregator to know devices it controls.

d3a_api_client/redis_aggregator.py
faizan2590: I just tested it again, with pattern subscription wasn't able to get response whereas by exact subscription was getting response accordingly.

cli.py
spyrostz: Even the filename can be considered a 'path' (relative path to the current directory). I would also prefer if this variable was named simulation_info_path (I actually care more about its appearance as a CLI argument rather than in the function argument, I think that `--simulation-info-path` is much more clear than `--simulation-info` for the user of the CLI).

Moving forward to this direction, this is really not info, right ? As far as I understand this is a configuration file (it is used in order to preconfigure the connection to a simulation), therefore it looks to my eyes that a name `--simulation-config-path` might be more appropriate, assuming that I understood the intent of this file correctly.

spyrostz (approved):
One comment, LGTM once resolved.

d3a_api_client/setups/test_create_aggregator.py
hannesdiedrich: Exactly, this is the aggregator-script here. If the aggregator would know the uuids of the devices it controls, this step would be not needed.

d3a_api_client/utils.py
```diff
@@ -276,14 +278,17 @@ def log_bid_offer_confirmation(message):
...
-    if os.environ['JSON_FILE_PATH'] != "":
+    if os.environ.get('JSON_FILE_PATH', "") != "":
         with open(os.environ['JSON_FILE_PATH']) as json_file:
             simulation_info = json.load(json_file)
-            simulation_id = simulation_info['uuid']
-            domain_name = simulation_info['domain_name']
-            websockets_domain_name = simulation_info['web_socket_domain_name']
+        if set(simulation_info.keys()) != {"username", "name", "uuid", "domain_name",
+                                           "web_socket_domain_name", "global_settings", "registry"}:
+            raise SimulationInfoException(f"Simulation_file doesn't have all the keys")
```
hannesdiedrich, line 291: Please be consistent in the naming. Please change 
- `Simulation_file` here in the error message
- `SimulationInfoException`
- `api-client-summary` in the resources folder
to the proposal in Spyros comment (or different, as long as they are the same)
Thanks!

d3a_api_client/utils.py
```diff
@@ -276,14 +278,17 @@ def log_bid_offer_confirmation(message):
...
                         websockets_domain_name=websocket_domain_name_from_env):
-    if os.environ['JSON_FILE_PATH'] != "":
+    if os.environ.get('JSON_FILE_PATH', "") != "":
         with open(os.environ['JSON_FILE_PATH']) as json_file:
             simulation_info = json.load(json_file)
-            simulation_id = simulation_info['uuid']
-            domain_name = simulation_info['domain_name']
-            websockets_domain_name = simulation_info['web_socket_domain_name']
+        if set(simulation_info.keys()) != {"username", "name", "uuid", "domain_name",
+                                           "web_socket_domain_name", "global_settings", "registry"}:
```
hannesdiedrich, line 290: Probably premature, but we will need a validation method that checks for these keys.
This validator should live in the d3a-interface and should be called also before downloading the simulation-config-file. 
But please feel free to skip it. Maybe for some time when we finally settled on the in formations we put in the simulation-config-file.

hannesdiedrich: I just found my old comment 
https://github.com/gridsingularity/d3a-api-client/pull/94#discussion_r596171500
That you didnt reply to. What do you think?

unit_tests/test_simulation_info.py
```diff
@@ -0,0 +1,60 @@
...
+                self.api_client_path,
+                "resources/malformed-api-client-summary-84d221fa-46e1-49e0-823b-26cfb3425a5a.json"
+            )
+            get_simulation_info()
+
+    def test_default_simulation_info_is_under_setup_module_if_base_path_isnt_provided(self):
+        set_simulation_file_env(base_setup_path=None, simulation_info='test.json',
+                                run_on_redis=False)
+        assert os.environ['JSON_FILE_PATH'] == os.path.join(self.api_client_path,
+                                                            'setups/test.json')
```
hannesdiedrich, line 48: I thought we agreed on not supporting this: 
https://github.com/gridsingularity/d3a-api-client/pull/94#discussion_r596166482

d3a_api_client/redis_aggregator.py
hannesdiedrich: I just had a look in the docs and there is no place where they explain that `psubscribe` can only subscribe to patterns. I remember testing this in the past and it worked. We also do in other places.
Please have another look, there might be something else wrong here.

spyrostz, line 53: WRT the discussion above, give this a try, it should work. To complement Hannes' answer from before, we also do this in the `RedisSimulationCommunication` class in d3a-web, therefore it should be possible to do so. 
```suggestion
        channel_dict["aggregator_response"] = self._aggregator_response_callback
        self.pubsub.psubscribe(**channel_dict)
```

hannesdiedrich (commented):
Left some comments

unit_tests/test_simulation_info.py
faizan2590: Its not the same as linked comment, here `simulation_info ` two lines above via `set_simulation_file_env ` is set as as `test.json` and thats what I'm testing if found in `setups` folder (being the default base location of api-client scripts), whereas what you said in the linked comment(which I agreed) was to not simply pick the first `.json` being found in targeted directory.

d3a_api_client/utils.py
faizan2590: Sorry, after reading your comment again, realised that didn't fully adapted ie I did added an assertion logic to compare  keys if the read json is the right format but didn't abstracted it into a validator method, thanks for pointing it again, will abstract into a validation method and make part of `d3a-interface`

unit_tests/test_simulation_info.py
hannesdiedrich: Indeed, sorry, I misunderstood here.

d3a_api_client/setups/test_create_aggregator.py
hannesdiedrich: I see now. You mean for the direct connection to the device we still need it. 
Yes, this starts the discussion again, when we stop maintaining the direct connections. Will move this discussion to the D3ASIM-3365 ticket and marking this here as resolved as it is not related.

cli.py
faizan2590: Added: https://github.com/gridsingularity/d3a-api-client/pull/94/commits/63e7237b5f1b757983047df2f81154ab801aa569

d3a_api_client/utils.py
faizan2590: https://github.com/gridsingularity/d3a-api-client/pull/94/commits/302c1dc2d64d90d29244a6376e97da49878838bc

d3a_api_client/setups/test_create_aggregator.py
faizan2590: > Exactly, this is the aggregator-script here. If the aggregator would know the uuids of the devices it controls, this step would be not needed.

Before that device owner needs to select an aggregator too which is ATM done by the same script.

>  You mean for the direct connection to the device we still need it.

Thats a separate matter, if device owner should directly trade on behalf of their device or always select an aggregator. What I meant was selection of aggregator is done by device owner and ATM its only possible via RestDeviceClient's method `select_aggregator `, and to instantiate `RestDeviceClient ` for an specific device, one needs uuid for respective device.

setup.py
```diff
@@ -6,7 +6,7 @@
         REQUIREMENTS = [r.partition('#')[0] for r in req if not r.startswith('-e')]
         REQUIREMENTS.extend(
             ['d3a-interface @ '
-             f'git+https://github.com/gridsingularity/d3a-interface.git@master'])
+             f'git+https://github.com/gridsingularity/d3a-interface.git@feature/D3ASIM-3218'])
```
faizan2590, line 9: Reminder to remove before merging.

README.md
```diff
@@ -50,35 +50,37 @@ pip install git+https://github.com/gridsingularity/d3a-api-client.git
...
-### Interacting with the CLI
-You can run one of the setup scripts residing in the `d3a_api_client/setups` folder, or you can 
-set the path for your scripts directory as list in the `d3a_api_client.constants` .
-In order to run a module you can use the following command
-```
-d3a-api-client run --setup {module_name}
-```
+### Interacting via CLI
+It helps to trigger your client script interact with your interested d3a's collaboration/CN conveniently via bash terminal.
+In order to get help of supported flag with its minimal explanation, please do the following
```
hannesdiedrich, line 55:
```suggestion
In order to get help, please run:
```

README.md
```diff
@@ -50,35 +50,37 @@ pip install git+https://github.com/gridsingularity/d3a-api-client.git
...
 
-### Interacting with the CLI
-You can run one of the setup scripts residing in the `d3a_api_client/setups` folder, or you can 
-set the path for your scripts directory as list in the `d3a_api_client.constants` .
-In order to run a module you can use the following command
-```
-d3a-api-client run --setup {module_name}
-```
+### Interacting via CLI
+It helps to trigger your client script interact with your interested d3a's collaboration/CN conveniently via bash terminal.
```
hannesdiedrich, line 54: I dont see the information in this sentence.

README.md
```diff
@@ -50,35 +50,37 @@ pip install git+https://github.com/gridsingularity/d3a-api-client.git
...
+```
+Following is the more in depth explanation of individual flags supported,
+- base-setup-path --> Sets the base path where you client script resides else `d3a_api_client/setups` is taken as default for your client-scripts. User could provided both an absolute and relative file path.
+- setup --> Name of your api-client module/script.
+- username --> Username of agent authorized to communicate with respective collaboration/CN.
+- password --> Password of respective agent
+- domain-name --> D3A domain name
+- web-socket --> D3A websocket URI
+- simulation-config-path --> Path of JSON format file that contains agent's point of interest's collaboration/CN's info (It could downloaded from the registry page of UI) 
+- run-on-redis --> This flag targets the local testing of api client, where no user authentication is needed and needs to be skipped.
```
hannesdiedrich, line 68:
```suggestion
- run-on-redis --> This flag targets the local testing of api client, where no user authentication is needed. A locally running redis server d3a simulation is needed here.

```

cli.py
```diff
@@ -75,22 +75,21 @@ def main(log_level):
 @click.option('-d', '--domain-name', default=DEFAULT_DOMAIN_NAME,
               type=str, help="D3A domain name")
 @click.option('-w', '--web-socket', default=DEFAULT_WEBSOCKET_DOMAIN,
-              type=str, help="D3A websocket URL")
-@click.option('-i', '--simulation-info', type=str,
-              help="Simulation File info (accept absolute and relative path)")
+              type=str, help="D3A websocket URI")
+@click.option('-i', '--simulation-config-path', type=str,
+              help="Simulation Config info (accept absolute and relative path)")
```
hannesdiedrich, line 80:
```suggestion
              help="path to simulation config file (accept absolute and relative path)")
```

d3a_api_client/redis_aggregator.py
```diff
@@ -45,12 +45,13 @@ def _connect_to_simulation(self, is_blocking=True):
...
         event_channel = f'external-aggregator/*/{self.aggregator_uuid}/events/all'
         channel_dict = {event_channel: self._events_callback_dict,
+                        "aggregator_response": self._aggregator_response_callback,
                         f"external-aggregator/*/{self.aggregator_uuid}/response/batch_commands":
                             self._batch_response,
                         }
 
         self.pubsub.psubscribe(**channel_dict)
-        self.pubsub.subscribe(**{"aggregator_response": self._aggregator_response_callback})
+        # self.pubsub.subscribe(**{"aggregator_response": self._aggregator_response_callback})
```
hannesdiedrich, line 54: please remove.

d3a_api_client/utils.py
```diff
@@ -281,17 +279,16 @@ def log_bid_offer_confirmation(message):
...
-            raise SimulationInfoException(f"Simulation_file doesn't have all the keys")
-        simulation_id = simulation_info['uuid']
-        domain_name = simulation_info['domain_name']
-        websockets_domain_name = simulation_info['web_socket_domain_name']
+def get_simulation_config(simulation_id=simulation_id_from_env, domain_name=domain_name_from_env,
+                          websockets_domain_name=websocket_domain_name_from_env):
+    if os.environ.get('SIMULATION_CONFIG_FILE_PATH', "") != "":
+        with open(os.environ['SIMULATION_CONFIG_FILE_PATH']) as json_file:
+            simulation_config = json.load(json_file)
+        from d3a_interface.simulation_config import validate_simulation_config
```
hannesdiedrich, line 287: please remove lazy import.

hannesdiedrich (commented):
Left some proposals for changing the documentation and some other comments.
There is also a `dump.rdb` which made it into the commit. Maybe add it ti .gitignore?

d3a_api_client/utils.py
```diff
@@ -259,6 +262,9 @@ def log_market_progression(message):
                                                          CUSTOMER_WEBSOCKET_DOMAIN_NAME)
 
 
+simulation_id_from_env = os.environ.get("API_CLIENT_SIMULATION_ID", API_CLIENT_SIMULATION_ID)
```
fievelk, line 265: Since this is a module constant, it would be great if you could use uppercase letters and put it on the top of the module, after the imports, as recommended by PEP8.

d3a_api_client/utils.py
```diff
@@ -271,3 +277,20 @@ def log_bid_offer_confirmation(message):
                          f"{round(energy, 2)} kWh at {price} cts/kWh")
     except Exception as e:
         logging.error(f"Logging bid/offer info failed.{e}")
+
+
+def get_simulation_config(simulation_id=simulation_id_from_env, domain_name=domain_name_from_env,
```
fievelk, line 282: I think that this method could benefit from some refactoring. You are accepting a `simulation_id` parameter, but in some cases it won't be used (when `SIMULATION_CONFIG_FILE_PATH` is available). This might lead to unexpected behaviour for the user.
Also, the `simulation_id_from_env` constant is `""` by default (as defined in `constants.API_CLIENT_SIMULATION_ID`) and I think this would cause exceptions later on.

d3a_api_client/utils.py
```diff
@@ -271,3 +277,20 @@ def log_bid_offer_confirmation(message):
...
+                          websockets_domain_name=websocket_domain_name_from_env):
+    if os.environ.get('SIMULATION_CONFIG_FILE_PATH', "") != "":
```
fievelk, line 284: I have the feeling that this check can be problematic for some reasons:
- `.get` will return `None` by default, and the truth value of `""` is `False` by default;
- if `SIMULATION_CONFIG_FILE_PATH` has not been defined, it will be `None` (and not `""`) so the current checks won't notice this issue.

I'd suggest replacing it with:
```python
if os.environ.get('SIMULATION_CONFIG_FILE_PATH'):
```

README.md
faizan2590: Any suggestions?, else will remove it then

hannesdiedrich: Yeah, remove it I would say.

d3a_api_client/utils.py
faizan2590: That was actually the motivation of this method to supersede the read values from `SIMULATION_CONFIG_FILE_PATH ` else take from the environment variable.

fievelk: I understand that the general behaviour is what you describe. But in general, when you pass an argument to a method, you expect that argument to be used (and not some default). Why not doing the opposite instead? If the user doesn't specify any argument, then you check the content of the `SIMULATION_CONFIG_FILE_PATH`?

.gitignore
```diff
@@ -131,4 +131,5 @@ dmypy.json
 .idea/
 
 .docker/
-.dump.rdb
\ No newline at end of file
+.dump.rdb
+dump.rdb
```
fievelk, line 135: Just as a small reference for the future: https://github.com/gridsingularity/d3a-api-client/pull/120#discussion_r598592011

fievelk: (If you use pyCharm this can be done automatically: `Editor > General > Ensure  every saved file ends with a line break`)

cli.py
fievelk: Great addition to the docs! What about slightly changing the `help` value here as well? Something like:
```suggestion
              help="Path of the folder containing client scripts (accepts absolute and relative path)")
```

cli.py
```diff
@@ -67,21 +75,62 @@ def main(log_level):
...
 
     except D3AException as ex:
         raise click.BadOptionUsage(ex.args[0])
     except ModuleNotFoundError as ex:
         log.error("Could not find the specified module")
+
+
+def set_simulation_file_env(base_setup_path, simulation_config_path, run_on_redis):
+    if run_on_redis is True:
+        os.environ["SIMULATION_CONFIG_FILE_PATH"] = ""
```
fievelk, line 117:
```suggestion
        os.environ.pop('SIMULATION_CONFIG_FILE_PATH', None)
```

cli.py
```diff
@@ -67,21 +75,62 @@ def main(log_level):
...
+        return
+    if simulation_config_path is None:
```
fievelk, line 119: Are we testing for `None` values or, more generally, for false truthiness values? This version would cover both:
```suggestion
    if not simulation_config_path:
```

cli.py
```diff
@@ -67,21 +75,62 @@ def main(log_level):
...
+        raise ApiSimulationConfigException(f"simulation_config_path must be provided")
+    elif os.path.isabs(simulation_config_path):
+        os.environ["SIMULATION_CONFIG_FILE_PATH"] = simulation_config_path
+    elif base_setup_path is None:
+        os.environ["SIMULATION_CONFIG_FILE_PATH"] = \
+            os.path.join(api_client_path, 'setups', simulation_config_path)
+    elif base_setup_path is not None:
```
fievelk, line 126: Same as above (for broader false values like `""`)

cli.py
```diff
@@ -67,21 +75,62 @@ def main(log_level):
...
     except ModuleNotFoundError as ex:
         log.error("Could not find the specified module")
+
+
+def set_simulation_file_env(base_setup_path, simulation_config_path, run_on_redis):
+    if run_on_redis is True:
+        os.environ["SIMULATION_CONFIG_FILE_PATH"] = ""
+        return
+    if simulation_config_path is None:
+        raise ApiSimulationConfigException(f"simulation_config_path must be provided")
```
fievelk, line 120: The `f''` prefix is not necessary here

README.md
```diff
@@ -64,7 +63,7 @@ Following is the more in depth explanation of individual flags supported,
 - domain-name --> D3A domain name
 - web-socket --> D3A websocket URI
 - simulation-config-path --> Path of JSON format file that contains agent's point of interest's collaboration/CN's info (It could downloaded from the registry page of UI) 
-- run-on-redis --> This flag targets the local testing of api client, where no user authentication is needed and needs to be skipped.
+- run-on-redis --> This flag targets the local testing of api client, where no user authentication. A locally running redis server d3a simulation is needed here.
```
hannesdiedrich, line 66:
```suggestion
- run-on-redis --> This flag targets the local testing of api client, where no user authentication. A locally running redis server and d3a simulation is needed here.
```

hannesdiedrich (approved):
One last tiny comment, LGTM

README.md
```diff
@@ -50,35 +50,36 @@ pip install git+https://github.com/gridsingularity/d3a-api-client.git
...
-```
-d3a-api-client run --setup {module_name}
-```
+### Interacting via CLI
+In order to get help, please run:
+```
+d3a-api-client run --help
+```
+Following is the more in depth explanation of individual flags supported,
+- base-setup-path --> Sets the base path where you client script resides else `d3a_api_client/setups` is taken as default for your client-scripts. User could provided both an absolute and relative file path.
```
fievelk, line 59:
```suggestion
- base-setup-path --> Sets the base path where your client script resides, otherwise `d3a_api_client/setups` is taken as default for your client scripts. Users can provide either an absolute or a relative file path.
```

README.md
```diff
@@ -50,35 +50,36 @@ pip install git+https://github.com/gridsingularity/d3a-api-client.git
...
+- setup --> Name of your api-client module/script.
+- username --> Username of agent authorized to communicate with respective collaboration/CN.
+- password --> Password of respective agent
+- domain-name --> D3A domain name
+- web-socket --> D3A websocket URI
+- simulation-config-path --> Path of JSON format file that contains agent's point of interest's collaboration/CN's info (It could downloaded from the registry page of UI)
```
fievelk, line 65: I'm not sure I understood exactly what you mean here by "agent's point of interest". This file contains the information regarding the collaboration/CN, correct?

README.md
fievelk:
```suggestion
- run-on-redis --> This flag targets the local testing of the API client, where no user authentication is required. A locally running redis server and d3a simulation are needed here.
```

README.md
```diff
@@ -50,35 +50,36 @@ pip install git+https://github.com/gridsingularity/d3a-api-client.git
...
+- run-on-redis --> This flag targets the local testing of api client, where no user authentication. A locally running redis server d3a simulation is needed here.
 
-### Authentication for REST API
-Authentication is done implicitly. The d3a-client is reading the following two variable from the environment
-and requests a auth token from the D3A API which will be reused internally for all communication.
-```
-API_CLIENT_USERNAME
-API_CLIENT_PASSWORD
+#### Examples
+- For local testing of api-client
```
fievelk, line 69:
```suggestion
- For local testing of the API client:
```

README.md
```diff
@@ -50,35 +50,36 @@ pip install git+https://github.com/gridsingularity/d3a-api-client.git
...
-  -u, --username TEXT     D3A username
-  -p, --password TEXT     D3A password
+d3a-api-client --log-level ERROR run --setup test_redis_aggregator --run-on-redis
 ```
+- For testing your api client script on remote server hosting d3a's collaboration/CNs.
+    - If your client script resides on `d3a_api_client/setups`
+    ```
+    d3a-api-client run -u <username> -p <password> --setup test_create_aggregator --simulation-config-path <your-downloaded-simulation-config-file-path>
+    ```
+    - If your client script resides on any specific directory, then it needs to set via --base-setup-path
```
fievelk, line 78:
```suggestion
    - If your client script resides on a different directory, then its path needs to be set via `--base-setup-path`
```
(In general, variable and parameter names should be put between single ticks in markdown)

cli.py
```diff
@@ -67,21 +75,62 @@ def main(log_level):
...
+    elif os.path.isabs(simulation_config_path):
```
fievelk, line 121: In the description of the `--simulation-config-path` you wrote that the user can provide both absolute and relative paths. However, it seems to me that you're only allowing absolute paths here.

cli.py
fievelk: This applies to other parts of the changes as well.

cli.py
faizan2590: Have a look at the logic following it where the case of relative path is covered.

cli.py
```diff
@@ -67,21 +75,62 @@ def main(log_level):
...
+        os.environ["SIMULATION_CONFIG_FILE_PATH"] = simulation_config_path
+    elif base_setup_path is None:
+        os.environ["SIMULATION_CONFIG_FILE_PATH"] = \
+            os.path.join(api_client_path, 'setups', simulation_config_path)
+    elif base_setup_path is not None:
+        if os.path.isabs(base_setup_path):
```
fievelk, line 127: As above: in the description of the `--base-setup-path` argument it's written that both absolute and relative paths are allowed. However, here only absolute paths are accepted if I got it right.

cli.py
fievelk: Ahhh now I get it, I think I'm mistaken on this

cli.py
```diff
@@ -67,21 +75,62 @@ def main(log_level):
...
+            os.environ["SIMULATION_CONFIG_FILE_PATH"] = \
+                os.path.join(base_setup_path, simulation_config_path)
+        else:
+            os.environ["SIMULATION_CONFIG_FILE_PATH"] = \
+                os.path.join(os.getcwd(), base_setup_path, simulation_config_path)
+
+    else:
+        os.environ["SIMULATION_CONFIG_FILE_PATH"] = \
+            os.path.join(os.getcwd(), simulation_config_path)
```
fievelk, line 136: I personally prefer methods that return values rather than changing dictionaries in place (I think we talked about it a few days ago on slack). It would be good to create a method that just creates the correct path, and subsequently assign the returned value to the `os.environ['SIMULATION_CONFIG_FILE_PATH']` variable. 
For example:
```python
if run_on_redis:
    os.environ['SIMULATION_CONFIG_FILE_PATH'] = None
else:
    os.environ['SIMULATION_CONFIG_FILE_PATH'] = create_simulation_config_path(
        base_setup_path, simulation_config_path)
```
and the method might look something like:
```python
def create_simulation_config_path(base_setup_path, simulation_config_path):
    # For example
    if os.path.isabs(simulation_config_path):
        return simulation_config_path
    # and so on
```

Hassan754:
LGTM after adapting to Pierpaolo's comment

d3a_api_client/aggregator.py
```diff
@@ -50,11 +50,13 @@ def __init__(self, simulation_id, domain_name, aggregator_name,
 
     def _connect_to_simulation(self):
         user_aggrs = self.list_aggregators()
+        print(f"user_aggrs: {user_aggrs}")
```
fievelk, line 53: This was probably a debug leftover

d3a_api_client/resources/api-client-summary-84d221fa-46e1-49e0-823b-26cfb3425a5a.json
```diff
@@ -0,0 +1,90 @@
+{
```
fievelk, line 1: I would rename this folder to either `factories` or `fixtures` (and possibly move it inside the `unit_tests/` folder).
I highly recommend using [factory_boy](https://github.com/FactoryBoy/factory_boy) for this. You can check what has already been done in d3a-web: https://github.com/gridsingularity/d3a-web/blob/master/simulation/tests/factories.py

d3a_api_client/rest_device.py
```diff
@@ -1,12 +1,13 @@
 import logging
+
```
fievelk, line 2: I'm not sure that this blank line is needed: `concurrent.futures` is part of the standard library as well I think.

d3a_api_client/utils.py
```diff
@@ -271,3 +277,24 @@ def log_bid_offer_confirmation(message):
                          f"{round(energy, 2)} kWh at {price} cts/kWh")
     except Exception as e:
         logging.error(f"Logging bid/offer info failed.{e}")
+
+
+def get_simulation_config(simulation_id=None, domain_name=None, websockets_domain_name=None):
+    if os.environ.get('SIMULATION_CONFIG_FILE_PATH') not in ["", None]:
```
fievelk, line 283:
```suggestion
    if os.environ.get('SIMULATION_CONFIG_FILE_PATH'):
```

d3a_api_client/rest_device.py
```diff
@@ -15,20 +16,20 @@
...
-    def __init__(self, simulation_id, device_id,
-                 domain_name=domain_name_from_env,
-                 websockets_domain_name=websocket_domain_name_from_env,
+    def __init__(self, device_id=None, simulation_id=None,
+                 domain_name=None,
+                 websockets_domain_name=None,
                  autoregister=False, start_websocket=True,
                  sim_api_domain_name=None):
-        self.simulation_id = simulation_id
+        self.simulation_id, self.domain_name, self.websockets_domain_name = \
```
fievelk, line 24: What's the use-case for this? If I had to instantiate a new `RestDeviceClient` I would expect to fetch my variables and then use them in the constructor:
```python
simulation_id, domain_name, websockets_domain_name = get_variables_from_file(filename)
RestDeviceClient(
    simulation_id=simulation_id,
    domain_name=domain_name,
    websockets_domain_name=websockets_domain_name)
```

d3a_api_client/setups/test_create_aggregator.py
```diff
@@ -1,11 +1,16 @@
 from pendulum import today
+import os
```
fievelk, line 2: These imports should all come above the `pendulum` import (since they belong to the standard library).

fievelk: (Having them in alphabetical order is also a good practice, in case you want to)

d3a_api_client/setups/test_create_aggregator.py
```diff
@@ -28,16 +33,18 @@ def on_market_cycle(self, market_info):
             return
 
         for device_event in market_info["content"]:
-            if "available_energy_kWh" in device_event["device_info"] and \
-                    device_event["device_info"]["available_energy_kWh"] > 0.0:
-                self.add_to_batch_commands.offer_energy(device_event["area_uuid"], price=1,
-                                                        energy=device_event["device_info"]["available_energy_kWh"] / 2)
+            if key_in_dict_and_not_none_and_greater_than_zero(
+                    device_event, "available_energy_kWh"):
```
fievelk, line 37: Single quotes vs double quotes :P

d3a_api_client/utils.py
```diff
@@ -11,11 +11,26 @@
...
     CUSTOMER_WEBSOCKET_DOMAIN_NAME
 from d3a_interface.utils import get_area_name_uuid_mapping, key_in_dict_and_not_none, \
     RepeatingTimer
 
+from d3a_api_client.constants import DEFAULT_DOMAIN_NAME, DEFAULT_WEBSOCKET_DOMAIN, \
+    API_CLIENT_SIMULATION_ID
+
+DOMAIN_NAME_FROM_ENV = os.environ.get("API_CLIENT_DOMAIN_NAME", DEFAULT_DOMAIN_NAME)
+
+
```
fievelk, line 25: Why are you opting to have blank lines between each constant?

unit_tests/test_simulation_config.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class TestSimulationInfo(unittest.TestCase):
+
+    def setUp(self) -> None:
+        os.environ['SIMULATION_CONFIG_FILE_PATH'] = ""
+        self.api_client_path = os.path.dirname(inspect.getsourcefile(d3a_api_client))
+
+    def tearDown(self) -> None:
+        os.environ['SIMULATION_CONFIG_FILE_PATH'] = ""
```
fievelk, line 20:
```suggestion
        os.environ.pop('SIMULATION_CONFIG_FILE_PATH', None)
```

unit_tests/test_simulation_config.py
```diff
@@ -0,0 +1,62 @@
...
+            self.api_client_path,
+            "resources/api-client-summary-84d221fa-46e1-49e0-823b-26cfb3425a5a.json"
+        )
+        simulation_id, domain_name, websockets_domain_name = get_simulation_config()
+        assert simulation_id == "84d221fa-46e1-49e0-823b-26cfb3425a5a"
+        assert domain_name == "localhost"
+        assert websockets_domain_name == "ws://localhost/external-ws"
+
+    def test_assertion_is_raised_if_incorrect_info_is_provided(self):
+        with self.assertRaises(ValidationError):
```
fievelk, line 39: Great, I like this test

d3a_api_client/resources/api-client-summary-84d221fa-46e1-49e0-823b-26cfb3425a5a.json
fievelk: Edit: since you're reading these files, using factory boy wouldn't help here.

fievelk (commented):
Left some comments! 👍

README.md
faizan2590: `agent's point of interest` --> each users requesting it would only getting its own approved areas and selected aggregator info. May be you can propose better wording here

fievelk: Ah, I see. Not easy...what about:
```suggestion
- `--simulation-config-path`: path to the JSON file that contains the user's collaboration/CN info (it can be downloaded from the registry page of the D3A website)
```
I think there's no need to be more specific, since you're already saying that this file will only contain information related to the user.

d3a_api_client/rest_device.py
faizan2590: It is for user's convenience to not explicitly need to pass it but to have an option to auto read for the respective simulation config file.

d3a_api_client/utils.py
faizan2590: just a habit, is there any PEP8 for that too ?

faizan2590: Actually there is: https://www.python.org/dev/peps/pep-0008/#blank-lines

d3a_api_client/rest_device.py
fievelk: Hmm ok. I personally think it makes the class more complicated. But if you prefer it, that's good for me :)

d3a_api_client/utils.py
fievelk: Yes, so as far as I read there shouldn't be empty lines between constants.

faizan2590: Yes, adapted accordingly!

d3a_api_client/resources/api-client-summary-84d221fa-46e1-49e0-823b-26cfb3425a5a.json
faizan2590: > and possibly move it inside the unit_tests/ folder)

Its deliberately kept here as `unit_tests` is outside `api_client_path`

fievelk: Can you elaborate on this? What's the issue that would be generated if we moved it into `unit_tests/`?

cli.py
```diff
@@ -27,12 +30,15 @@
...
 from d3a_interface.utils import iterate_over_all_modules
+from d3a_interface.api_simulation_config import ApiSimulationConfigException
+import d3a_api_client
 
 import d3a_api_client.setups as setups
 from d3a_api_client.constants import SETUP_FILE_PATH, DEFAULT_DOMAIN_NAME, DEFAULT_WEBSOCKET_DOMAIN
 
 
 log = getLogger(__name__)
+api_client_path = os.path.dirname(inspect.getsourcefile(d3a_api_client))
```
fievelk, line 41: It's the first time I see this strategy used to retrieve a path. I would have usually done something like:
```python
import os
current_dir = os.path.dirname(os.path.abspath(__file__))  # Directory of the cli.py script
api_client_path = os.path.join(current_dir, 'd3a_api_client')
```
I have no real reason to ask for a change as of now, just wanted to mention it in case you find it useful for the future.

cli.py
fievelk: This (and the other one above) wasn't addressed. Do we have specific reasons to keep it as it is?

cli.py
```diff
@@ -112,25 +114,20 @@ def load_client_script(base_setup_path, setup_module_name):
...
 
 
-def set_simulation_file_env(base_setup_path, simulation_config_path, run_on_redis):
+def create_simulation_config_path(base_setup_path, simulation_config_path, run_on_redis):
     if run_on_redis is True:
-        os.environ["SIMULATION_CONFIG_FILE_PATH"] = ""
-        return
-    if simulation_config_path is None:
-        raise ApiSimulationConfigException(f"simulation_config_path must be provided")
+        return ""
```
fievelk: The reason why I suggested removing `run_on_redis` from this function is that the path to the simulation config can/should be created regardless of Redis. If you are using Redis, then you don't need the simulation config path at all. If you are not using Redis, then you need to compute the path. But the check for Redis doesn't belong to the logic for the costruction of the path.
If you think about it in terms of unit testing and atomicity of tasks, it makes sense to split the two things.

d3a_api_client/setups/test_create_aggregator.py
fievelk: Single quotes were also not addressed in the changes. Let's keep it in mind for the next PRs :)

unit_tests/test_simulation_config.py
```diff
@@ -44,19 +44,23 @@ def test_assertion_is_raised_if_incorrect_info_is_provided(self):
             get_simulation_config()
 
     def test_default_simulation_info_is_under_setup_module_if_base_path_isnt_provided(self):
-        set_simulation_file_env(base_setup_path=None, simulation_config_path='test.json',
-                                run_on_redis=False)
+        os.environ["SIMULATION_CONFIG_FILE_PATH"] = \
```
fievelk, line 47: I think this way to break lines is somehow encouraged by some PyCharm default settings. However, it is actually discouraged by PEP8:
> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

fievelk (dismissed):
Left some comments. They don't have to be necessarily addressed in this PR, but it would be cool to keep some of them in mind for the next PRs to make the codestyle coherent throughout the repos.

unit_tests/test_simulation_config.py
fievelk: In this case, this could become:
```python
os.environ["SIMULATION_CONFIG_FILE_PATH"] = create_simulation_config_path(
            base_setup_path=None, simulation_config_path='test.json', run_on_redis=False)
```

faizan2590:
> Left some comments. They don't have to be necessarily addressed in this PR, but it would be cool to keep some of them in mind for the next PRs to make the codestyle coherent throughout the repos.

Thanks for comments related to code style, I would be very much in favour of starting an internal wiki for all of us to uniformly comply the same coding style.

cli.py
faizan2590: Thats something I took as a [reference](https://github.com/gridsingularity/d3a/blob/master/src/d3a/d3a_core/util.py#L42) from `d3a` repo.

d3a_api_client/resources/api-client-summary-84d221fa-46e1-49e0-823b-26cfb3425a5a.json
faizan2590: The reason I used `api_client_path ` and kept resource files under `d3a_api_client` to have a uniform reference for test being executed locally as well as via `tox` as they both were taking different `cwd`

unit_tests/test_simulation_config.py
faizan2590: https://github.com/gridsingularity/d3a-api-client/pull/94/commits/502901535a99d23ce71dc26480ac8f9c7da5a49e

cli.py
faizan2590: https://github.com/gridsingularity/d3a-api-client/pull/94/commits/502901535a99d23ce71dc26480ac8f9c7da5a49e

d3a_api_client/resources/api-client-summary-84d221fa-46e1-49e0-823b-26cfb3425a5a.json
fievelk: But why not just specifying another path in the tests, without joining it to the `d3a_api_client` module location? These resources are not part of the `d3a_api_client`, they're specifically used in tests (as fixtures) and we should be able to de-couple the two things without breaking any functionality in the client.

fievelk (approved):

PR #97, opened 2021-02-05
Feature/d3 asim 3210

hannesdiedrich:


unit_tests/test_grid_fee_caclulation.py
```diff
@@ -0,0 +1,151 @@
...
+        if fee_type == 'next_market_fee':
+            leaf_name_expected_fee = {21.3: ['PV 2.1', 'Load 1.2'],
+                                      21.1: ['PV 2.1', 'Load 3'],
+                                      15.0: ['Market Maker', 'Load 3']}
+        else:
+            leaf_name_expected_fee = {16.3: ['PV 2.1', 'Load 1.2'],
+                                      17.1: ['PV 2.1', 'Load 3'],
+                                      13.0: ['Market Maker', 'Load 3']}
+        for expected_fee, leaf_names in leaf_name_expected_fee.items():
+            assert isclose(expected_fee, self.grid_fee_calc.calculate_grid_fee(
```
Hassan754, line 102: I'm aware of how isClose works , but why would the values slightly differ at first place ?

d3a_api_client/aggregator.py
```diff
@@ -33,14 +34,15 @@ def received_message(self, message):
             self.command_response_buffer.append(message)
 
 
-class Aggregator(RestDeviceClient):
+class Aggregator(RestDeviceClient, GridFeeCalculation):
```
Hassan754, line 37: I would prefer if you use composition in here , add an instance attribute `_grid_fee_calculation = GridFeeCalculation()` and then refactor accordingly .
This would save us troubles later dealing with multiple inheritance issues .
when in doubt, I always ask : 
A is  B? -> inheritance
A has B -> Composition
And we're clearly in the second case

d3a_api_client/grid_fee_calculation.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def __init__(self):
+        self.latest_grid_stats_tree = {}
+        self.paths_to_root_mapping = {}
+        self.market_name_grid_fee_mapping = {"last_market_fee": {},
+                                             "next_market_fee": {}}
+
+    def _handle_grid_stats(self, message):
+        for device_event in message["content"]:
+            if device_event["event"] == "grid stats":
```
Hassan754, line 17:
```suggestion
            if device_event["event"] == "grid_stats":
```
Minor , we can skip it .

d3a_api_client/grid_fee_calculation.py
```diff
@@ -0,0 +1,102 @@
...
+        return out_list
+
+    @staticmethod
+    def _find_lowest_intersection_market(in_list, intersection):
+        last_li = in_list[0]
+        for li in in_list:
+            if li not in intersection:
+                return last_li
+            last_li = li
+        return last_li
```
Hassan754, line 55: Cool !

d3a_api_client/grid_fee_calculation.py
```diff
@@ -0,0 +1,102 @@
...
+        """
+        if not self.latest_grid_stats_tree:
+            logging.info("Grid fees can not be calculated because there were no grid_stats sent yet.")
+            return None
+
+        if target_market_or_device_name is None:
+            if start_market_or_device_name not in self.market_name_grid_fee_mapping[fee_type]:
+                return self.market_name_grid_fee_mapping[fee_type][
+                    self.paths_to_root_mapping[start_market_or_device_name][-1]]
+            else:
```
Hassan754, line 70: As I understood the first case is for the leaves and the second is for markets .
I think adding a comment in here and there in the class will help us later on when we get back to code

Hassan754 (commented):
Interesting PR, thanks for tackling it.
Left some comments , LGTM in general

d3a_api_client/grid_fee_calculation.py
```diff
@@ -0,0 +1,102 @@
...
+
+            all_markets_along_path = set([lowest_intersection_market] +
+                                         path_start_market_stripped + path_target_market_stripped +
+                                         [start_market_or_device_name] + [target_market_or_device_name])
+
+        total_grid_fees = 0
+        for ma in all_markets_along_path:
+            if key_in_dict_and_not_none(self.market_name_grid_fee_mapping[fee_type], ma):
+                total_grid_fees += self.market_name_grid_fee_mapping[fee_type][ma]
+
```
Hassan754, line 101: I'm not sure if I missed it , but where is the percent fee equation ?

d3a_api_client/grid_fee_calculation.py
hannesdiedrich: No, you are right! Its better to have an underscore here.

unit_tests/test_grid_fee_caclulation.py
hannesdiedrich: This happens some times when floating precision propagate in calculations. Here the calculated grid_fee was something like `12.999993` and the expected grid_fee was `13`

d3a_api_client/aggregator.py
hannesdiedrich: Ok, thanks!

d3a_api_client/grid_fee_calculation.py
spyrostz: In addition to that, is there a need for an extra event, other than the market_cycle one ? If the problem is when to decide to calculate the grid fees or not, it can be done fairly quickly by checking if the grid_stats_tree contains children or not, removing the need for another event type.

d3a_api_client/grid_fee_calculation.py
```diff
@@ -0,0 +1,102 @@
...
+            if device_event["event"] == "grid stats":
+                self.latest_grid_stats_tree = device_event["grid_stats_tree"]
+                message["content"].remove(device_event)
+        self._populate_grid_fee_mappings()
+
+    def _populate_grid_fee_mappings(self):
+        for fee_type in ["last_market_fee", "next_market_fee"]:
+            self._get_grid_fee_area_mapping_from_dict(self.latest_grid_stats_tree, fee_type)
+
+        self._get_all_paths_in_grid_stats_dict(self.latest_grid_stats_tree, [])
```
spyrostz, line 26: I think this is not optimal, because these methods walk 3 times over the grid tree, while just once would be enough. Why not having just one walk over the tree nodes, and populate the class members while being on every node of the tree? For instance, one method could populate the market_name_grid_fee_mapping variable for both last_market_fee, next_market_fee, and also the self.paths_to_root_mapping in one go.

d3a_api_client/grid_fee_calculation.py
hannesdiedrich: If you have a look into the d3a PR ( https://github.com/gridsingularity/d3a/pull/1045 )
I didn't add a new event type, I only marked the grid_stats by setting the "event" member in the market_event.
The one grid_stats information is prepended to all market_events of one aggregator.
So if the aggregator controls 2 devices, it will get a list of two market_cycle_events plus one grid_stats information.
Now I only explicitly market it as such. Let me know if you need more info.

d3a_api_client/grid_fee_calculation.py
```diff
@@ -0,0 +1,102 @@
...
+            self.paths_to_root_mapping[child_name] = parent_path
+            if "children" in child_stats:
+                self._get_all_paths_in_grid_stats_dict(child_stats["children"], sub_path)
+
+    @staticmethod
+    def _strip_away_intersection_from_list(in_list, intersection):
+        out_list = copy(in_list)
+        for inter in intersection:
+            out_list.remove(inter)
+        return out_list
```
spyrostz, line 46: I am not sure which of the 2 is better WRT performance, but if you want to convert it to an one-liner, there is a set method calculating this:
```suggestion
        return list(set(in_list).summetric_difference(set(intersection)))
```

hannesdiedrich: Will do, looks better IMO.

d3a_api_client/grid_fee_calculation.py
hannesdiedrich: Will add explaining comments

Hassan754 (commented):
I'm still thinking that we're implementing heavy calculations in each tick/cycle when some users won't even use this feature
So I'm thinking we can pass an argument at instanciation ,if set to True we so create the extra attribute and calculations

d3a_api_client/aggregator.py
```diff
@@ -34,11 +35,12 @@ class Aggregator(RestDeviceClient):
 
     def __init__(self, simulation_id, domain_name, aggregator_name,
                  websockets_domain_name, accept_all_devices=True):
-        super().__init__(
-            simulation_id=simulation_id, device_id="", domain_name=domain_name,
-            websockets_domain_name=websockets_domain_name, autoregister=False,
-            start_websocket=False)
+        RestDeviceClient.__init__(self, simulation_id=simulation_id, device_id="",
```
Hassan754, line 38: I think we can revert this back to super()

d3a_api_client/redis_aggregator.py
```diff
@@ -34,6 +36,7 @@ def __init__(self, aggregator_name, accept_all_devices=True,
         self._subscribe_to_response_channels()
         self.executor = ThreadPoolExecutor(max_workers=MAX_WORKER_THREADS)
         self.lock = Lock()
+        self.latest_grid_stats_tree = {}
```
Hassan754, line 39: This is not needed anymore ,right?

hannesdiedrich:
> I'm still thinking that we're implementing heavy calculations in each tick/cycle when some users won't even use this feature
> So I'm thinking we can pass an argument at instanciation ,if set to True we so create the extra attribute and calculations

I you mean `handle_grid_stats`, it is only triggered every market cycle and I wouldn't specifically call it `heavy calculations` as is just looping over a dict. Consequently, in the CN case, it is only done every 15 mins, so no problem IMO. This is only on the client side too, so for us on the d3a-web or d3a side there is no change. 
Also, i dont think there will be a lot of cases where the Aggregator does not want to know the grid_fees, so your proposed flag will probably set to True nearly all of the times.

Hassan754:
> I you mean `handle_grid_stats`, it is only triggered every market cycle and I wouldn't specifically call it `heavy calculations` as is just looping over a dict. Consequently, in the CN case, it is only done every 15 mins, so no problem IMO. This is only on the client side too, so for us on the d3a-web or d3a side there is no change.
> Also, i dont think there will be a lot of cases where the Aggregator does not want to know the grid_fees, so your proposed flag will probably set to True nearly all of the times.

I see, alright then

Hassan754 (dismissed):

spyrostz (approved):
LGTM

Hassan754 (approved):

spyrostz (approved):
LGTM

PR #98, opened 2021-02-08
D3ASIM-3201: Add get_aggregators_list function + Updated README

Hassan754:


README.md
```diff
@@ -152,6 +152,17 @@ commands in order to react to an event simultaneously for each owned device.
     ``` 
     aggregator = AutoAggregator(<aggregator_name>)
     ```
+### How to list your aggregators
+To list your aggregators, its simulation id and the registered devices, you should :
```
faizan2590, line 156:
```suggestion
To list your aggregators, its configuration id and the registered devices, you should :
```

d3a_api_client/utils.py
```diff
@@ -166,6 +166,26 @@ def get_area_uuid_and_name_mapping_from_simulation_id(collab_id, domain_name):
         return area_name_uuid_map
 
 
+def get_aggregators_list(domain_name):
+    """
+    Returns a list of aggregators for the logged in user
+    """
+    jwt_key = retrieve_jwt_key_from_server(domain_name)
+    from sgqlc.endpoint.http import HTTPEndpoint
```
faizan2590, line 174: Any specific reason for lazy import?

faizan2590 (approved):
LGTM though left some comments.

d3a_api_client/utils.py
Hassan754: @faizan2590 There was no need, it was used several places in the file so I did some refactoring now

d3a_api_client/commands.py
```diff
@@ -14,13 +14,13 @@ def offer_energy(self, area_uuid, energy, price):
         return self._add_to_buffer(area_uuid, Commands.OFFER, {"energy": energy, "price": price})
 
     def offer_energy_rate(self, area_uuid, energy, rate):
-        return self._add_to_buffer(area_uuid, Commands.OFFER, {"energy": energy, "rate": rate})
+        return self._add_to_buffer(area_uuid, Commands.OFFER, {"energy": energy, "price": rate})
```
Hassan754, line 17: Side bug fix

spyrostz (approved):
LGTM, one important comment.

d3a_api_client/commands.py
spyrostz: You need to multiply rate * energy to calculate the correct price value.

Hassan754: @spyrostz It depends on whether the user will want to pass rate here or the price , what do you think would make more sense ?

fievelk: Given the name of the method, I actually think the user would likely pass the rate

Hassan754: @fievelk I agree, @spyrostz what do you think?

spyrostz: The method is named `offer_energy_rate` , therefore the user needs to pass the energy rate in cents / kWh here. The api parameter is named price, therefore the API requires the price in cents. Therefore the client has to act as a translator and convert the rate -> price. At least that's my understanding, also have a look at the offer_energy_rate in the rest / redis devices, you will see the conversion.

d3a_api_client/utils.py
```diff
@@ -166,6 +167,19 @@ def get_area_uuid_and_name_mapping_from_simulation_id(collab_id, domain_name):
         return area_name_uuid_map
 
 
+def get_aggregators_list(domain_name):
```
hannesdiedrich, line 170: IMO the domain_name should be optional here and the global environmental setting should be used as default.

hannesdiedrich (commented):
One comment, LGTM once resolved

hannesdiedrich (dismissed):
LGTM

d3a_api_client/redis_aggregator.py
```diff
@@ -199,7 +199,7 @@ def _on_tick(self, message):
 
     def _on_trade(self, message):
         logging.info(f"<-- {message.get('buyer')} BOUGHT {round(message.get('energy'), 4)} kWh "
-                     f"at {round(message.get('price'), 2)}/kWh -->")
+                     f"at {round(message.get('price'), 2)} cents/kWh -->")
```
spyrostz, line 202: Price is in cents, not cents / kWh. If you want the energy rate you can either divide price / energy, or use the message["offer"]["energy_rate"] IIRC.

d3a_api_client/redis_client_base.py
```diff
@@ -268,7 +268,7 @@ def _on_tick(self, msg):
     def _on_trade(self, msg):
         message = json.loads(msg["data"])
         logging.info(f"<-- {message.get('buyer')} BOUGHT {round(message.get('energy'), 4)} kWh "
-                     f"at {round(message.get('price'), 2)}/kWh -->")
+                     f"at {round(message.get('price'), 2)} cents/kWh -->")
```
spyrostz, line 271: Same as above

spyrostz (approved):
2 comments, LGTM once resolved.

hannesdiedrich (approved):
LGTM

d3a_api_client/utils.py
```diff
@@ -47,6 +48,23 @@ def _get_request(self, endpoint_suffix, data):
         return data["transaction_id"], get_request(endpoint, data, self.jwt_token)
 
 
+def post_graphql_request(domain_name, query, headers=None, url=None, authenticate=True):
```
faizan2590, line 51:
```suggestion
def execute_graphql_request(domain_name, query, headers=None, url=None, authenticate=True):
```
As using `post` gives an impression that it would only be a `mutation` and no `query`

d3a_api_client/utils.py
```diff
@@ -166,6 +167,21 @@ def get_area_uuid_and_name_mapping_from_simulation_id(collab_id, domain_name):
...
+    """
+    if not domain_name:
+        domain_name = os.environ.get("API_CLIENT_DOMAIN_NAME")
+    query = 'query { aggregatorsList { configUuid name  devicesList { deviceUuid } } }'
+
+    data = post_graphql_request(domain_name=domain_name, query=query)
+    if key_in_dict_and_not_none(data, "errors"):
+        return ast.literal_eval(data["errors"][0]["message"])
+    else:
+        return data["data"]["aggregatorsList"]
```
faizan2590, line 182:
```suggestion
    return ast.literal_eval(data["errors"][0]["message"]) if key_in_dict_and_not_none(data, "errors") else return data["data"]["aggregatorsList"]
```
Could be one liner here.

faizan2590 (approved):
Some enhancement comments, other than that LGTM

faizan2590 (approved):
LGTM

hannesdiedrich (approved):
LGTM

PR #102, opened 2021-02-12
D3ASIM-3301: Add Precaution for None energy value

Hassan754:


d3a_api_client/aggregator.py
```diff
@@ -32,6 +32,9 @@ def received_message(self, message):
         elif "command" in message:
             self.command_response_buffer.append(message)
 
+        if "event" in message or "command" in message:
+            self.client._on_event_or_response(message)
```
hannesdiedrich, line 36: Thanks a lot for that, i was just seeing that the logs are missing in the REST aggregator.

d3a_api_client/redis_aggregator.py
```diff
@@ -198,6 +198,9 @@ def _on_tick(self, message):
                              function_name="on_tick")
 
     def _on_trade(self, message):
+        if message.get('energy') is None:
+            logging.error(f"Received null energy value {message}")
+            return
```
hannesdiedrich, line 203: We also need to be prepared for other wrong values. Look at the `price` here. It could also be None.

hannesdiedrich: I suggest to create a validation function (or decorator) that checks for the price and energy value. What do you think?

Hassan754: I agree , on it

PR #103, opened 2021-02-15
D3ASIM-3302: Do not raise exception and force the websocket to discon…

spyrostz:
…nect in case there is an error in the handling of the incoming messages. Corrected logging of incoming trade messages for the aggregator.

### DIFF
--- a/d3a_api_client/aggregator.py
+++ b/d3a_api_client/aggregator.py
@@ -1,5 +1,5 @@
 import logging
-
+import traceback
 from d3a_api_client.commands import ClientCommandBuffer
 from d3a_api_client.utils import logging_decorator, blocking_get_request, \
     blocking_post_request
@@ -14,23 +14,27 @@ class AggregatorWebsocketMessageReceiver(WebsocketMessageReceiver):
         super().__init__(rest_client)
 
     def received_message(self, message):
-        if "event" in message:
-            if message["event"] == "market":
-                self.client._on_market_cycle(message)
-            elif message["event"] == "tick":
-                self.client._on_tick(message)
-            elif message["event"] == "trade":
-                self.client._on_trade(message)
-            elif message["event"] == "finish":
-                self.client._on_finish(message)
-            elif message["event"] == "selected_by_device":
-                self.client._selected_by_device(message)
-            elif message["event"] == "unselected_by_device":
-                self.client._unselected_by_device(message)
-            else:
-                logging.error(f"Received message with unknown event type: {message}")
-        elif "command" in message:
-            self.command_response_buffer.append(message)
+        try:
+            if "event" in message:
+                if message["event"] == "market":
+                    self.client._on_market_cycle(message)
+                elif message["event"] == "tick":
+                    self.client._on_tick(message)
+                elif message["event"] == "trade":
+                    self.client._on_trade(message)
+                elif message["event"] == "finish":
+                    self.client._on_finish(message)
+                elif message["event"] == "selected_by_device":
+                    self.client._selected_by_device(message)
+                elif message["event"] == "unselected_by_device":
+                    self.client._unselected_by_device(message)
+                else:
+                    logging.error(f"Received message with unknown event type: {message}")
+            elif "command" in message:
+                self.command_response_buffer.append(message)
+        except Exception as e:
+            logging.error(f"Error while processing incoming message {message}. Exception {e}.\n"
+                          f"{traceback.format_exc()}")
 
 
 class Aggregator(RestDeviceClient):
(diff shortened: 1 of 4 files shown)


d3a_api_client/rest_device.py
```diff
@@ -166,9 +166,20 @@ def _on_tick(self, message):
...
+        logging.info(f"<-- {message.get('buyer')} BOUGHT {round(message.get('energy'), 4)} kWh "
                      f"at {round(message.get('price'), 2)}/kWh -->")
+
+    def _on_trade(self, message):
+        if "context" not in message:
+            # Device message
+            self._log_trade_info(message)
+        else:
+            # Aggregator message
+            for individual_trade in message["content"]:
```
faizan2590, line 180:
```suggestion
            for "individual_trade" in message["context"]:
```
Not 100% sure though

faizan2590 (dismissed):
One recommendation, other than that LGTM

d3a_api_client/rest_device.py
```diff
@@ -166,9 +166,20 @@ def _on_tick(self, message):
         self.callback_thread.submit(execute_function_util, function=function,
                                     function_name="on_tick")
 
-    def _on_trade(self, message):
-        logging.info(f"<-- {message.get('buyer')} BOUGHT {round(message.get('energy'),4)} kWh "
+    @staticmethod
+    def _log_trade_info(message):
+        logging.info(f"<-- {message.get('buyer')} BOUGHT {round(message.get('energy'), 4)} kWh "
                      f"at {round(message.get('price'), 2)}/kWh -->")
```
hannesdiedrich, line 172: So we are saying that `message.get('energy')` never returns None and skip the None check now?

hannesdiedrich: Just for clarification, I get the error fixing below.

d3a_api_client/websocket_device.py
```diff
@@ -16,22 +16,26 @@ def __init__(self, rest_client):
...
+                    self.client._on_finish(message)
+                else:
+                    logging.error(f"Received message with unknown event type: {message}")
+            elif "command" in message:
+                self.command_response_buffer.append(message)
 
-        if "event" in message or "command" in message:
-            self.client._on_event_or_response(message)
+            if "event" in message or "command" in message:
+                self.client._on_event_or_response(message)
```
Hassan754, line 35: Can you please add these 2 lines of code to the Aggregator websocket
```d3a_api_client/aggregator.py:16```

File: d3a_api_client/rest_device.py (hunk 2 above)
Comment (hannesdiedrich) [reply d3a_api_client/rest_device.py:172]: Forget it, yes, you added a new log message that should tell us directly what is happening for better debugging later.

**Review by hannesdiedrich — APPROVED** [review]
LGTM

File: d3a_api_client/rest_device.py (hunk 1 above)
Comment (spyrostz) [reply d3a_api_client/rest_device.py:180]: Not really. The message["content"] in the aggregator message contains a list of dicts.

File: d3a_api_client/rest_device.py (hunk 2 above)
Comment (spyrostz) [reply d3a_api_client/rest_device.py:172]: It indeed never returns None, and even if this happens at some point due to a d3a bug, the Exception will be caught and logged by the websocket_device.py -> received_message() and we'll get notified of both the traceback and the incoming message for debugging.

File: d3a_api_client/websocket_device.py (hunk 3 above)
Comment (spyrostz) [reply d3a_api_client/websocket_device.py:35]: Darn I forgot that this method is overridden. Will do that, thanks for pointing it out!

File: d3a_api_client/aggregator.py [hunk 4]
```diff
@@ -14,23 +14,27 @@ def __init__(self, rest_client):
...
+                elif message["event"] == "finish":
+                    self.client._on_finish(message)
+                elif message["event"] == "selected_by_device":
+                    self.client._selected_by_device(message)
+                elif message["event"] == "unselected_by_device":
+                    self.client._unselected_by_device(message)
+                else:
+                    logging.error(f"Received message with unknown event type: {message}")
+            elif "command" in message:
+                self.command_response_buffer.append(message)
```
Hassan754, line 34:
```python
            elif "command" in message:
        if "event" in message or "command" in message:
            self.client._on_event_or_response(message)
                
```
Do you mind , as a side fix, adding the on_event_or_response callback here, I noticed that due to inheritance this one two needs to be explicitly added

hannesdiedrich: Yes, I vote for that too, thanks!

hannesdiedrich: I also promised Sarah that :-)

spyrostz: Thanks for also catching that ! I'll add it now

PR #104, opened 2021-02-15
 Added update_bid/ update_offer to aggregator batch commands

Hassan754:


d3a_api_client/commands.py
```diff
@@ -16,12 +19,18 @@ def offer_energy(self, area_uuid, energy, price):
     def offer_energy_rate(self, area_uuid, energy, rate):
         return self._add_to_buffer(area_uuid, Commands.OFFER, {"energy": energy, "rate": rate})
 
+    def update_offer(self, area_uuid, energy, price):
+        return self._add_to_buffer(area_uuid, "update_offer", {"energy": energy, "price": price})
```
hannesdiedrich, line 23: minor, but you could add update_offer and update_bid to the `Offers` class.

hannesdiedrich (approved):
One enhancement comment, once addressed, LGTM

d3a_api_client/commands.py
Hassan754: @hannesdiedrich  You mean to create an Enum for storing these string values ?

Hassan754: @hannesdiedrich If I understood correctly I should create an Enum class for these String values ?

hannesdiedrich: @Hassan754 is is already there https://github.com/gridsingularity/d3a-api-client/blob/fc5f01facac3c9f0f297ef9921b3387817f7114c/d3a_api_client/enums.py#L4 please add the "new" commands

hannesdiedrich: Sorry, I just saw that I miswrote in my first comment I meant the `Commands` class of course.

Hassan754: @hannesdiedrich This is not possible because the Commands Enum class is consumed the `RedisClientBase` , it is iterated and subscribed to all commands in it.   d3a_api_client/redis_client_base.py:48
Some of them are common channels name so this will throw error
If we're to add these values to an Enum, I can create a specific enum which inherits from the CommandsENum and extends as much as needed

d3a_api_client/commands.py
```diff
@@ -16,12 +19,18 @@ def offer_energy(self, area_uuid, energy, price):
     def offer_energy_rate(self, area_uuid, energy, rate):
         return self._add_to_buffer(area_uuid, Commands.OFFER, {"energy": energy, "rate": rate})
```
spyrostz, line 20: I recall a discussion we had some PRs ago, I think that d3a API does not support the "rate" parameter. Therefore this needs to change:
```suggestion
        return self._add_to_buffer(area_uuid, Commands.OFFER, {"energy": energy, "price": rate * energy})
```

Hassan754: @spyrostz the mentioned PR is not merged yet, it is fixed there so they should sync upon merging

spyrostz: That works for me, however make sure that they are not using this in the researcher's scripts (I am mentioning because IIRC they usually use the energy rate, and if you do not apply this fix their scripts will fail). If they do not use it then feel free to merge this PR, otherwise just port the fix in this PR (IIRC it is only 2 these lines that you need to adapt)

d3a_api_client/commands.py
hannesdiedrich: @Hassan754 `update_offer` and `update_bid` are also commands that should be available for the redis client. 
So I see no harm in that. You will need to adapt `RedisClient._response_topics` then.
I dont know how urgent this PR here, but if it is really important you could go and create an enhancement story for after the deploy that homoginzes these command enums.

spyrostz (approved):
One comment, LGTM otherwise

d3a_api_client/commands.py
Hassan754: Done

d3a_api_client/commands.py
Hassan754: @hannesdiedrich Actually it's quiet urgent, Sarah is constantly pinging me , anyway I created a separate ticket https://gridsingularity.atlassian.net/browse/D3ASIM-3319

hannesdiedrich (approved):
LGTM

Hassan754:
@spyrostz @hannesdiedrich Adapted to your reviews

hannesdiedrich (approved):
LGTM

faizan2590 (approved):
LGTM

PR #105, opened 2021-02-15
D3ASIM-3302: Adapted the check of the message type to adhere to the f…

spyrostz:
…ormat of the aggregator trade event message.


hannesdiedrich (approved):
LGTM

PR #107, opened 2021-02-16
 Fix bug with list dso stats

Hassan754:


d3a_api_client/rest_market.py
```diff
@@ -53,7 +53,12 @@ def grid_fees(self, fee_cents_per_kWh):
 
     @logging_decorator('dso_market_stats')
     def list_dso_market_stats(self, selected_markets):
-        transaction_id, posted = self._get_request('dso-market-stats', {"market_slots": selected_markets})
+        # TODO: this is kept for the backward compatibility, should be removed
+        return self.last_market_dso_stats()
```
spyrostz, line 57: Can you explain a bit further the rationale behind that ? As far as I see, the selected_markets parameter is passed but never used in the method.

integration_tests/test_aggregator_batch_commands.py
```diff
@@ -40,18 +41,24 @@ def on_market_cycle(self, market_info):
...
                         device_event["device_info"]["energy_requirement_kWh"] > 0.0:
                     self.add_to_batch_commands.bid_energy(area_uuid=device_event["area_uuid"], price=30,
                                                           energy=device_event["device_info"][
                                                                      "energy_requirement_kWh"] / 2) \
-                        .list_bids(area_uuid=device_event["area_uuid"]) \
-                        .last_market_stats(area_uuid=device_event["area_uuid"])
+                        .list_bids(area_uuid=device_event["area_uuid"])
+            self.add_to_batch_commands.grid_fees(area_uuid=self.redis_market.area_uuid, fee_cents_kwh=5)
+            self.add_to_batch_commands.last_market_dso_stats(self.redis_market.area_uuid)
+            self.add_to_batch_commands.last_market_stats(self.redis_market.area_uuid)
```
spyrostz, line 52:
```suggestion
                        .list_bids(area_uuid=device_event["area_uuid"])
                        .grid_fees(area_uuid=self.redis_market.area_uuid, fee_cents_kwh=5)
                        .last_market_dso_stats(self.redis_market.area_uuid)
                        .last_market_stats(self.redis_market.area_uuid)
```
Matter of preference, I like this JS-style method chaining a lot though.

d3a_api_client/commands.py
```diff
@@ -47,22 +47,22 @@ def device_info(self, area_uuid):
         return self._add_to_buffer(area_uuid, Commands.DEVICE_INFO, {})
 
     def last_market_stats(self, area_uuid):
-        return self._add_to_buffer(area_uuid, "market_stats", {})
+        return self._add_to_buffer(area_uuid, "market_stats", {"data": {}})
```
spyrostz, line 50: I think you do not need the "data" part at all here. IIRC, the "data" you probably saw in the d3a code as a key on the payload is added automatically by the requests library IIRC, therefore you should not need this here. Same goes for the rest of the methods too.

spyrostz (commented):
Some comments, will rereview once addressed.

d3a_api_client/rest_market.py
Hassan754: @spyrostz I left this is in here for backward compatibility , there are still some scripts which use the list_dso_market_stats and they pass market slots, so I did not remove the deprecated attribute to not break raise exceptions, I did add a comment though

d3a_api_client/commands.py
Hassan754: @spyrostz I couldn't find out why , but sending commands on behalf of A market takes so long and breaks if the `data` key is not sent.
In the same time sending it with tis key works like charm , still investigating anw

Hassan754: Screen Shot 2021-02-17 at 18 48 50
@spyrostz this is from before the batch commands feature , I'm still not sure why the data key is needed for the market commands and not for the regular devices ones .

Hassan754: @spyrostz Found it : 
in `src/d3a/d3a_core/redis_connections/aggregator_connection.py:158` we are loading the data and saving it inside the `pending_batch_commands`
and when processing the command (for example : 
```python
    def market_stats_callback(self, payload):
        market_stats_response_channel = f"{self.channel_prefix}/response/market_stats"
        payload_data = payload["data"] \
            if isinstance(payload["data"], dict) else json.loads(payload["data"])
```
You can see that in here we are trying to get the data key again, this is the root of problem

spyrostz: @Hassan754 thanks for taking the time to dig it out! That's okay then all good from my side

spyrostz (approved):
LGTM

hannesdiedrich (approved):
I think we decided to deprecate all the market_stats endpoints right?
So this PR only fixes this just until this is tackled, right?
LGTM

Hassan754:
> I think we decided to deprecate all the market_stats endpoints right?
> So this PR only fixes this just until this is tackled, right?
> LGTM

@hannesdiedrich you are right, these should be removed when tackling the deprecating task

PR #108, opened 2021-02-18
D3ASIM-3319: Refactored the Commands Enum and its usages

Hassan754:


faizan2590 (commented):
Please update one of the existing setup client script([potentially this one](https://github.com/gridsingularity/d3a-api-client/blob/master/d3a_api_client/setups/test_redis_aggregator.py)) with examples targeting this feature.

Hassan754:
> Please update one of the existing setup client script([potentially this one](https://github.com/gridsingularity/d3a-api-client/blob/master/d3a_api_client/setups/test_redis_aggregator.py)) with examples targeting this feature.

@faizan2590 I think you meant to comment on [this PR](https://github.com/gridsingularity/d3a-api-client/pull/109)

faizan2590:
> > Please update one of the existing setup client script([potentially this one](https://github.com/gridsingularity/d3a-api-client/blob/master/d3a_api_client/setups/test_redis_aggregator.py)) with examples targeting this feature.
> 
> @faizan2590 I think you meant to comment on [this PR](https://github.com/gridsingularity/d3a-api-client/pull/109)

No, I meant this PR where you added [commands to enum](https://github.com/gridsingularity/d3a-api-client/pull/108/files#diff-93c7a331b07f8f0dcbc7528f6aab51e85ee91308f25deb93d1e09624d66867caR12-R16) but I don't see it being used an  examples in any setup files

Hassan754:
> > > Please update one of the existing setup client script([potentially this one](https://github.com/gridsingularity/d3a-api-client/blob/master/d3a_api_client/setups/test_redis_aggregator.py)) with examples targeting this feature.
> > 
> > 
> > @faizan2590 I think you meant to comment on [this PR](https://github.com/gridsingularity/d3a-api-client/pull/109)
> 
> No, I meant this PR where you added [commands to enum](https://github.com/gridsingularity/d3a-api-client/pull/108/files#diff-93c7a331b07f8f0dcbc7528f6aab51e85ee91308f25deb93d1e09624d66867caR12-R16) but I don't see it being used as an examples in any setup files

@faizan2590 I only refactored an existing enum , I don't see the point of testing this , it's not even a feature

faizan2590:
> > > > Please update one of the existing setup client script([potentially this one](https://github.com/gridsingularity/d3a-api-client/blob/master/d3a_api_client/setups/test_redis_aggregator.py)) with examples targeting this feature.
> > > 
> > > 
> > > @faizan2590 I think you meant to comment on [this PR](https://github.com/gridsingularity/d3a-api-client/pull/109)
> > 
> > 
> > No, I meant this PR where you added [commands to enum](https://github.com/gridsingularity/d3a-api-client/pull/108/files#diff-93c7a331b07f8f0dcbc7528f6aab51e85ee91308f25deb93d1e09624d66867caR12-R16) but I don't see it being used as an examples in any setup files
> 
> @faizan2590 I only refactored an existing enum , I don't see the point of testing this , it's not even a feature

In your old PR where you [abstracted batch commands API to python functions](https://github.com/gridsingularity/d3a-api-client/pull/80/files#diff-93c7a331b07f8f0dcbc7528f6aab51e85ee91308f25deb93d1e09624d66867caR4-R11), back then `update_offer/update_bid/device_info/grid_fees/market_stats /dso_market_stats ` weren't part of it and added in this PR. Right?

If yes, then simply while adding it now, you would have tested it locally too to validate these newly added commands are functional via any client script. So all Im asking is to either update any existing client script or create one using these as a reference for user

faizan2590 (commented):
Thanks for adding integration test too, would be nice to add the newly added batch commands with at least a one liner explanation in README or as a reference in one of the example scripts

Hassan754:
> Thanks for adding integration test too, would be nice to add the newly added batch commands with at least a one liner explanation in README or as a reference in one of the example scripts

@faizan2590 the only commands that should be added are update_offer / update_bid , as other stats are deprecated
I think I'll tag @fievelk to add the documentation for these two in his PR, as for now there is no `update_offer/update_bid` functions, they only exist in aggregator commads

d3a_api_client/enums.py
```diff
@@ -9,16 +9,26 @@ class Commands(Enum):
...
+    Commands.UPDATE_BID: "update_bid",
     Commands.DELETE_OFFER: "delete_offer",
     Commands.DELETE_BID: "delete_bid",
     Commands.LIST_OFFERS: "list_offers",
     Commands.LIST_BIDS: "list_bids",
-    Commands.DEVICE_INFO: "device_info"
+    Commands.DEVICE_INFO: "device_info",
+    Commands.GRID_FEES: "grid_fees",
+    Commands.MARKET_STATS: "market_stats",
+    Commands.DSO_MARKET_STATS: "dso_market_stats"
```
hannesdiedrich, line 31: I thought we decided to deprecate these. I think this was only discussed in PRs and only mentioned in planning once. Shall I create a ticket for that @spyrostz ?

d3a_api_client/redis_client_base.py
```diff
@@ -48,7 +47,7 @@ def __init__(self, area_id, client_id, autoregister=True, redis_url='redis://loc
     def _subscribe_to_response_channels(self, pubsub_thread=None):
         channel_subs = {
             self._response_topics[c]: self._generate_command_response_callback(c)
-            for c in Commands
+            for c in Commands if c in self._response_topics
```
hannesdiedrich, line 50: Can you please explain, why you dont want to subscribe to the response channels of the "new" commands e.g. UPDATE_BID ?

hannesdiedrich (commented):
Two questions, will review again once addressed.

d3a_api_client/enums.py
Hassan754: @hannesdiedrich Yes I am aware that we are to deprecate these , but this was not in the context of this ticket

d3a_api_client/redis_client_base.py
Hassan754: Well it's not that I don't want to, I kept the existing callbacks and the main goal of the ticket was to remove the hardcoded strings from Commands class .
This is why I added the rest of commands to the enum and did not touch any logic in code
However, you are right ; I should probably add callbacks for the `update_offer`, `update_bid` , will do it

hannesdiedrich: why not the responses of all of the `Commands`?

Hassan754: I'm not sure why these showed now , why are they implemented from the d3a side and not in here? I I thought that them not being implemented is because they are not needed ?
However, I really think that this got way beyond this ticket, as we need to create callbacks for several responses , add new commands ( Both the Rest and Redis Devices/markets ) 
I suggest creating another ticket to deal with this, what do you think ?

hannesdiedrich: You dont need to create new callbacks, all responses are handled in `_wait_and_consume_command_response`.

hannesdiedrich: Or rather `_publish_and_wait`

Hassan754: @hannesdiedrich yes but I mean the methods that would need to call this function, like `def update_offer` and so on

hannesdiedrich: I see, you are right. But still I would vote for adding these in the frame of this story. @spyrostz What do you think?

Hassan754: I'm in favor of not adding any response callback for the commands , as we have callbacks for events at this moment, and we have the on_event_or_response that can deal will all responses

d3a_api_client/enums.py
spyrostz: Please create a ticket for that, these are probably not going to be needed again.

hannesdiedrich: https://gridsingularity.atlassian.net/browse/D3ASIM-3351

d3a_api_client/redis_client_base.py
spyrostz: I agree with Hannes here, the responses are needed, and you cannot get away easily with handling this on _on_event_or_response. These commands are actually the synchronous mode of operation of our client, where the execution is blocked until the command response from the d3a is received. Therefore they are very useful if a user wants to wait until one action has finished. 
You could potentially wait for everything that arrives on _on_event_or_response and then unblock only the response that matters to you, however I see a disadvantage compared to subscribing to the response. _on_event_or_response is on a different thread than the typical caller of these methods (on_market_cycle or on_tick) , therefore you would have to introduce locks / polling, whereas in this case this is already handled and no new code needs to be written.

Hassan754: @spyrostz great, one more thing- do you think we should tackle the mentioned feature in this PR or you advise creating a ticket for it?

spyrostz: @Hassan754 it is not in the context of the ticket I agree, however it should not be a big change, and this PR has not bloated yet, therefore it will be relatively easy to review. IMO let's deal with it here in order for the issue to not get deprioritized/forgotten. Thanks for dealing with it !

Hassan754: Deal 👌

d3a_api_client/commands.py
```diff
@@ -47,22 +47,21 @@ def device_info(self, area_uuid):
...
 
     def grid_fees(self, area_uuid, fee_cents_kwh):
-        return self._add_to_buffer(area_uuid, "grid_fees", {"data": {"fee_const": fee_cents_kwh}})
+        return self._add_to_buffer(area_uuid, Commands.GRID_FEES, {"data": {"fee_const": fee_cents_kwh}})
 
     def _add_to_buffer(self, area_uuid, action, args):
         if area_uuid and action:
             self._commands_buffer.append(
-                {area_uuid: {"type": command_enum_to_command_name(action)
-                if type(action) == Commands else action, **args}})
```
hannesdiedrich: I think this extra precaution could stay. Of course, it is only an extra safety-net for the programmer, but still. Thanks!

hannesdiedrich (approved):
One comment, once addressed, LGTM

hannesdiedrich (approved):
LGTM

d3a_api_client/rest_device.py
```diff
@@ -114,7 +122,13 @@ def bid_energy_rate(self, energy, rate):
             'bid', {"energy": energy, "price": rate * energy})
         if posted:
             response = self.dispatcher.wait_for_command_response('bid', transaction_id)
-            log_bid_offer_confirmation(response)
```
spyrostz, line 117: I think that you can still leave it here, or am I mistaken ?

spyrostz (approved):
One comment, LGTM

d3a_api_client/rest_device.py
```diff
@@ -114,7 +122,13 @@ def bid_energy_rate(self, energy, rate):
             'bid', {"energy": energy, "price": rate * energy})
         if posted:
             response = self.dispatcher.wait_for_command_response('bid', transaction_id)
```
Hassan754, line 124:
```suggestion
            response = self.dispatcher.wait_for_command_response('bid', transaction_id)
            log_bid_offer_confirmation(response)
```

hannesdiedrich (dismissed):
LGTM

spyrostz (approved):
LGTM

faizan2590 (approved):
LGTM

PR #109, opened 2021-02-22
D3ASIM-3335: Adapted integration tests batch command new response format

Hassan754:


faizan2590 (approved):

hannesdiedrich (approved):
Please make sure that also the redis Aggregator scripts work as expected. Also, this needs to be communicated with the client users as it will probably break their scripts. 
LGTM if all this is taken care of.

Hassan754:
> Please make sure that also the redis Aggregator scripts work as expected. Also, this needs to be communicated with the client users as it will probably break their scripts.
> LGTM if all this is taken care of.

@hannesdiedrich Spyros said that after the current tasks we're tackling for the new info_trade and market_info and so on, we will notify the users with all the changes as their scripts will be broken for sure, so I'm not willing to merge this until the other tickets are being merged

integration_tests/test_aggregator_batch_commands.py
```diff
@@ -107,12 +107,11 @@ def on_market_cycle(self, market_info):
                 if transaction is None:
                     self.errors += 1
                 else:
-                    for response in transaction["responses"]:
-                        for area_response in response:
-                            if area_response['status'] == 'error':
+                    for area_uuid, response in transaction["responses"].items():
+                        for command_dict in response:
+                            if command_dict["status"] == "error":
```
fievelk, line 112: We're moving towards using single quotes for strings whenever possible (as of our style conventions). It would be great to have this aligned among all repos!

integration_tests/test_aggregator_batch_commands.py
```diff
@@ -137,12 +136,12 @@ def on_market_cycle(self, market_info):
 
                     # The bids have been issued in the correct order
                     assert [
-                        bid['original_bid_price'] for bid in issued_bids
-                    ] == [27, 28, 29, 30]
+                               bid['original_bid_price'] for bid in issued_bids
+                           ] == [27, 28, 29, 30]
```
fievelk, line 140: Why this extra indentation?

integration_tests/test_aggregator_batch_commands.py
```diff
@@ -137,12 +136,12 @@ def on_market_cycle(self, market_info):
...
 
                     # The only two bids left are the last ones that have been issued
                     assert [bid['id'] for bid in current_bids] == \
-                        [bid['id'] for bid in issued_bids[-2:]]
+                           [bid['id'] for bid in issued_bids[-2:]]
```
fievelk, line 144: Same question about indentation here. I think we can safely follow PEP8 in these cases.
Also regarding backslashes (see this comment: https://github.com/gridsingularity/d3a-api-client/pull/94#discussion_r601273204)

fievelk: Update: In this case, PEP8 actually states that backslashes are ok with `assert` statements:
> Another such case is with assert statements

integration_tests/build_test_containers.sh
```diff
@@ -14,7 +14,7 @@ if [[ "$(docker images -q ${D3A_IMAGE_TAG} 2> /dev/null)" == "" ]]; then
   echo "Building d3a image ..." && \
   rm -rf tests/d3a && \
   cd tests/ && \
-  git clone -b feature/D3ASIM-3333 https://github.com/gridsingularity/d3a.git && \
+  git clone -b feature/D3ASIM-3335 https://github.com/gridsingularity/d3a.git && \
```
hannesdiedrich, line 17: Please revert to '3333' before merging. Thanks!

hannesdiedrich (approved):
One comment, once addressed, LGTM

fievelk (approved):

PR #110, opened 2021-02-23
D3ASIM-3325: Added small fixes.

hannesdiedrich:
Two small fixed:
1. if autoregister is set to True in the RestClient, the wesocket should be established in order to receive the response of the d3a.
2. When there is a http response, there `resp.text` is not a dict, but only a string.


spyrostz (approved):
LGTM

Hassan754 (approved):

PR #111, opened 2021-03-02
D3ASIM-3333: Adapted to new market_info, tick_info event dict format.

hannesdiedrich:


d3a_api_client/redis_aggregator.py
```diff
@@ -182,15 +183,18 @@ def execute_batch_commands(self, is_blocking=True):
                 raise RedisAPIException(f'API registration process timed out.')
 
     def _on_event_or_response(self, message):
-        logging.info(f"A new message was received. Message information: {message}")
+        log_msg = copy(message)
+        if 'grid_tree' in log_msg:
+            log_msg['grid_tree'] = '{...}'
```
spyrostz, line 188: I assume you did that in order to avoid (more than useful) verbose logging, right ?

d3a_api_client/utils.py
```diff
@@ -264,3 +264,25 @@ def log_bid_offer_confirmation(message):
...
+    return outdict
+
+
+def _flatten_info_dict(indict: dict, outdict: dict):
+    """
+    Flattens market_info/tick_info information trees
+    outdict will hold references to all area subdicts of indict
+    """
+    for area_name, area_dict in indict.items():
+        outdict[area_name] = area_dict
```
spyrostz, line 286: As discussed, in general I agree with the flattening process, however using names as a key here can be potentially dangerous, hence the research team needs to be aware of potential name collisions. In general I would feel far safer if the users used uuids to call these functions, however it depends on the research team

d3a_api_client/utils.py
```diff
@@ -264,3 +264,25 @@ def log_bid_offer_confirmation(message):
                          f"{round(energy, 2)} kWh at {price} cts/kWh")
     except Exception as e:
         logging.error(f"Logging bid/offer info failed.{e}")
+
+
+def flatten_info_dict(indict: dict) -> dict:
```
spyrostz, line 269: Another optional idea: the flattened data can only be obtained by using a function on the grid_tree data in this implementation. Another approach could be, to save the last market info and tick_info in the Client class. That way the user could automatically access the market_info using `self.market_info` and `self.flat_market_info` , same for tick_info too.

d3a_api_client/redis_aggregator.py
hannesdiedrich: Exactly. I dont think that the grid_tree would be useful on INFO level. I could add another DEBUG log message that contains it?

d3a_api_client/utils.py
hannesdiedrich: You are right. I agree. I wanted to discuss this (but got interrupted and forgot then).
Let me ask the research team for their opinion.

d3a_api_client/utils.py
hannesdiedrich: I like the latter a lot and it is straight forward too. Will add it.

d3a_api_client/aggregator.py
```diff
@@ -138,12 +141,33 @@ def execute_batch_commands(self):
...
+        uuids = [area_uuid for area_uuid, area_dict in self.latest_grid_tree_flat.items()
+                 if "area_name" in area_dict and area_dict["area_name"] == name]
+        if len(uuids) == 1:
+            return uuids[0]
+        else:
+            raise ValueError(f"There are multiple areas named {name} in the tree")
+
+    def _buffer_grid_tree(self, message):
+        self.latest_grid_tree = message["grid_tree"]
+        self.latest_grid_tree_flat = flatten_info_dict(self.latest_grid_tree)
```
spyrostz, line 154: Again a bit subjective, but how about making this method a decorator in order to be a bit prettier and not see its call from the body of _on_trade, _on_tick ... methods?

hannesdiedrich: I was trying to implement this as a decorator, but then I reached the limit of my implementation skills, because inside this function `self.latest_grid_tree` needs to be accessed and gave up. Will try it one more time when adapting to the comments here.

d3a_api_client/aggregator.py
```diff
@@ -138,12 +141,33 @@ def execute_batch_commands(self):
             self._client_command_buffer.clear()
             return self.dispatcher.wait_for_command_response('batch_commands', transaction_id)
 
+    def get_uuid_from_area_name(self, name):
+        uuids = [area_uuid for area_uuid, area_dict in self.latest_grid_tree_flat.items()
+                 if "area_name" in area_dict and area_dict["area_name"] == name]
+        if len(uuids) == 1:
+            return uuids[0]
+        else:
+            raise ValueError(f"There are multiple areas named {name} in the tree")
```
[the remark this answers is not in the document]
hannesdiedrich: I agree. Although this is premature, i think this is better like you suggested.
We could populate this mapping do this every market slot and save some effort for the client. Will do so.

d3a_api_client/redis_aggregator.py
```diff
@@ -182,49 +189,55 @@ def execute_batch_commands(self, is_blocking=True):
...
     def calculate_grid_fee(self, start_market_or_device_name: str,
                            target_market_or_device_name: str = None,
-                           fee_type: str = "next_market_fee"):
+                           fee_type: str = "current_market_fee"):
         return self.grid_fee_calculation.calculate_grid_fee(start_market_or_device_name,
                                                             target_market_or_device_name, fee_type)
 
+    def _create_area_name_uuid_mapping(self):
+        self.area_name_uuid_mapping = \
+            create_area_name_uuid_mapping_from_tree_info(self.latest_grid_tree_flat)
```
faizan2590, line 209: This method seems like an overkill as  `create_area_name_uuid_mapping_from_tree_info ` could directly be called too where `_create_area_name_uuid_mapping ` method is called.

d3a_api_client/setups/test_create_aggregator.py
```diff
@@ -24,24 +23,25 @@ def on_market_cycle(self, market_info):
...
-            logging.debug(f"Batch command placed on the new market: {response}")
+        market_uuid = self.get_uuid_from_area_name("Market")
+        for area_uuid, area_dict in self.latest_grid_tree_flat.items():
+            if "asset_info" not in area_dict or area_dict["asset_info"] is None:
+                continue
+
+            logging.info(
+                f"current_market_fee: {self.grid_fee_calculation.calculate_grid_fee(area_uuid, market_uuid)}")
+            if "available_energy_kWh" in area_dict["asset_info"] and \
+                    area_dict["asset_info"]["available_energy_kWh"] > 0.0:
```
faizan2590, line 34:
```suggestion
            if key_in_dict_and_not_none_and_greater_than_zero(area_dict["asset_info"], "available_energy_kWh")::
```

integration_tests/environment.py
```diff
@@ -1,18 +1,20 @@
 from os import system
+import os
 
 
 def before_all(context):
     system(f'docker stop $(docker ps -a -q) && docker rm $(docker ps -a -q)')
     system(f'docker network create integtestnet')
     system('bash integration_tests/build_test_containers.sh')
+    os.environ["API_CLIENT_RUN_ON_REDIS"] = "true"
```
faizan2590, line 9: It has already been set under `tox`

integration_tests/test_aggregator_batch_commands.py
```diff
@@ -70,12 +71,10 @@ def on_market_cycle(self, market_info):
                             self.errors += 1
             logging.info(f"Batch command placed on the new market")
 
-        for target_market in ["Grid", "House 1", "House 2"]:
-            self.grid_fees_market_cycle_next_market[target_market] = self.calculate_grid_fee("load", target_market)
+        self.require_grid_fees(self.grid_fees_market_cycle_next_market, "current_market_fee")
```
faizan2590, line 74: Same here to homogenise market naming convention `grid_fees_market_cycle_next_market ` vs  `current_market_fee `

faizan2590 (commented):
Left some comments, in general looks good though

integration_tests/test_aggregator_batch_commands.py
hannesdiedrich: https://github.com/gridsingularity/d3a/pull/1083/files/7e06bd6375fc898eea4e0840869abaaa9043680e#r591295763

integration_tests/environment.py
hannesdiedrich: yeah, I had some trouble with that. I wanted to run the integrationtests locally without `tox`. 
This line above didnt have any impact (only setting it directly in the shell with `export API_CLIENT_RUN_ON_REDIS="true"` worked.
Will remove the line above.

d3a_api_client/setups/test_redis_aggregator.py
```diff
@@ -14,63 +12,61 @@ def __init__(self, *args, **kwargs):
...
+        logging.debug(f"AGGREGATORS_BATCH_RESPONSE: {market_stats}")
 
 
 aggregator = AutoAggregator(
-    aggregator_name="faizan_aggregator"
+    aggregator_name="test_aggr"
 )
 
-# aggregator.delete_aggregator(is_blocking=True)
+aggregator.delete_aggregator(is_blocking=True)
```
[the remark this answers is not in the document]
hannesdiedrich: yes, remnant of testing. will revert here again.

d3a_api_client/redis_aggregator.py
hannesdiedrich: you are right, will change this.

README.md
```diff
@@ -236,7 +236,7 @@ The algorithm retrieves the path between `start_market_or_device_name` and `targ
 and accumulates all corresponding grid fees along the way. Market and device names are supported.
 `target_market_or_device_name` is optional, if left blank, only the grid fee of the 
 `start_market_or_device_name` is returned. 
-The user can chose between `next_market_fee` and `last_market_fee`, which is toggled by providing 
+The user can chose between `current_market_fee` and `last_market_fee`, which is toggled by providing
```
Hassan754, line 239:
```suggestion
The user can choose between `current_market_fee` and `last_market_fee`, which is toggled by providing 
```

d3a_api_client/aggregator.py
```diff
@@ -47,6 +49,9 @@ def __init__(self, simulation_id, domain_name, aggregator_name,
         self.aggregator_uuid = None
         self._client_command_buffer = ClientCommandBuffer()
         self._connect_to_simulation()
+        self.latest_grid_tree = {}
+        self.latest_grid_tree_flat = {}
+        self.area_name_uuid_mapping = {}
```
Hassan754, line 54: Maybe after adding `self.area_name_uuid_mapping` we can get rid of the `self.device_uuid_list` in order to not have separate  objects keeping track of the devices

d3a_api_client/utils.py
```diff
@@ -264,3 +264,54 @@ def log_bid_offer_confirmation(message):
...
+        f(self, message)
+    return wrapper
+
+
+def create_area_name_uuid_mapping_from_tree_info(latest_grid_tree_flat: dict) -> dict:
+    area_name_uuid_mapping = {}
+    for area_uuid, area_dict in latest_grid_tree_flat.items():
+        if "area_name" in area_dict:
+            if area_uuid in area_name_uuid_mapping:
+                area_name_uuid_mapping[area_dict["area_name"]].append(area_uuid)
```
Hassan754, line 314: As we agreed the are name should be unique in the grid , so why building the object as list of several areas as value ?

d3a_api_client/grid_fee_calculation.py
```diff
@@ -42,28 +39,28 @@ def _find_lowest_intersection_market(in_list, intersection):
...
+                           fee_type: str = "current_market_fee"):
         """
         Calculates grid fees along path between two assets or markets in the grid
         """
         if not self.latest_grid_stats_tree:
             logging.info("Grid fees can not be calculated because there were no grid_stats sent yet.")
             return None
 
-        if target_market_or_device_name is None:
+        if target_market_or_device_uuid is None:
```
Hassan754, line 52: Thanks for converting the selection by name to selection by uuid !

d3a_api_client/redis_aggregator.py
```diff
@@ -182,49 +189,52 @@ def execute_batch_commands(self, is_blocking=True):
...
-        else:
-            # Aggregator message
-            for individual_trade in message["content"]:
-                self._log_trade_info(individual_trade)
-        function = lambda: self.on_trade(message)
-        self.executor.submit(execute_function_util, function=function,
+        # Aggregator message
+        for individual_trade in message["trade_list"]:
+            self._log_trade_info(individual_trade)
+        self.executor.submit(execute_function_util, function=lambda: self.on_trade(message),
```
Hassan754, line 233: So we were trying to handle a `Device message` at the aggregator level, which is not possible , right ?

d3a_api_client/redis_client_base.py
```diff
@@ -251,9 +252,12 @@ def _on_market_cycle(self, msg):
...
                              function_name="on_market_cycle")
 
-    def _on_event_or_response(self, msg):
-        message = json.loads(msg["data"])
-        logging.info(f"A new message was received. Message information: {message}")
+    def _on_event_or_response(self, log_msg):
+        message = json.loads(log_msg["data"])
+        log_msg = copy(message)
+        if 'grid_tree' in log_msg:
+            log_msg['grid_tree'] = '{...}'
```
Hassan754, line 259: `log_msg['grid_tree'] = '{...}'` why would you log this  value `{...}`?

Hassan754 (commented):
Left some questions , LGTM in general. Thanks !

d3a_api_client/aggregator.py
hannesdiedrich: `self.area_name_uuid_mapping` is a `Dict(str, list)` where the uuids are in the values. For the check which devices the aggregator controls an "additional" list is OK IMO, so proposing to keep it as is, except others vote differently.

d3a_api_client/redis_aggregator.py
hannesdiedrich: Exactly.

d3a_api_client/utils.py
hannesdiedrich: We did not agree on that, AFAIR. 
You were the one that implemented the "unique names in the same parent area" , which is still the case and it is still allowed to have two areas named 'load' in the whole grid as long as they dont have the same parent.

d3a_api_client/redis_client_base.py
Hassan754: Popping it sounds nice , I vote for it

d3a_api_client/aggregator.py
Hassan754: I think we should keep it as it is unless you make the `area_name_uuid_mapping` as `Dict(str, str)` (The case where a name can exist only once in the grid

integration_tests/steps/device.py
```diff
@@ -95,7 +95,7 @@ def step_impl(context):
     # placing bids and offers on every market cycle.
     # Should stop if an error occurs or if the simulation has finished
     counter = 0  # Wait for five minutes at most
-    while context.device.errors == 0 and context.device.status != "finished" and counter < 900:
+    while context.device.errors == 0 and context.device.status != "finished" and counter < 300:
```
Hassan754, line 98: I don't think this is enough for the integration tests to finish running, but feel free to skip it for now , I'll test it locally once possible

d3a_api_client/redis_client_base.py
fievelk: @hannesdiedrich I think this suggestion wasn't addressed (in case you wanted to use `pop()`).

Hassan754 (approved):
LGTM

d3a_api_client/setups/auto_offer_bid_on_device.py
```diff
@@ -35,7 +35,7 @@ def on_trade(self, trade_info):
 
 
 # Connects one client to the load device
-load = AutoOfferBidOnMarket('load', autoregister=True)
+# load = AutoOfferBidOnMarket('load', autoregister=True)
```
fievelk, line 38: Do we want to keep this as documentation, or is it a leftover?

d3a_api_client/setups/grid_operator_api_template.py
```diff
@@ -0,0 +1,266 @@
...
+import logging
+from datetime import time
+from time import sleep
+from d3a_api_client.redis_market import RedisMarketClient
+from pendulum import from_format
+from d3a_interface.constants_limits import DATE_TIME_FORMAT, TIME_FORMAT
+from d3a_api_client.types import aggregator_client_type
+from d3a_api_client.utils import get_area_uuid_from_area_name_and_collaboration_id
+from d3a_api_client.rest_market import RestMarketClient
+import os
```
fievelk, line 18: Nit: these imports can be refactored

integration_tests/build_test_containers.sh
```diff
@@ -14,7 +14,7 @@ if [[ "$(docker images -q ${D3A_IMAGE_TAG} 2> /dev/null)" == "" ]]; then
   echo "Building d3a image ..." && \
   rm -rf tests/d3a && \
   cd tests/ && \
-  git clone https://github.com/gridsingularity/d3a.git && \
+  git clone -b feature/D3ASIM-3333 https://github.com/gridsingularity/d3a.git && \
```
fievelk, line 17: We can restore the original version of this and point to `master` now, right?

fievelk (commented):
I have one question regarding `build_test_containers.sh` and pointing to the `master` branch. Besides that, just some minor comments.

integration_tests/build_test_containers.sh
hannesdiedrich: yes, already created a local commit. Thanks!

d3a_api_client/setups/grid_operator_api_template.py
hannesdiedrich: This is code from Chris. Will not touch it, ok?

d3a_api_client/setups/auto_offer_bid_on_device.py
hannesdiedrich: Dont know why this is there, will revert it (will be removed in 3398 anyway). Thanks!

integration_tests/steps/device.py
hannesdiedrich: This is only for one integrationtest. 5 Minutes should be plenty actually. Will not be needed in 3398 neither.

fievelk (approved):
LGTM!

PR #113, opened 2021-03-05
 D3ASIM-3348: Websocket consumer for live forecasted data.

faizan2590:
In order to locally test it:

- Update the environment variables under `.envvars`
- Launch the websocket server via `python live_data/websocket_subscriber/test_scripts/ws_server.py.py`
- Launch the main script to consume messages from mqtt and ws and forward the device reading to respective CNs via `python live_data/main.py`
- To test sending the mocked data to ws, update the `producer.py` script with respective device name and its energy reading and launch via `python live_data/websocket_subscriber/test_scripts/producer.py`


d3a_api_client/constants.py
```diff
@@ -8,3 +8,4 @@
 
 DEFAULT_DOMAIN_NAME = 'https://d3aweb.gridsingularity.com'
 DEFAULT_WEBSOCKET_DOMAIN = 'wss://d3aweb.gridsingularity.com/external-ws'
+CONSUMER_WEBSOCKET_DOMAIN_NAME = 'ws://localhost:4000'
```
spyrostz, line 11: I think it is more clear named as CUSTOMER_WEBSOCKET_DOMAIN_NAME or similar

mqtt_subscriber/main.py
```diff
@@ -2,16 +2,22 @@
...
 
     # Connect to the MQTT broker
-    mqtt_connection = MQTTConnection(topic_api_client_mapping)
     try:
-        mqtt_connection.run_forever()
+        p1 = Process(target=create_live_data_consumer())
+        p1.start()
+
+        p2 = Process(target=MQTTConnection(topic_api_client_mapping).run_forever())
+        p2.start()
```
spyrostz, line 18: The problem with processes is that if one of the 2 crashes, this will not result in the mqtt_subscriber pod terminating. For instance, think of the case where the websocket process dies, and the mqtt process continues to operate. The pod will not restart, therefore the websocket process will never restart (the pod will keep on running only with the mqtt process). 
On another note, I am not sure it is going to work as expected in dev env, because the main process will terminate (there is no `join()` call for the processes as far as I see)

websocket_subscriber/live_data_consumer.py
```diff
@@ -0,0 +1,76 @@
...
+from d3a_api_client.websocket_device import WebsocketMessageReceiver
+from d3a_api_client.constants import MAX_WORKER_THREADS
+from d3a_api_client.websocket_device import retry_coroutine
+from d3a_api_client.utils import RestCommunicationMixin
+from d3a_api_client.utils import retrieve_jwt_key_from_server
+
+
+class LiveDataWebsocketMessageReceiver(WebsocketMessageReceiver):
+    def __init__(self, rest_client):
+        super().__init__(rest_client)
```
spyrostz, line 20: You can skip that, and the base class constructor will be automatically invoked

websocket_subscriber/live_data_consumer.py
```diff
@@ -0,0 +1,76 @@
...
+            logging.error(f"Received message with unknown event type: {message}")
+
+    def received_message(self, message):
+        try:
+            if "event" in message:
+                self._handle_event_message(message)
+        except Exception as e:
+            logging.error(f"Error while processing incoming message {message}. Exception {e}.\n"
+                          f"{traceback.format_exc()}")
+
```
spyrostz, line 35: Since there are no commands nor events here, there is no need for 2 separate methods. A single `received_message` method should suffice, therefore you can delete the `_handle_event_message` and immediately call `self.client.set_energy_forecast(**message['data'])` from the `received_message` method.

websocket_subscriber/live_data_consumer.py
```diff
@@ -0,0 +1,76 @@
...
+        )
+        self.websocket_thread.start()
+        self.callback_thread = ThreadPoolExecutor(max_workers=MAX_WORKER_THREADS)
+
+    @logging_decorator('set_energy_forecast')
+    def set_energy_forecast(self, *args, **kwargs):
+        self.simulation_id = kwargs['simulation_id']
+        self.device_id = kwargs['area_uuid']
+        transaction_id, posted = self._post_request('set_energy_forecast',
+                                                    {"energy_forecast": kwargs['energy_wh']})
```
spyrostz, line 76: I am fairly certain that this function is already there, otherwise the mqtt subscriber would not work. Also, here you pass the responsibility of finding out the simulation_id and area_uuid to the customer, while we can find out this information (we do the same for mqtt, we match the area names from CNs that are supposed to receive measurements, and we automatically find out the simulation_id and area_uuid)

spyrostz (commented):
Some comments, I'll have another look once implemented

integration_tests/test_aggregator_batch_commands.py
```diff
@@ -79,3 +79,11 @@ def on_tick(self, tick_info):
...
     def on_finish(self, finish_info):
         self.status = "finished"
+
+#
+# device = BatchAggregator(aggregator_name="My_aggregator")
+#
+# from time import sleep
+#
+# while True:
+#     sleep(0.1)
```
fievelk, line 89: This can be safely removed :P

fievelk (approved):
Just a minor comment about commented code. LGTM

mqtt_subscriber/main.py
faizan2590: After offline discussion

- No need to have `.join` as python library do so.
- Both `MQTT`  and `websocket`  could be launch as two separate processes via main.py script.

live_data/main.py
```diff
@@ -0,0 +1,30 @@
...
+    except Exception as e:
+        logging.error(f"WS Subscriber failed with error {e}")
+        logging.error(traceback.format_exc())
+
+    try:
+        p2 = Process(target=MQTTConnection(topic_api_client_mapping).run_forever())
+        p2.start()
+    except Exception as e:
+        logging.error(f"MQTT Subscriber failed with error {e}")
+        logging.error(traceback.format_exc())
```
spyrostz, line 26: I think you can make it a bit more DRY with a `create_process_nonblocking` function.

live_data/mqtt_subscriber/oli_broker.py
```diff
@@ -47,7 +47,10 @@ def on_message(self, client, userdata, msg):
             energy = payload["value"]
 
             # Transmit power values to CN
+            print(f"self.topic_api_client_dict: {self.topic_api_client_dict}")
+            print(f"msg.topic: {msg.topic}")
```
spyrostz, line 51: Please either convert these to debug logs or remove these before merging if you do not need them anymore

live_data/websocket_subscriber/live_data_consumer.py
```diff
@@ -0,0 +1,66 @@
...
+from d3a_api_client.websocket_device import retry_coroutine
+from d3a_api_client.rest_device import RestDeviceClient
+
+
+class LiveDataWebsocketMessageReceiver(WebsocketMessageReceiver):
+
+    def received_message(self, message):
+        try:
+            if "event" in message and message["event"] == "live_data":
+                for api_args in self.client.device_api_client_mapping[message['data']['area_name']]:
```
spyrostz, line 22: Please rename the `area_name` to the slightly more generic `device_identifier`

live_data/websocket_subscriber/live_data_consumer.py
```diff
@@ -0,0 +1,66 @@
...
+                    RestDeviceClient(**api_args).set_energy_forecast(
+                        message['data']['energy_wh'], do_not_wait=True
```
spyrostz, line 24: `energy_wh` -> energy_kWh

live_data/websocket_subscriber/live_data_consumer.py
```diff
@@ -0,0 +1,66 @@
...
+from d3a_api_client.rest_device import RestDeviceClient
+
+
+class LiveDataWebsocketMessageReceiver(WebsocketMessageReceiver):
+
+    def received_message(self, message):
+        try:
+            if "event" in message and message["event"] == "live_data":
+                for api_args in self.client.device_api_client_mapping[message['data']['area_name']]:
+                    RestDeviceClient(**api_args).set_energy_forecast(
```
spyrostz, line 23: Instead of constructing a new RestDeviceClient object for all messages, you can reuse the `self.client` IIRC

live_data/websocket_subscriber/test_scripts/ws_server.py
```diff
@@ -0,0 +1,46 @@
+import asyncio
```
spyrostz, line 1: If you want to leave it for testing it's ok, however the server I am currently implementing is quite different from this implementation (mainly because there is no need to broadcast a message, and also because I have to wait 15 mins before every transmission)

live_data/main.py
```diff
@@ -0,0 +1,30 @@
+import logging
```
spyrostz, line 1: General comment, live_data is a bit vague for the package name. A much better name IMO is `live_data_subscriber` or `live_data_consumer` to denote the fact that this service is subscribing to external data sources.

spyrostz (commented):
Some comments, LGTM once fixed though

live_data/websocket_subscriber/__init__.py
```diff
@@ -0,0 +1,6 @@
+from live_data.websocket_subscriber.live_data_consumer import LiveDataConsumer
+
+
+def create_live_data_consumer(device_api_client_mapping):
+    print(f"create_live_data_consumer")
+    live_data = LiveDataConsumer(device_api_client_mapping)
```
spyrostz, line 6: This is also not needed, a `lambda: LiveDataConsumer(device_api_client_mapping)` inside the main.py should suffice if you want to defer execution of this method because it is blocking.

live_data/websocket_subscriber/live_data_consumer.py
faizan2590: `set_energy_forecast ` is accepts value in `Wh`

live_data/websocket_subscriber/live_data_consumer.py
spyrostz: After offline discussion with @faizan2590 , this is not feasible because the same websocket is used for multiple devices, therefore the rest client will have to know to which device to send this measurement. Should remain as is.

live_data/websocket_subscriber/test_scripts/ws_server.py
faizan2590: Will leave it for future to instantly test it

live_data_subscriber/mqtt/oli_broker.py
```diff
@@ -21,10 +21,12 @@ class MQTTBrokerConnectionError(Exception):
 class MQTTConnection:
     def __init__(self, topic_api_client_dict):
         self.topic_api_client_dict = topic_api_client_dict
+        print(f"self.topic_api_client_dict: {self.topic_api_client_dict}")
```
faizan2590, line 24: To be removed before merging.

live_data_subscriber/mqtt/oli_broker.py
```diff
@@ -47,10 +49,9 @@ def on_message(self, client, userdata, msg):
             energy = payload["value"]
 
             # Transmit power values to CN
-            print(f"self.topic_api_client_dict: {self.topic_api_client_dict}")
             print(f"msg.topic: {msg.topic}")
+            print(f"self.topic_api_client_dict: {self.topic_api_client_dict}")
```
faizan2590: To be removed before merging.

spyrostz (dismissed):
LGTM

live_data_subscriber/mqtt/__init__.py
```diff
@@ -22,14 +22,15 @@
 }
 
 
-def generate_topic_api_client_args_mapping():
+def generate_api_client_args_mapping():
     sim_api_domain_name = environ["API_CLIENT_SIM_API_DOMAIN_NAME"]
     external_api_domain_name = environ["API_CLIENT_DOMAIN_NAME"]
     websocket_api_domain_name = environ["API_CLIENT_WEBSOCKET_DOMAIN_NAME"]
 
     cn_mapping = list_running_canary_networks_and_devices_with_live_data(sim_api_domain_name)
```
fievelk, line 30: If the token authentication fails or there are no Canary Networks available, the script will crash at this point because of this line: https://github.com/gridsingularity/d3a-api-client/blob/7116667243cb34097d3a63855219f72b72c5f0cc/d3a_api_client/utils.py#L214

I think it would be good to add a check in `list_running_canary_networks_and_devices_with_live_data` to make sure that:
- `data` is returned (and it's not None)
- there have been no errors during the request

live_data_subscriber/.envvars
```diff
@@ -3,12 +3,12 @@
...
 # in order to set up the API client and MQTT connections
-
 MQTT_USERNAME=""
 MQTT_PASSWORD=""
 MQTT_DOMAIN_NAME=""
 API_CLIENT_USERNAME=""
 API_CLIENT_PASSWORD=""
-API_CLIENT_DOMAIN_NAME="http://localhost:8000"
-API_CLIENT_WEBSOCKET_DOMAIN_NAME="ws://localhost:8000/external-ws"
+API_CLIENT_DOMAIN_NAME=""
```
fievelk, line 11: Is there a specific reason why we're not using something like [python-dotenv](https://github.com/theskumar/python-dotenv) here, rather than running `set -a ; source .envvars ; set +a`?

faizan2590: Its been there before this PR, I would actually ask @spyrostz of having this as a preference.

live_data_subscriber/mqtt/__init__.py
fievelk: On this line: https://github.com/gridsingularity/d3a-api-client/blob/7116667243cb34097d3a63855219f72b72c5f0cc/live_data_subscriber/mqtt/__init__.py#L37

if the `cn_mapping` is empty (when there are no CNs) the loop won't do anything, but the method will still return `topic_api_client_mapping` and `device_api_client_mapping`. So the program will keep running but it will fail later on. Do you think this should be handled somehow, or is this expected?

faizan2590: > If the token authentication fails , the script will crash at this point because of this line
To me it sounds like a good thing, as authentication must correctly done to continue reading data from energy meters and to be forwarded to respective CNs.

> If there are no Canary Networks available, the script will crash at this point because of this line
It didn't crashed for me, only that the received data from respective energy meter simply isn't forwarded as the mapping is empty due to no CN

faizan2590: > the program will keep running but it will fail later on. Do you think this should be handled somehow, or is this expected?

Thanks for prolonged testing, the culprit was at this [line](https://github.com/gridsingularity/d3a-api-client/pull/113/files/6cfe04baa216f6f066e4f4d2ace87bb3dd3f4b04?file-filters%5B%5D=.py&hide-deleted-files=true#diff-2ac8c9997c6823e9284fe32ec6bfe4666ea735e13ce528bd7be42b6f7aa094d0R43), when after some some interval CN's device mapping is refreshed in order to take into account the newly added CN, earlier it was only updating the mapping for `mqtt` but after incorporating for `ws` the returned format was changed, that failed the script later, applied a fix via this [commit](https://github.com/gridsingularity/d3a-api-client/pull/113/commits/6c5cdefbd36b131b72409de614f84d7573f19987). Feel free to give another try.

fievelk: > To me it sounds like a good thing, as authentication must be correctly done to continue reading data from energy meters and to be forwarded to respective CNs.

I agree with you that the program should stop. I think it would be better if it would exit gracefully rather than crashing, though. Do you think we could show the user a message saying that something went wrong with the authentication? This would probably involve changing this part: https://github.com/gridsingularity/d3a-api-client/blob/6c5cdefbd36b131b72409de614f84d7573f19987/d3a_api_client/utils.py#L60-L64

Rather than returning `None`, we could raise an exception and then handle it where the method is called:
```python
try:
    data = execute_graphql_request(domain_name=domain_name, query=query)
except SomeCustomAuthenticationException as ex:
    # Exit gracefully with a message
    sys.exit(ex)  # The message might need to be written differently
```

What do you think?

d3a_api_client/utils.py
```diff
@@ -60,6 +60,7 @@ def execute_graphql_request(domain_name, query, headers=None, url=None, authenti
     if authenticate:
         jwt_key = retrieve_jwt_key_from_server(domain_name)
         if jwt_key is None:
+            logging.error(f"authentication failed")
```
hannesdiedrich, line 63: Minor, but I would not leave this in "error" level because it could potentially spam our logs.

live_data_subscriber/websocket/consumer.py
```diff
@@ -0,0 +1,77 @@
...
+        self.start_websocket_connection()
+        self.run_forever()
+
+    def start_websocket_connection(self):
+        self.dispatcher = LiveDataWebsocketMessageReceiver(self)
+        self.websocket_thread = LiveDataWebsocketThread(
+            self.websockets_domain_name, self.domain_name, self.dispatcher
+        )
+        self.websocket_thread.start()
+        self.callback_thread = ThreadPoolExecutor(max_workers=MAX_WORKER_THREADS)
```
hannesdiedrich, line 72: AFAIS, this is not used, right?

d3a_api_client/utils.py
fievelk: @hannesdiedrich Isn't this logged only on the client side? Shouldn't the program fail if the authentication is not successful?

hannesdiedrich: @fievelk You are right, this function is used both for the normal client and the WSComsumer which we will run on our cluster. So, leave it as is for now. If we get spammed by this error message, we can deal with it later.

live_data_subscriber/websocket/test_scripts/ws_server.py
```diff
@@ -0,0 +1,46 @@
+import asyncio
+import logging
+import websockets
+from websockets import WebSocketServerProtocol
+
+logging.basicConfig(level=logging.INFO)
+
+
+class Server:
```
hannesdiedrich, line 9: Please find a less generic way for this class. 
I would say at least `WSServer`.

live_data_subscriber/websocket/test_scripts/ws_server.py
```diff
@@ -0,0 +1,46 @@
...
+            await self.send_to_clients(message)
+
+
+def start_ws_server():
+    logging.info(f"start_ws_server")
+    server = Server()
+    start_server = websockets.serve(server.ws_handler, 'localhost', 4000)
+    loop = asyncio.get_event_loop()
+    loop.run_until_complete(start_server)
+    loop.run_forever()
```
hannesdiedrich, line 42: Aren't these two different handlers? Or can both functionalities be combined?
Copy/Paste from the asyncio.events:
```
    def run_forever(self):
        """Run the event loop until stop() is called."""
        raise NotImplementedError

    def run_until_complete(self, future):
        """Run the event loop until a Future is done.

        Return the Future's result, or raise its exception.
        """
        raise NotImplementedError
```

live_data_subscriber/websocket/test_scripts/ws_server.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+if __name__ == '__main__':
+    start_ws_server()
```
hannesdiedrich, line 46: I dont find a place in the codediff where you actually use this whole module. Can you help me please?

live_data_subscriber/__init__.py
```diff
@@ -65,10 +68,21 @@ def generate_topic_api_client_args_mapping():
...
+                device_api_client_mapping[device_name].append(
+                    {"device_id": device_uuid, **api_client_args}
+                )
+
+    logging.info(f"Connecting to the following MQTT topics {topic_api_client_mapping}")
+    logging.info(f"Connecting to the following WS devices {device_api_client_mapping}")
+    return topic_api_client_mapping, device_api_client_mapping
+
+
+def refresh_cn_and_device_list(last_time_checked, api_client_dict, is_mqtt=False):
```
hannesdiedrich, line 83: You have `api_client_dict` as input variable, but you never alter it. Is it really needed?

hannesdiedrich: Also, I dont see the need for returning `last_time_checked`. Were is it used outside of this function?

live_data_subscriber/__init__.py
```diff
@@ -22,14 +24,15 @@
...
-def generate_topic_api_client_args_mapping():
+def generate_api_client_args_mapping():
     sim_api_domain_name = environ["API_CLIENT_SIM_API_DOMAIN_NAME"]
     external_api_domain_name = environ["API_CLIENT_DOMAIN_NAME"]
     websocket_api_domain_name = environ["API_CLIENT_WEBSOCKET_DOMAIN_NAME"]
 
     cn_mapping = list_running_canary_networks_and_devices_with_live_data(sim_api_domain_name)
 
     topic_api_client_mapping = {v: [] for _, v in allowed_devices_name_mapping.items()}
+    device_api_client_mapping = {k: [] for k, _ in allowed_devices_name_mapping.items()}
```
hannesdiedrich, line 35: so for the future, a new `allowed_devices_name_mapping` has to be created, rigth? For that we need information from the costumer, right?

hannesdiedrich: How about making this an input variable and only returning one `device_api_client_mapping` (not a tuple).
This way you can decide in `refresh_cn_and_device_list` which mapping is used.

live_data_subscriber/websocket/test_scripts/ws_server.py
hannesdiedrich: OK, thanks for the offline discussion. This is minor then (as this is only for testing)

hannesdiedrich (commented):
Left some comments, happy to review again.

live_data_subscriber/__init__.py
faizan2590: > so for the future, a new allowed_devices_name_mapping has to be created, rigth? For that we need information from the costumer, right?

May be, not sure, probably depends on client's requirement to either support both mqtt as well as ws or one of it for their devices, for later case having a separate `allowed_devices_name_mapping ` for mqtt and ws sounds better.

> How about making this an input variable...
Sorry didn't completely get you ,Input variable to what?

live_data_subscriber/websocket/test_scripts/ws_server.py
faizan2590: After testing it both were needed. Since Im implementing it for the first time and not yet an expert so may not have the best answer for your question, `run_until_complete `  is an asynchronous iterator that do the job for objects being passed whereas `run_forever ` is to hold the script from exiting, so you would see `run_forever ` in `server` and `consumer´ script but not in `producer` script.

live_data_subscriber/mqtt/__init__.py
faizan2590: Im not sure if gracefully shutting would help k8s to restart the service on its own. @spyrostz @hannesdiedrich ?

live_data_subscriber/websocket/consumer.py
faizan2590: yes, you  are right, removing it

live_data_subscriber/__init__.py
hannesdiedrich: What i meant is that you are performing the same steps for creating `device_api_client_mapping` and `topic_api_client_mapping`. The only difference in the future will be the `allowed_devices_name_mapping`, so why not making this an input variable for `generate_api_client_args_mapping`?

live_data_subscriber/__init__.py
hannesdiedrich: I see, you pulled this out of the `MQTTConnection` class to use it also in `LiveDataWebsocketMessageReceiver`. 
IMO how it is implemented now it is not really pretty.I would vote for change by reference and omit the return values here. But this violates the guideline we agreed on (prefer return values over change by reference) and may also be premature. 
So, leave as is then

live_data_subscriber/__init__.py
faizan2590: Adapted via https://github.com/gridsingularity/d3a-api-client/pull/113/commits/b80f0a53a24d3d51f97f66d5b6d470c9897248f9

live_data_subscriber/mqtt/__init__.py
spyrostz: @faizan2590 no matter how the service is terminated the service will be restarted by K8s, but I see little help in that regard, since IIRC sys.exit also raises an exception `SystemExit` , therefore the application will still terminate with an unhandled exception, and I suspect that the annoying traceback will still be present. IMO the only advantage in terminating gracefully in this case is to mute the traceback from an exception that you do not want to see. 

Give it a try, but if the traceback is still reported in this case I would personally prefer to not use sys.exit at all, because there is no gain over what we have now.

fievelk: I think I might be missing some part of the broader view for this PR: since I am running the script locally as described in the description of the PR, I'm not aware of the role of k8s. Exiting gracefully would only make sense if this script was used by a user that might need some kind of and readable output. If it's part of an automated pipeline, this is not so important I guess :)

live_data_subscriber/.envvars
spyrostz: Because this is supposed to be a convenient function for local testing, not a way to set configuration in real operation of the live_data subscriber. These env variables are all supposed to be handled by the K8 config map in the deployed live_data_subscriber, therefore the .env file would not be useful. Therefore I would be very hesitant to add a dependency + Python code that reads .env files without any need for such code in production, especially since it is an one-liner to read the file locally.

live_data_subscriber/mqtt/__init__.py
spyrostz: This is essentially the subscriber for the data reported for Simris. Running the script locally is mainly for testing purposes, and not by users other than us. The main purpose of this script is to live in our cluster, connect to the Simris server via websocket to read the measurements for their devices, and propagate these measurements to d3a.

faizan2590: `retrieve_jwt_key_from_server ` is being used in many other places, if I change it now from returning `None` to raising an exception other parts of codes has to be adapted too. I would propose to create an small ticket for that as it seems to deviates from the main purpose of this ticket.

hannesdiedrich:
@faizan2590 I meant:
Lets have only the `allowed_devices_name_mapping` decide how the `key_name` is populated in `generate_api_client_args_mapping`. For now is suggest to have two different `allowed_devices_name_mapping`, onw for the MQTT and one for the WS.
for the WS it could be a temporary `{dev_name1: dev_name1, dev_name2: dev_name2, ..}` until we know better from the customer.

live_data_subscriber/__init__.py
```diff
@@ -81,12 +60,15 @@ def ws_mapping(api_client_args, live_data_device_uuids, api_client_mapping,
...
-        if is_mqtt:
-            mqtt_mapping(api_client_args, live_data_device_uuids, api_client_mapping,
-                         uuid_name_mapping)
-        else:
-            ws_mapping(api_client_args, live_data_device_uuids, api_client_mapping,
-                       uuid_name_mapping)
+        for device_uuid in live_data_device_uuids:
+            device_name = uuid_name_mapping[device_uuid]
+            if device_name in allowed_devices_name_mapping:
+                key_name = allowed_devices_name_mapping[device_name] if is_mqtt else device_name
```
hannesdiedrich, line 66: you dont need the mqtt flag any more! That was the point actually

faizan2590: Please give this commit another look: https://github.com/gridsingularity/d3a-api-client/commit/381250b20a941f95ba8c06591e3b13b6fd21d4a5
`is_mqtt`  is no more used inside `generate_api_client_args_mapping`

live_data_subscriber/__init__.py
```diff
@@ -60,15 +47,19 @@ def generate_topic_api_client_args_mapping():
...
-    return topic_api_client_mapping
+            if device_name in allowed_device_mapping:
+                api_client_mapping[allowed_device_mapping[device_name]].append(
+                    {"device_id": device_uuid, **api_client_args}
+                )
+    return api_client_mapping
+
+
+def refresh_cn_and_device_list(last_time_checked, api_client_dict,
+                               default_api_client_map):
```
hannesdiedrich, line 60: Isnt it rather a `allowed_device_mapping` than a `default_api_client_map` ? 
```suggestion
                               allowed_device_mapping):
```

hannesdiedrich (dismissed):
One last comment, once addressed, LGTM

hannesdiedrich (approved):
LGTM

spyrostz (approved):
LGTM

PR #117, opened 2021-03-16
D3ASIM-3349: First implementation of device simulator.

spyrostz:


mqtt_subscriber/device_profile_player/main.py
```diff
@@ -0,0 +1,65 @@
...
+    @staticmethod
+    def _list_all_available_filenames_and_paths(profile_path):
+        filenames = os.listdir(profile_path)
+        return [(filename, os.path.join(profile_path, filename))
+                for filename in filenames
+                if filename.endswith(".csv")]
+
+    async def _scheduled_call(self, websocket):
+        current_timestamp = now().timestamp()
+        offset_seconds = current_timestamp % 900
```
hannesdiedrich, line 32: Wow, i thought there is already some pre implemented function in datetime or pendulum, but there indeed isn't. Thanks! (we should add a PR for this on pendulum ;-) )

mqtt_subscriber/device_profile_player/main.py
```diff
@@ -0,0 +1,65 @@
...
+        requested_datetime = from_timestamp(current_timestamp - offset_seconds)\
+            .replace(year=self._year_to_replay)
```
hannesdiedrich, line 34: This is really just a precaution, right? `sleep_until` should deal with waiting a minute to resume, right?

mqtt_subscriber/device_profile_player/main.py
```diff
@@ -0,0 +1,65 @@
...
+                "device_id": device_name
+            }
+            await websocket.send(json.dumps(data_to_send))
+
+    async def sleep_until(self, resume_at):
+        await asyncio.sleep((resume_at - now()).total_seconds())
+
+    async def run(self, websocket, _):
+        while True:
+            next_cycle_datetime = now() + duration(seconds=900)
```
hannesdiedrich, line 49: Probably `premature optimization` (as you would call it :-) )but: How about making this wait interval configurable? (only if you think it is worth it)

hannesdiedrich (dismissed):
Minor comments, LGTM anyway

mqtt_subscriber/device_profile_player/main.py
spyrostz: Actually this is something the simulator needs. At the moment, the profiles that we use contain values for past years (2020). The simulator will need to run in realtime though. Therefore, we need a way to detect which year from the profiles do we want to simulate. I know that this can be easily solved not in code, but by preprocessing the files and changing the dates to 2021, however the change was easy enough to incorporate into the code.

mqtt_subscriber/device_profile_player/main.py
spyrostz: I was also thinking about it and shrugged it off as premature optimization, but you are right, will do so. Thanks! 
Speaking of, one more thing that I have not set at the moment are CLI parameters. I know it is minor because this is not a customer-facing script, however I think it makes sense to add them since we add more configuration parameters to it. Will add this and ping you again to rereview, thanks

mqtt_subscriber/device_profile_player/main.py
hannesdiedrich: Actually I mean the `from_timestamp(current_timestamp - offset_seconds)`. Not the year_to_replay.

spyrostz: Well, this is also required too :D The simulator is supposed to have a 15-min cycle, but it is not the same as the CN, because it does not run at exactly 00:15, 00:30. 00:45 .... but at any time in between. This mimics the behavior of external measurement sources, since from what we have seen they are not 100% consistent in time. 
Therefore, assuming that the simulator cycle is on 00:20, 00:35, 00:50 ... there are no values at the profiles for these timestamps, because we quantize them to exact 15 min intervals. So I have to use the previous time slot, which makes it obligatory to map 00:20 -> 00:15, 00:35->00:30 etc.

hannesdiedrich: I am stupid. I thought that 900 seconds are one minute (OMG!) Now I fully understand and of course agree.
Please make these 900 seconds also configurable in the frame of https://github.com/gridsingularity/d3a-api-client/pull/117/files/bb36be7be7d4c05e1c8aeb9180367818e7beefc6#r595065235 Thanks a lot!

mqtt_subscriber/device_profile_player/main.py
faizan2590, line 32:
```suggestion
        offset_seconds = current_timestamp % RELOAD_CN_DEVICE_LIST_TIMEOUT_SECONDS
```

May be a better name for constant variable, just pasted it here as its also `900sec` equivalent to standard one market interval.

mqtt_subscriber/device_profile_player/main.py
```diff
@@ -0,0 +1,65 @@
...
+
+    @staticmethod
+    def _list_all_available_filenames_and_paths(profile_path):
+        filenames = os.listdir(profile_path)
+        return [(filename, os.path.join(profile_path, filename))
+                for filename in filenames
+                if filename.endswith(".csv")]
+
+    async def _scheduled_call(self, websocket):
+        current_timestamp = now().timestamp()
```
faizan2590, line 31:
```suggestion
        from time import time
        current_timestamp = time()
```
Since here timestamp being used anyway so why not skipping it via deriving it from `pendulum`

mqtt_subscriber/device_profile_player/main.py
spyrostz: Have a look again, I just committed a change that makes it configurable.

mqtt_subscriber/device_profile_player/main.py
spyrostz: You are not, it is essentially a change that hides much more than it seems from a first glance. Already made this configurable via CLI, thanks!

mqtt_subscriber/device_profile_player/main.py
spyrostz: Sure, since we plan to use UTC timezone these are probably more or less equivalent, changing this.

hannesdiedrich (approved):
LGTM

faizan2590 (approved):
LGTM

PR #120, opened 2021-03-19
D3ASIM-3210: Fixed grid_fee calculation.

hannesdiedrich:


unit_tests/test_grid_fee_calculation.py
```diff
@@ -144,9 +144,37 @@ def test_grid_fee_is_calculated_correctly_for_device_to_parent(self, fee_type):
...
             leaf_name_expected_fee = {2.1: ['House 2.1', 'Load 2.1'],
-                                      1.2: ['Load 1.2', 'House 1.2', ]}
+                                      1.2: ['Load 1.2', 'House 1.2']}
         for expected_fee, leaf_names in leaf_name_expected_fee.items():
             assert isclose(expected_fee, self.grid_fee_calc.calculate_grid_fee(
                 start_market_or_device_name=leaf_names[0],
                 target_market_or_device_name=leaf_names[1],
                 fee_type=fee_type))
+
+    @parameterized.expand([['last_market_fee'], ['next_market_fee']])
```
fievelk, line 154: Why not directly this?
```suggestion
    @parameterized.expand(['last_market_fee', 'next_market_fee'])
```

unit_tests/test_grid_fee_calculation.py
```diff
@@ -144,9 +144,37 @@ def test_grid_fee_is_calculated_correctly_for_device_to_parent(self, fee_type):
...
+        else:
+            leaf_name_expected_fee = {4.1: ['Load 2.1', 'Street 2'],
+                                      2.2: ['Load 1.2', 'Street 1']}
+        for expected_fee, leaf_names in leaf_name_expected_fee.items():
+            assert isclose(expected_fee, self.grid_fee_calc.calculate_grid_fee(
+                start_market_or_device_name=leaf_names[0],
+                target_market_or_device_name=leaf_names[1],
+                fee_type=fee_type))
+
+    @parameterized.expand([['last_market_fee']])
```
fievelk, line 168: Same as above: why not using this directly?
```suggestion
    @parameterized.expand(['last_market_fee'])
```

d3a_api_client/setups/test_create_aggregator.py
```diff
@@ -111,4 +111,4 @@ def on_finish(self, finish_info):
 last_market_stats = rest_market.last_market_stats()
 
 while not aggr.is_finished:
-    sleep(0.5)
+    sleep(0.5)
```
fievelk, line 114: As a common practice, it's recommended to always add a newline at the end of the files. That's why some IDEs have the option to automatically add it when saving. Some info I found:
- [Why should text files end with a newline?](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline/729795)
- [What's the point in adding a new line to the end of a file?](https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file)
- [Why is it recommended to have empty line in the end of a source file?](https://stackoverflow.com/questions/2287967/why-is-it-recommended-to-have-empty-line-in-the-end-of-a-source-file)

fievelk (dismissed):
Some smaller comments, LGTM.

d3a_api_client/setups/test_create_aggregator.py
hannesdiedrich: Thanks a lot! This was removed due to some copy and paste. Will add it again.

unit_tests/test_grid_fee_calculation.py
hannesdiedrich: Ah no, this should be run for both `last_market_fee` `current_market_fee`, only remnant of testing. Will adapt it.

unit_tests/test_grid_fee_calculation.py
hannesdiedrich: Let me check that

unit_tests/test_grid_fee_calculation.py
```diff
@@ -137,14 +137,42 @@ def test_grid_fee_is_calculated_correctly_for_market_to_market(self, fee_type):
...
-                                      1.2: ['Load 1.2', 'House 1.2', ]}
+                                      1.2: ['Load 1.2', 'House 1.2']}
+        for expected_fee, leaf_names in leaf_name_expected_fee.items():
+            assert isclose(expected_fee, self.grid_fee_calc.calculate_grid_fee(
+                start_market_or_device_name=leaf_names[0],
+                target_market_or_device_name=leaf_names[1],
+                fee_type=fee_type))
+
+    @parameterized.expand(['last_market_fee', 'next_market_fee'])
+    def test_grid_fee_is_calculated_correctly_for_device_to_street(self, fee_type):
```
faizan2590, line 155:
```suggestion
    def test_grid_fee_is_calculated_correctly_for_device_to_intermediate_area(self, fee_type):
```
Its better to avoid referring to a hard coded area name

faizan2590 (approved):
One enhancement comment, other than that LGTM

unit_tests/test_grid_fee_calculation.py
hannesdiedrich: Hmm, I don't agree. This is a test for a specific case. The whole module depends on the representation in the top.
And here in this test case it is important that the target area is the street and not the house. `intermediate_area` is not a correct description for the street, IMO. If you could think of a more universal way of describing "two levels up", let me know and I will change it.

fievelk (approved):

unit_tests/test_grid_fee_calculation.py
hannesdiedrich: @faizan2590 if you agree, I would push now.

faizan2590: @hannesdiedrich Please go ahead then!

PR #121, opened 2021-03-23
D3ASIM-3203: Adapted cornercase where pubsub thread is reused and cha…

spyrostz:
…nnels/patterns with same names are overwritten. Saved callback to a variable and call the already subscribed callback in case the channel is already subscribed to.


hannesdiedrich (approved):
LGTM

faizan2590 (approved):
LGTM

PR #125, opened 2021-03-30
D3ASIM-3405: fix CLI entry point script

fievelk:
The cli endpoint is currently not found after pip's installation. The issue is only reproducible while installing the package without pip's `-e` option.

## Fix

This PR fixes the issue by placing the `cli.py` module in one of the packages discovered by setuptools.


Hassan754 (approved):
LGTM

setup.py
```diff
@@ -35,7 +35,7 @@
     install_requires=REQUIREMENTS,
     entry_points={
         'console_scripts': [
-            'd3a-api-client = cli:main',
+            'd3a-api-client = d3a_api_client.cli:main',
```
hannesdiedrich, line 38: Hmm, I first thought about this too. Can you help me understand?
Because the cli.py is not part of the d3a_api_client package, it is in the root folder right?

hannesdiedrich (approved):
One question, LGTM though

setup.py
fievelk: Exactly. Since I was curious to understand how this is usually done, I've check the docker-compose repo. They put all the cli-related code in a package (`compose/cli/`) inside the main `compose` one:
https://github.com/docker/compose/blob/84afa518e8144c9620562c023c0b62dd7c0b4305/setup.py#L93-L95

fievelk: The list defined by [`packages` in `setup.py`](https://github.com/gridsingularity/d3a-api-client/blob/2c2fa27b24ad3f69004cf322c8f14413ea3396fa/setup.py#L32) states which python packages should be included in the distribution. In our case, the root folder is not a package and it doesn't belong to the list:
```python
from setuptools import find_packages

find_packages(exclude=["tests"])
# ['unit_tests', 'd3a_api_client', 'd3a_api_client.setups']
```
I think this has to do with the issue we experienced.

hannesdiedrich: @fievelk Sorry, but AFAIU this does not answer my question. 
the cli.py is not in the `d3a_api_client`, right?

fievelk: Ah sorry, I misunderstood your question. It was not in the `d3a_api_client` package before. This PR moved it inside that package.

hannesdiedrich: Screenshot 2021-03-30 at 14 28 10

Ahh, yes. These renaming or moving changes in the github diff are really hidden.

hannesdiedrich: Thanks a lot!

fievelk: > Ahh, yes. These renaming or moving changes in the github diff are really hidden.

I agree, I always miss those changes!

PR #126, opened 2021-03-31
Fix d3 asim 3218: Fixed configuration file handling.

hannesdiedrich:


d3a_api_client/aggregator.py
```diff
@@ -33,8 +34,10 @@ def _handle_event_message(self, message):
 
 class Aggregator(RestDeviceClient):
 
-    def __init__(self, aggregator_name, simulation_id=None, domain_name=DOMAIN_NAME_FROM_ENV,
-                 websockets_domain_name=WEBSOCKET_DOMAIN_NAME_FROM_ENV, accept_all_devices=True):
+    def __init__(self, aggregator_name, simulation_id=simulation_id_from_env(),
+                 domain_name=domain_name_from_env(),
+                 websockets_domain_name=websocket_domain_name_from_env(),
```
fievelk, line 39: I think it is a good idea to change the current behavior. Some special attention is needed in this case, as you would be using callables as default arguments.
> Python’s default arguments are evaluated once when the function is defined, not each time the function is called

This means that the default value returned by these functions would always be the same, even if the user decided to change the environment variable at some point. See this example:
```python
import os
os.environ['SOME_SETTING_VALUE'] = 'my cool content'

def callable_function():
    return os.getenv('SOME_SETTING_VALUE')

callable_function()
# 'my cool content'

def function_with_callable_defaults(content=callable_function()):
    print(content)

function_with_callable_defaults()
# my cool content

# Now we change the environment variable
os.environ['SOME_SETTING_VALUE'] = 'some new different content'

function_with_callable_defaults()
# my cool content
```
As you see, the content of the default argument didn't change! This can get quite nasty unfortunately.

fievelk: One common way to overcome this would be to do something like this:
```suggestion
    def __init__(self, aggregator_name, simulation_id=None, ...):
             simulation_id = simulation_id if simulation_id else simulation_id_from_env()
```

d3a_api_client/aggregator.py
```diff
@@ -50,6 +53,7 @@ def __init__(self, aggregator_name, simulation_id=None, domain_name=DOMAIN_NAME_
 
     def _connect_to_simulation(self):
         user_aggrs = self.list_aggregators()
+        print(user_aggrs)
```
fievelk: This is probably a debugging leftover :)

d3a_api_client/aggregator.py
hannesdiedrich: Actually, the result of the callable (`simulation_id_from_env()`) IS evaluated each time the `__init__` method is called. It was the whole motivation behind my change.
However, will change it to your suggestion as it more explicite.

fievelk: Wait, are you sure about this? This is surprising to me and I can't reproduce it:
```python
import os
os.environ['SOME_SETTING_VALUE'] = 'my cool content'

def callable_function():
    return os.getenv('SOME_SETTING_VALUE')

callable_function()
# 'my cool content'

class AggregatorFake:
    def __init__(self, simulation_id=callable_function()):
        self.simulation_id = simulation_id

AggregatorFake().simulation_id
# 'my cool content'

# Now we change the environment variable
os.environ['SOME_SETTING_VALUE'] = 'some new different content'

AggregatorFake().simulation_id
# 'my cool content'  (not updated)
```

hannesdiedrich: weird, you are right, can not reproduce it, too. Will go your proposed way.

d3a_api_client/cli.py
```diff
@@ -72,30 +73,51 @@ def main(log_level):
...
-    ) if run_on_redis is False else ""
+    if simulation_config_path is not None and not run_on_redis:
+        config_file_path = create_simulation_config_path(base_setup_path, simulation_config_path)
+        config = read_simulation_config_file(config_file_path)
+        os.environ["API_CLIENT_DOMAIN_NAME"] = config["domain_name"]
+        os.environ["API_CLIENT_WEBSOCKET_DOMAIN_NAME"] = config["web_socket_domain_name"]
+        os.environ["API_CLIENT_SIMULATION_ID"] = config["uuid"]
+    else:
+        os.environ["API_CLIENT_DOMAIN_NAME"] = domain_name \
+            if domain_name is not None else domain_name_from_env()
```
fievelk, line 102: This won't break if `domain_name == ''`. If that should be avoided, I'd recommend something like:
```suggestion
        os.environ["API_CLIENT_DOMAIN_NAME"] = (
            domain_name if domain_name else domain_name_from_env())
```

d3a_api_client/cli.py
```diff
@@ -72,30 +73,51 @@ def main(log_level):
...
     os.environ["API_CLIENT_RUN_ON_REDIS"] = "true" if run_on_redis else "false"
-    os.environ["SIMULATION_CONFIG_FILE_PATH"] = create_simulation_config_path(
-        base_setup_path, simulation_config_path
-    ) if run_on_redis is False else ""
+    if simulation_config_path is not None and not run_on_redis:
+        config_file_path = create_simulation_config_path(base_setup_path, simulation_config_path)
+        config = read_simulation_config_file(config_file_path)
+        os.environ["API_CLIENT_DOMAIN_NAME"] = config["domain_name"]
+        os.environ["API_CLIENT_WEBSOCKET_DOMAIN_NAME"] = config["web_socket_domain_name"]
+        os.environ["API_CLIENT_SIMULATION_ID"] = config["uuid"]
```
fievelk, line 99: I've seen you're validating these values right below, in `validate_settings_are_set_before_launch()`. However, I suppose there's the possibility that these keys are not in the dictionary at all, and a `KeyError` exception would be raised here without being catched.

hannesdiedrich: Have a look into `read_simulation_config_file`
it calls the validator `validate_api_simulation_config` which checks for the right format and content

d3a_api_client/setups/test_create_aggregator.py
```diff
@@ -32,15 +32,17 @@ def on_market_cycle(self, market_info):
             return
 
         for device_event in market_info['content']:
+            if 'device_info' not in device_event or device_event['device_info'] is None:
```
fievelk, line 35:
```suggestion
            if not device_event.get('device_info'):
```
(Just optional, a bit more concise)

unit_tests/test_simulation_config.py
```diff
@@ -20,46 +19,39 @@ def setUp(self) -> None:
...
-    def test_simulation_info_returns_the_default_when_no_json_file_path(self):
-        simulation_id, domain_name, websockets_domain_name = get_simulation_config()
-        assert simulation_id == SIMULATION_ID_FROM_ENV
-        assert domain_name == DOMAIN_NAME_FROM_ENV
-        assert websockets_domain_name == WEBSOCKET_DOMAIN_NAME_FROM_ENV
+    def test_get_sim_id_and_domain_names_returns_the_defaults(self):
+        simulation_id, domain_name, websockets_domain_name = get_sim_id_and_domain_names()
+        assert simulation_id == simulation_id_from_env()
+        assert domain_name == domain_name_from_env()
+        assert websockets_domain_name == websocket_domain_name_from_env()
```
fievelk, line 26: I think it's great to test this. One caveat is that, with this code, we're not exactly testing that `simulation_id_from_env()` and the others return the values we put in the env. We're just testing that the values returned by `get_sim_id_and_domain_names()` are the same returned by `simulation_id_from_env()` and the others. However:
> we should be "testing the interface, and NOT the implementation" - doing so will allow us to change the implementation without causing our tests to break.

This means that, regardless of `simulation_id_from_env()` and the others, we should test that:
- If we specify env variables, the values returned by `get_sim_id_and_domain_names()` will be those variables
- If we don't specify env variables, the values returned by `get_sim_id_and_domain_names()` will be the default values.

What would you think about something like this?
```suggestion
    def test_get_sim_id_and_domain_names_returns_the_defaults(self):
        os.environ["API_CLIENT_SIMULATION_ID"] = "test-simulation-id"
        os.environ["API_CLIENT_DOMAIN_NAME"] = "test-domain-name"
        os.environ["API_CLIENT_WEBSOCKET_DOMAIN_NAME"] = "test-websocket-name"
        
        simulation_id, domain_name, websockets_domain_name = get_sim_id_and_domain_names()
        assert simulation_id == "test-simulation-id"
        assert domain_name == "test-domain-name"
        assert websockets_domain_name == "test-websocket-name"
```

In this way, we don't really care about how `get_sim_id_and_domain_names()` is implemented internally. What would you think?

fievelk (commented):
Left some comments. Really useful PR in my opinion!

d3a_api_client/cli.py
fievelk: That's cool! Thanks for pointing this out!

fievelk (dismissed):
Just one minor comment. LGTM.

fievelk (approved):

Hassan754 (dismissed):
LGTM

Hassan754 (approved):

fievelk (approved):

PR #134, opened 2021-04-15
Feature/d3 asim 3398 Deprecated direct offer bid handling between client and d3a

hannesdiedrich:


d3a_api_client/setups/test_create_aggregator.py
```diff
@@ -26,21 +27,27 @@ def on_market_cycle(self, market_info):
...
-                f"current_market_fee: {self.grid_fee_calculation.calculate_grid_fee(area_uuid, market_uuid)}")
-            if key_in_dict_and_not_none_and_greater_than_zero(area_dict, "available_energy_kWh"):
-                self.add_to_batch_commands.offer_energy(area_uuid=area_uuid, price=1,
-                                                        energy=area_dict["asset_info"]["available_energy_kWh"] / 2) \
-                    .list_offers(area_uuid=area_uuid)
-            if key_in_dict_and_not_none_and_greater_than_zero(area_dict, "energy_requirement_kWh"):
-                self.add_to_batch_commands.bid_energy(area_uuid=area_uuid, price=30,
-                                                      energy=area_dict["asset_info"]["energy_requirement_kWh"] / 2) \
-                    .list_bids(area_uuid=area_uuid)
+                f"current_market_fee: "
```
fievelk, line 33:
```suggestion
                "current_market_fee: "
```

d3a_api_client/setups/test_create_aggregator.py
```diff
@@ -26,21 +27,27 @@ def on_market_cycle(self, market_info):
...
+                f"{self.grid_fee_calculation.calculate_grid_fee(area_uuid, market_uuid)}")
+            if "asset_info" not in area_dict or area_dict["asset_info"] is None:
```
fievelk, line 35: Minor syntactic sugar:
```suggestion
            if not area_dict.get("asset_info"):
```

fievelk: Ah sorry, maybe you wanted to have a different behaviour when `asset_info` is `{}`?

d3a_api_client/setups/test_create_aggregator.py
```diff
@@ -26,21 +27,27 @@ def on_market_cycle(self, market_info):
...
+                if area_uuid == market_uuid:
+                    self.add_to_batch_commands.last_market_dso_stats(area_uuid=area_uuid). \
+                        grid_fees(area_uuid=area_uuid, fee_cents_kwh=5)
+            else:
+                if key_in_dict_and_not_none_and_greater_than_zero(area_dict["asset_info"],
```
fievelk, line 40: Just wanted to mention that `0` is considered to have a `False` value. So `key_in_dict_and_not_none_and_greater_than_zero` would be the same as evaluating `.get()`:
```python
area_dict = {'asset_info': {'available_energy_kWh': 0}}
if area_dict['asset_info'].get('available_energy_kWh'):
     print('true!')
else:
    print('false!')

false!
```
and
```python
area_dict = {'asset_info': {'available_energy_kWh': 1}}
if area_dict['asset_info'].get('available_energy_kWh'):
     print('true!')
else:
    print('false!')

true!
```

d3a_api_client/setups/test_create_aggregator.py
```diff
@@ -26,21 +27,27 @@ def on_market_cycle(self, market_info):
...
+                                                                  "available_energy_kWh"):
+                    energy = area_dict["asset_info"]["available_energy_kWh"] / 2
+                    self.add_to_batch_commands.offer_energy(area_uuid=area_uuid, price=1,
+                                                            energy=energy)
+
+                if key_in_dict_and_not_none_and_greater_than_zero(area_dict["asset_info"],
+                                                                  "energy_requirement_kWh"):
```
fievelk, line 47: This is unrelated to the current PR: I've seen this repeated in several scripts. Maybe it would help if we created a separate method that makes the intention more explicit. Something like: 
```suggestion
                if self.has_available_energy(area_dict):
```
and the checks happen inside the method.

d3a_api_client/redis_client_base.py
```diff
@@ -48,29 +40,23 @@ def __init__(self, area_id, client_id, autoregister=True, redis_url='redis://loc
         if autoregister:
             self.register(is_blocking=True)
 
-    def _subscribe_to_response_channels(self, pubsub_thread=None):
-        channel_subs = {
-            self._response_topics[c]: self._generate_command_response_callback(c)
```
fievelk, line 53: Since we're deleting these lines, I think the `_generate_command_response_callback()` method can also be removed, right?
https://github.com/gridsingularity/d3a-api-client/blob/7fc1c0712c934022b16a6912891dab744e777b0b/d3a_api_client/redis_client_base.py#L107-L123

d3a_api_client/redis_client_base.py
```diff
@@ -32,14 +26,12 @@ def wrapped(self, *args, **kwargs):
 
 
 class RedisClient(APIClientInterface):
```
fievelk, line 28: I would suggest to rename this to `RedisClientBase` to have the same name of the module (makes it easier to search in the project).

d3a_api_client/rest_market.py
```diff
@@ -1,45 +1,48 @@
...
-        self.device_id = area_id
+        self.area_id = area_id
         self.simulation_id = simulation_id if simulation_id else simulation_id_from_env()
         self.domain_name = domain_name if domain_name else domain_name_from_env()
         self.websockets_domain_name = websockets_domain_name \
             if websockets_domain_name else websocket_domain_name_from_env()
-        self.jwt_token = retrieve_jwt_key_from_server(domain_name)
-        self._create_jwt_refresh_timer(domain_name)
+        self.jwt_token = retrieve_jwt_key_from_server(self.domain_name)
+        self._create_jwt_refresh_timer(self.domain_name)
```
fievelk, line 20: Good catch here!

d3a_api_client/rest_market.py
```diff
@@ -1,45 +1,48 @@
...
         self.callback_thread = ThreadPoolExecutor(max_workers=MAX_WORKER_THREADS)
-        self.aggregator_prefix = get_aggregator_prefix(domain_name, simulation_id)
-        self.active_aggregator = None
 
     @logging_decorator('select-aggregator')
     def select_aggregator(self, aggregator_uuid):
         response = blocking_post_request(f'{self.aggregator_prefix}select-aggregator/',
                                          {"aggregator_uuid": aggregator_uuid,
-                                          "device_uuid": self.device_id}, self.jwt_token)
+                                          "device_uuid": self.area_id}, self.jwt_token)
```
fievelk, line 38: Do you think it would make sense to also change the parameter name on the d3a-web side? From `device_uuid` to `area_uuid`?

integration_tests/steps/device.py
```diff
@@ -25,97 +25,77 @@ def step_impl(context, setup_file: str, d3a_options: str):
         d3a_options (str): options to be passed to the d3a run command. E.g.: "-t 1s -d 12h"
     """
     sleep(3)
-    system(f'docker run -d --name d3a-tests --env REDIS_URL=redis://redis.container:6379/ '
-           f'--net integtestnet d3a-tests -l INFO run --setup {setup_file} '
-           f'--no-export --seed 0 --enable-external-connection {d3a_options}')
+    system(f"docker run -d --name d3a-tests --env REDIS_URL=redis://redis.container:6379/ "
```
fievelk, line 28:
```suggestion
    system("docker run -d --name d3a-tests --env REDIS_URL=redis://redis.container:6379/ "
```

d3a_api_client/setups/test_create_aggregator.py
hannesdiedrich: Yes, I reused your `_can_place_bid` and `_can_place_offer` in the integrationtests, but not here.
I agree, that it makes the intention more clear. 
We should review all `test_*` scripts in `/setups` and think about how user-friendly they are. For now i thought these are only for us for testing, but really they should be "good" examples. 
Do you think this is worth a separate task?

integration_tests/device.feature
```diff
@@ -2,39 +2,21 @@ Feature: Device Tests
 
 Scenario: API client can connect successfully to a load device and perform all operations
    Given redis container is started
-   And d3a is started using setup strategy_tests.external_devices (-t 1s -s 60m)
-   When the external client is started with test_load_connection
+   And d3a is started using setup strategy_tests.external_devices (-t 60s -s 60m -d 4h --slot-length-realtime 2s)
```
fievelk, line 5: Why are these changes necessary? Were there issues with the synchronization with d3a?

d3a_api_client/setups/test_create_aggregator.py
fievelk: Actually, I think this is what happens in `test_aggregator_ess.py` with `_can_place_offer` and `_can_place_bid`.

d3a_api_client/redis_client_base.py
hannesdiedrich: You are totally right. Also in the redis market (we dont need the subscription to the responses of `"market_stats", "grid_fees", "dso_market_stats"` any more. Its handled in the batch commands. Thanks!

d3a_api_client/setups/test_create_aggregator.py
fievelk: > We should review all test_* scripts in /setups and think about how user-friendly they are. For now i thought these are only for us for testing, but really they should be "good" examples.
Do you think this is worth a separate task?

Yes, I think you're definitely right. But I wouldn't do this in the current PR: having a separate task as you suggest would be better in my opinion.

d3a_api_client/setups/test_create_aggregator.py
hannesdiedrich: No, you are right, changing it.

d3a_api_client/redis_client_base.py
hannesdiedrich: I agree! Will do it!

integration_tests/device.feature
hannesdiedrich: Yes, at least this is what i suspect. I tried to eliminate the non-deterministic behaviour of the tests as well as making them shorter. They were only working before, because they were running for a long time. Eventually, they received a market_cycle event. But if you run the d3a only with `-d 4h` and without slowdown, the market_cycle event will receive the d3a-api-client only when the d3a is already finished. 
I hope this makes sense. Happy to explain more.

d3a_api_client/redis_client_base.py
hannesdiedrich: Actually, I dont see the need that these should be two classes. The potential `RedisClientBase` will not be used anywhere else. (only in auto_offer_bid_on_every_market.py and d3a_api_client/setups/auto_offer_bid_on_device.py which I will deleted also because there are not needed any more) 
EDIT: Actually, the future `RedisClientBase` could be reused in `RedisMarketClient`, too. OI wonder how this wasn't done before. I will probably see when implementing.

d3a_api_client/rest_market.py
hannesdiedrich: Probably, yes. `area_uuid` is more general and as do not expose this to the user, it would be better. Will change it in the d3a-web.

d3a_api_client/redis_client_base.py
fievelk: When you say "two classes" you mean `RedisClient` and `RedisDeviceClient`, right? If that's the case, I agree that refactoring them into one class seems to be good give our current use.

hannesdiedrich: I am currently trying to have one Baseclass `RedisClientBase` that works as Superclass for both `RedisDeviceClient` and `RedisMarketClient`. This would be the most clear way IMO.

fievelk: Ah ok! Now I get it 👍

d3a_api_client/__init__.py
```diff
@@ -20,7 +20,7 @@ def __init__(self, area_id, *args, **kwargs):
         pass
 
     @abstractmethod
-    def register(self, is_blocking):
+    def register(self):
```
fievelk, line 23: Why did you opt for removing the parameter here? Wouldn't it be better to keep the same signature for the method across interface and implementations?

d3a_api_client/redis_client_base.py
```diff
@@ -2,57 +2,73 @@
...
+        self._subscribed_aggregator_response_cb = None
         self._subscribe_to_response_channels(pubsub_thread)
         self.executor = ThreadPoolExecutor(max_workers=MAX_WORKER_THREADS)
         if autoregister:
             self.register(is_blocking=True)
 
+    @property
+    def _channel_prefix(self):
+        return f"{self.area_slug}"
+
```
fievelk: I have the feeling this method might be removed since it's just returning `self.area_slug`. The f-string here shouldn't be needed in any case, if slugify is returning a string. What do you think?

d3a_api_client/redis_client_base.py
```diff
@@ -2,57 +2,73 @@
...
-            f"{self.area_id}/response/register_participant": self._on_register,
-            f"{self.area_id}/response/unregister_participant": self._on_unregister,
+            f'{self._channel_prefix}/response/register_participant': self._on_register,
+            f'{self._channel_prefix}/response/unregister_participant': self._on_unregister,
+            f'{self._channel_prefix}/*': self._on_event_or_response
         }
 
     def _subscribe_to_response_channels(self, pubsub_thread=None):
         self.pubsub.psubscribe(**self.channel_subs)
+        if b'aggregator_response' in self.pubsub.patterns:
```
fievelk, line 52: I'm not sure whether we could use it, but I just found this on the redis gh page:
> If all string responses from a client should be decoded, the user can specify `decode_responses=True` to `Redis.__init__`. In this case, any Redis command that returns a string type will be decoded with the encoding specified.

d3a_api_client/redis_client_base.py
```diff
@@ -2,57 +2,73 @@
...
+            self._subscribed_aggregator_response_cb = self.pubsub.patterns[b'aggregator_response']
```
fievelk, line 53: What kind of object is `self.pubsub.patterns[b'aggregator_response']`? Is it a callable?

d3a_api_client/redis_client_base.py
```diff
@@ -2,57 +2,73 @@
...
         if pubsub_thread is None:
             self.pubsub.run_in_thread(daemon=True)
 
-    def register(self, is_blocking=False):
+    def _check_buffer_message_matching_command_and_id(self, message):
+        if key_in_dict_and_not_none(message, "transaction_id"):
+            transaction_id = message["transaction_id"]
+            if not any(command in ["register", "unregister"] and "transaction_id" in data and
+                       data["transaction_id"] == transaction_id
+                       for command, data in self._blocking_command_responses.items()):
```
fievelk, line 62: Nit: this would probably deserve a separate method, to explain better what we're doing in this loop.

d3a_api_client/redis_client_base.py
```diff
@@ -2,57 +2,73 @@
...
+                raise RedisAPIException("There is no matching command response in _blocking_command_responses.")
+        else:
+            raise RedisAPIException(
+                "The answer message does not contain a valid 'transaction_id' member.")
+
+    def _check_transaction_id_cached_out(self, transaction_id):
+        return transaction_id in self._transaction_id_buffer
```
fievelk, line 69: Transactions are never popped out of `self._transaction_id_buffer` in the new `RedisClientBase`, right?

d3a_api_client/redis_client_base.py
```diff
@@ -145,14 +125,42 @@ def _on_unregister(self, msg):
...
+    def _on_event_or_response(self, msg):
+        message = json.loads(msg["data"])
+        function = lambda: self.on_event_or_response(message)
+        self.executor.submit(execute_function_util, function=function,
+                             function_name="on_event_or_response")
+
+    def select_aggregator(self, aggregator_uuid, is_blocking=True):
+
+        if not self.area_uuid:
+            assert False
```
fievelk, line 137: Nit: it would be better to explicitly raise a custom exception rather than using this strategy to raise an `AssertionError`. Since this is mostly refactoring of existing code, it's not necessary to deal with this in this PR anyway.

d3a_api_client/redis_client_base.py
```diff
@@ -145,14 +125,42 @@ def _on_unregister(self, msg):
...
+        logging.info(f"Market: {self.area_slug} is trying to select aggregator {aggregator_uuid}")
+
+        transaction_id = str(uuid.uuid4())
+        data = {"aggregator_uuid": aggregator_uuid,
+                "device_uuid": self.area_uuid,
+                "type": "SELECT",
+                "transaction_id": transaction_id}
+        self.redis_db.publish(f'aggregator', json.dumps(data))
```
fievelk, line 145: Nit: unnecessary f-string (and single quotes)

d3a_api_client/redis_device.py
```diff
@@ -1,75 +1,20 @@
...
 
-    def unselect_aggregator(self, aggregator_uuid):
-        raise NotImplementedError("unselect_aggregator hasn't been implemented yet.")
+    def unregister(self, is_blocking=True):
+        super().unregister(is_blocking)
 
-    @property
-    def _channel_prefix(self):
-        return f"{self.device_id}"
+    def on_event_or_response(self, message):
```
fievelk, line 19: I think that `on_event_or_response` is not defined as an `@abstractmethod` on the interface, so it might be possible to remove it from here and from `RedisClientBase`. Or we might declar it as abstract on the interface, so that it will be clear that we want that method to be defined everywhere.

fievelk: Actually, it looks like this method doesn't have any implementation anywhere. Couldn't we just remove it?

fievelk (approved):
LGTM in general. Just a few questions and minor points.
Two broader notes if you do some further refactoring: there are some f-strings that could be normal strings, and some single-quotes that might be converted to double-quotes.

d3a_api_client/__init__.py
hannesdiedrich: You are right, remnant of refactorization. Changing it back

d3a_api_client/redis_client_base.py
hannesdiedrich: You are right, will remove it and will use the area_slug in the channel_subs function directly.

d3a_api_client/redis_client_base.py
hannesdiedrich: Sorry, this was messed up by me in the refactorization process. I will fix the whole `_subscribe_to_response_channels` actually.

d3a_api_client/redis_client_base.py
hannesdiedrich: I wanted to do it, but then I searched for a name of  the function. It would be the same as the function here `_check_buffer_message_matching_command_and_id` . So this is the main part of the functionality. So I vote for leaving it as is.

d3a_api_client/redis_client_base.py
hannesdiedrich: Indeed! I fix this in the frame of my comment above https://github.com/gridsingularity/d3a-api-client/pull/134#discussion_r618159818

d3a_api_client/redis_client_base.py
hannesdiedrich: Indeed, will add it.

d3a_api_client/redis_device.py
hannesdiedrich: Indeed, you are right with your last comment. This in case the user likes to react in every event or repsonse.

d3a_api_client/redis_client_base.py
hannesdiedrich: Hmm, I dont mind this being here. Also I found this comment: https://github.com/andymccurdy/redis-py/issues/1006#issuecomment-820626776

d3a_api_client/redis_client_base.py
fievelk: Since tests didn't fail because of this, it means that these functions are currently not tested, right? It would be great to do it so that we can spot errors automatically in the future.

d3a_api_client/redis_client_base.py
fievelk: Yeah, this is actually tricky. Maybe we could improve the readability of the method by turning the negative statement into a positive one. So, rather than:
```python
if not any(...):
    raise RedisAPIException(...)
```
we could have:
```python
if all(...):
    return True
else:
    raise RedisAPIException(...)
```
This is sometimes referred as the "Avoid negative conditionals" rule, and I think it helps readability.
Anyway, this is minor so feel free to leave it as it is if you want.

hannesdiedrich: @fievelk `not any() != all()` 
We only want to check whether the condition is true for at least one of the responses. Not for all. So i dont see room for improvement here.

d3a_api_client/redis_client_base.py
hannesdiedrich: Good catch! This error was introduced because of a wrong implementation of `_check_transaction_id_cached_out` which should be
```
        return transaction_id not in self._transaction_id_buffer
```
Nobody complained, because the registration process was alway happening. Even though the response was not handled correctly.
As we already test all this code in the integrationtests, I can not think of a way to test this specific case. 
I fixed the above mentioned function and also some logging in the redis client. Should be working now. 
let me know if you have an Idea how to test it. Thanks a lot!

d3a_api_client/redis_client_base.py
fievelk: I think my example should work though. It's not just replacing `not any()` with `all()`, it's also moving the exception into the `else` statement.
So we say:
> "if all the messages match the condition, then return True. Otherwise, raise an exception"

The version with `not any()` is saying:
> "if there is not even one message that matches the condition, then we raise an exception."

I hope I'm not getting confused by all the negations :P

fievelk: Sorry, I was getting confused earlier. In this case, I would just use
```python
if any(...):
    return True
else:
    raise RedisAPIException(...)
```
to avoid negative conditionals. Again, it doesn't need to be changed here but it might be useful in the future. I find these conditionals clearer than the negative ones.

integration_tests/test_aggregator_ess.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+class EssAggregator(RedisAggregator):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.errors = 0
+        self.status = "running"
+        self._setup()
+        self._has_tested_bids = False
+        self._has_tested_offers = False
```
Hassan754, line 17: A reusable class for these tests defining all the repetitive properties would be great

hannesdiedrich: Great idea! Will do that.

Hassan754 (approved):
LGTM, Thanks a lot

spyrostz (dismissed):
LGTM, also make sure to adapt the README once this is merged!

hannesdiedrich:
> LGTM, also make sure to adapt the README once this is merged!

Let me do it now actually. Thanks!

spyrostz (approved):
LGTM

Hassan754 (approved):

PR #141, opened 2021-04-26
D3ASIM-3438: Validate client is up to date during the jwt obtaining p…

Hassan754:
…hase


d3a_api_client/__init__.py
```diff
@@ -1,5 +1,8 @@
 from abc import ABC, abstractmethod
 
+# *IMPORTANT*: Don't manually change the version here. Use the 'bumpversion' utility.
+__version__ = '0.11.0'
```
Hassan754, line 4: This was also added here because there is no way to import the one in the setup.py, I'll update our documentation bump command to also bump this one

setup.cfg
```diff
@@ -6,3 +6,5 @@ tag = True
 [bumpversion:file:setup.py]
 search = VERSION = "{current_version}"
 
+[bumpversion:file:d3a_api_client/__init__.py]
+search = __version__ = "{current_version}"
```
fievelk, line 10: Do you think it would be possible to just use the `__version__` variable you added in `d3a_api_client/__init__.py`, and reference it in `setup.py`? In that way we would have only one variable for it (and bumpversion would just update that one).

fievelk: Another possible way to go would be to create a `VERSION` file and read it from `setup.py` (as described in this answer on SO: https://stackoverflow.com/a/2073599/1916588).

Hassan754: @fievelk seems okay for me

Hassan754: I'll take the first proposal as we're using the __version__ in the __init__ of d3a-web as well

d3a_api_client/__init__.py
hannesdiedrich: Just a question here. Is bumbversion able to change the version here, too? I thought that it only looked into setup.py

d3a_api_client/utils.py
```diff
@@ -369,3 +370,12 @@ def read_simulation_config_file(config_file_path):
 def get_sim_id_and_domain_names():
     return simulation_id_from_env(), domain_name_from_env(), websocket_domain_name_from_env()
 
+
+def validate_client_up_to_date(response):
+    if "Api-Version" in response.headers:
+        remote_version = response.headers.get("Api-Version")
```
hannesdiedrich, line 376: When I see the corresponding d3a-web PR, the api version is saved in the `API-VERSION` member of the header, not in `Api-Version`. Can you please explain?

hannesdiedrich (commented):
Two questions, Will look into it again.

d3a_api_client/utils.py
fievelk, line 376: Nit: since you've already checked for the existance of the `"Api-Version"` key, in the second like you could also omit the `.get()` and just use `["Api-Version"]`. Or it could become something like 
```suggestion
remote_version = response.headers.get("Api-Version")
if not remote_version:
    return

if __version__ < remote_version:
    ...
```

d3a_api_client/utils.py
```diff
@@ -369,3 +370,12 @@ def read_simulation_config_file(config_file_path):
...
+        from d3a_api_client import __version__
```
fievelk, line 377: Is it necessary to have this import here, or could it be moved above at module level?

setup.cfg
```diff
@@ -3,6 +3,5 @@ current_version = 0.11.0
 commit = True
 tag = True
 
-[bumpversion:file:setup.py]
-search = VERSION = "{current_version}"
-
+[bumpversion:file:d3a_api_client/__init__.py]
+search = __version__ = "{current_version}"
```
fievelk, line 7: Nit: it would be better to restore the newline here. On PyCharm 2021 you can find the setting here:
> `Settings` > `Editor` > `General` > `Ensure every saved file ends with a line break`.

d3a_api_client/__init__.py
Hassan754: @hannesdiedrich Yes I pointed the setup.cfg to it

d3a_api_client/utils.py
Hassan754: Screen Shot 2021-04-30 at 10 16 56
@hannesdiedrich It seems all response headers are returned in this case, whatever the sent key was AuThOrIzatiON-HEADER is translated to Authorization-header, I'm not 100% sure which party is responsible for this behavior though

d3a_api_client/utils.py
fievelk: @Hassan754 upping this question in case you haven't seen it :)

Hassan754: Yes we can safely remove the local import

d3a_api_client/utils.py
hannesdiedrich: Darn, I could not find anything in the internet ( or rather I could not find appropriate search words :-) 
@fievelk Do you have an idea why `request` or something else changes these keys here?

fievelk: HTTP headers' keys are case-insensitive (see: [RFC-7230](https://tools.ietf.org/html/rfc7230) and [this SO question](https://stackoverflow.com/questions/5258977/are-http-headers-case-sensitive)).
I wouldn't be too surprised if `requests` would allow you to query the headers with both `response.headers["Api-Version"]` and `response.headers["API_VERSION"]`. It would be interesting if you could give it a try.

hannesdiedrich: Wow, thanks! I suggest to go with a consistent `"API-VERSION"` ok?

spyrostz: In order to complete the picture here, the CamelCase is just how Chrome developer tools present the headers. The protocol itself dictates the headers to be lower case before encoded https://tools.ietf.org/html/rfc7540#section-8.1.2 . Therefore all HTTP-compliant clients should be able to retrieve header values no matter the style of the header. To the actual comment I personally have no strong opinion on that, but uppercase looks more familiar to me due to habit.

spyrostz (approved):
LGTM

hannesdiedrich (dismissed):
Please ping Chris before or when you push so he is aware of the new limitation when costumers complain. Thanks!
LGTM

spyrostz (dismissed):
LGTM

d3a_api_client/__init__.py
Hassan754, line 4:
```suggestion
__version__ = '0.12.0'
```

faizan2590 (approved):

hannesdiedrich (approved):
LGTM

PR #146, opened 2021-05-19
 Fix endpoints bugs

Hassan754:


faizan2590 (approved):
LGTM

d3a_api_client/rest_device.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, area_id, simulation_id=None, domain_name=None, websockets_dom
         self.configuration_prefix = get_configuration_prefix(self.domain_name, self.simulation_id)
         self.active_aggregator = None
         self.endpoint_prefix = f"{self.domain_name}/external-connection/api/" \
-                               f"{self.simulation_id}/{self.area_id}/"
+                               f"{self.simulation_id}/{self.area_id}"
```
spyrostz, line 36: Is the trailing / really at fault here ? And if it really is, shouldn't the websocket URI also omit the trailing / ?

spyrostz (approved):
One comment, LGTM once answered/resolved.

d3a_api_client/rest_device.py
Hassan754: It is not a fault actually, this is just because later on we're calling `f"{self. endpoint_prefix}/some_suffix", so the url is becoming prefix//suffix

PR #147, opened 2021-05-25
D3ASIM-3447: Fixed REST endpoint creation.

hannesdiedrich:


d3a_api_client/aggregator.py
```diff
@@ -64,6 +64,10 @@ def __init__(self, aggregator_name, simulation_id=None, domain_name=None,
         self.latest_grid_tree_flat = {}
         self.area_name_uuid_mapping = {}
 
+    @property
+    def endpoint_prefix(self):
```
Hassan754, line 68: In the aggregator you can use the `d3a_api_client/rest_device.py:32` aggregator_prefix

hannesdiedrich: NO! This is the actual bug! Please have a look at the differences between these two.

Hassan754: This is the function that is returning the aggregator's prefix, it is the same except for the trailing slash 
```python
def get_aggregator_prefix(domain_name, simulation_id=None):
    return f"{domain_name}/external-connection/aggregator-api/{simulation_id}/"
```

hannesdiedrich: Ah yes, now we are talking, but you referenced to `d3a_api_client/rest_device.py:32` not to `d3a_api_client/utils.py:78`

Hassan754: It do call the get_aggregator_prefix from there

Hassan754 (approved):
LGTM

fievelk (approved):

PR #150, opened 2021-06-10
default_domain_name_for_extracting_area_uuid

faizan2590:


Hassan754 (dismissed):

fievelk (dismissed):

d3a_api_client/setups/extract_area_uuid_from_area_name.py
```diff
@@ -9,7 +9,7 @@
 
 # set simulation parameters
 simulation_id = str(sys.argv[3])
-domain_name = 'https://d3aweb-dev.gridsingularity.com'
+domain_name = 'https://d3aweb.gridsingularity.com'
```
hannesdiedrich, line 12: Ideally this is an environmental variable so the dev tester can set it too, right?

fievelk: Or even a CLI parameter would be good (if this script is used often).

faizan2590: Using default domain name from env, if user like can simply overwrite the env var: https://github.com/gridsingularity/d3a-api-client/pull/150/commits/5eaa2c5a78830404fbad1444436fe32c6fd36f0b

Hassan754 (approved):

d3a_api_client/utils.py
```diff
@@ -107,11 +107,11 @@ def get_area_uuid_from_area_name_and_collaboration_id(collab_id, area_name, doma
     return area_uuid
 
 
-def get_area_uuid_and_name_mapping_from_simulation_id(collab_id, domain_name):
+def get_area_uuid_and_name_mapping_from_simulation_id(collab_id):
```
fievelk, line 110: Why not keeping both options by making the domain name an optional param? For example:
```python
def get_area_uuid_and_name_mapping_from_simulation_id(collab_id, domain_name=None):
    domain_name = domain_name or domain_name_from_env()
    # ...
```

fievelk: I'm suggesting this because I think that it would be good to be able to pass an alternative value to these methods without having to change env variables.

fievelk (approved):
LGTM, just one non-blocking suggestion.

PR #155, opened 2021-07-01
API template scripts update_bid_offer deprecation

chrisd12:


Hassan754:
@chrisd12 you should add the update_existing=True to these function calls to make it behave like the deprecated update_offer function

chrisd12:
https://github.com/gridsingularity/d3a-api-client/blob/master/d3a_api_client/commands.py#L43-L44 @Hassan754 is this really required since replace_existing is True by default?

Hassan754:
Ah missed it, it's okay then
LGTM

Hassan754 (approved):

hannesdiedrich (approved):
LGTM

PR #156, opened 2021-07-06
D3ASIM-3460: Added histroical_data_api_client.

hannesdiedrich:
I added this client in order to convenience the BI team so that they can retrieve the profiles without asking the backend team. 
Do you think this should live in another repository?

I tried to think about a way of testing it, but most of the added code relies on a running backend endpoint.
Also, this is really a WIP as they will probably come up with things to improve.

### DIFF
--- a/d3a_api_client/utils.py
+++ b/d3a_api_client/utils.py
@@ -3,14 +3,11 @@ import json
 import logging
 import os
 import traceback
-import uuid
 from functools import wraps
 
 import requests
 from d3a_interface.api_simulation_config.validators import validate_api_simulation_config
-from d3a_interface.constants_limits import JWT_TOKEN_EXPIRY_IN_SECS
-from d3a_interface.utils import get_area_name_uuid_mapping, key_in_dict_and_not_none, \
-    RepeatingTimer
+from d3a_interface.utils import get_area_name_uuid_mapping, key_in_dict_and_not_none
 from sgqlc.endpoint.http import HTTPEndpoint
 from tabulate import tabulate
 
@@ -344,3 +341,26 @@ def get_name_from_area_name_uuid_mapping(area_name_uuid_mapping, asset_uuid):
             if area_uuid == asset_uuid:
                 return area_name
 
+
+def get_simulation_id_from_config_uuid(domain_name: str, config_uuid: str,
+                                       auth_headers: dict) -> str:
+    """Get simulation_id from configuration by performing graphql query."""
+    query = '''
+        query get_sim_id{
+            readConfiguration(uuid:"''' + config_uuid + '''"){
+                simulationResults
+            }
+        }
+    '''
+    ret_val = execute_graphql_request(domain_name, query, headers=auth_headers)
+    if "errors" in ret_val:
+        raise Exception(f"Error when acquiring the simulation_id {ret_val['errors']}")
+    return ret_val["data"]["readConfiguration"]["simulationResults"]
+
+
+def get_auth_headers_for_requests(domain_name: str) -> dict:
+    """
+    Get authentication header dict that can be reused for all requests with json response content.
+    """
+    jwt_token = retrieve_jwt_key_from_server(domain_name)
+    return {'Authorization': f'JWT {jwt_token}', 'Content-Type': 'application/json'}
--- a/historical_data_api_client/cli.py
+++ b/historical_data_api_client/cli.py
@@ -1,0 +1,110 @@
+"""
+Copyright 2018 Grid Singularity
+This file is part of D3A.
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""
+import json
+import logging
+import os
+from datetime import datetime
+from logging import getLogger
+
+import click
+from click.types import Choice
+from click_default_group import DefaultGroup
+from colorlog.colorlog import ColoredFormatter
+
+from d3a_api_client.utils import get_simulation_id_from_config_uuid, get_auth_headers_for_requests
+from historical_data_api_client.get_historical_data import get_historical_device_profiles
+from historical_data_api_client.utils import validate_correct_datetime_format, DATE_TIME_FORMAT
+
+log = getLogger(__name__)
+
+
+@click.group(name='d3a-api-client', cls=DefaultGroup, default='run', default_if_no_args=True,
+             context_settings={'max_content_width': 120})
+@click.option('-l', '--log-level', type=Choice(list(logging._nameToLevel.keys())), default='INFO',
+              show_default=True, help="Log level")
+def main(log_level):
+    handler = logging.StreamHandler()
+    handler.setLevel(log_level)
+    handler.setFormatter(
+        ColoredFormatter(
+            "%(log_color)s%(asctime)s.%(msecs)03d %(levelname)-8s (%(lineno)4d) %(name)-30s: "
+            "%(message)s%(reset)s",
+            datefmt="%H:%M:%S"
+        )
+    )
+    root_logger = logging.getLogger()
+    root_logger.setLevel(log_level)
+    root_logger.addHandler(handler)
+
+
+@main.command()
+@click.option('-u', '--username', type=str, help="D3A username", required=None)
+@click.option('-p', '--password', type=str, help="D3A password", required=None)
+@click.option('-c', '--configuration-id', type=str, help="configuration uuid", required=True)
+@click.option('-a', '--area-uuid', type=str, help="area uuid", required=True)
+@click.option('-d', '--domain-name', default="https://d3aweb.gridsingularity.com",
+              type=str, help="D3A domain URL")
+@click.option('-s', '--start-time', type=str, default=None,
+              help="start time (Format(YYYY-MM-DDThh:mm)")
+@click.option('-e', '--end-time', type=str, default=None,
+              help="end time (Format(YYYY-MM-DDThh:mm)")
+@click.option('-t', '--time-resolution-hours', type=int, default=1,
+              help="time resolution in hours")
+@click.option('-r', '--export-file-path', type=str, default=None,
+              help="path to export json file")
+def run(username, password, domain_name, configuration_id, area_uuid, start_time, end_time,
+        time_resolution_hours, export_file_path):
+    try:
+        if username is None:
+            if "API_CLIENT_USERNAME" not in os.environ:
+                raise Exception("Please provide a username either via cli or env variable")
+        else:
+            os.environ["API_CLIENT_USERNAME"] = username
+        if password is None:
+            if "API_CLIENT_PASSWORD" not in os.environ:
+                raise Exception("Please provide a password either via cli or env variable")
+        else:
+            os.environ["API_CLIENT_PASSWORD"] = password
+
+        if start_time is None:
+            start_time = "2018-01-01T00:00"
+        if end_time is None:
+            end_time = datetime.now().strftime(DATE_TIME_FORMAT)
+        validate_correct_datetime_format(start_time)
+        validate_correct_datetime_format(end_time)
+
+        auth_headers = get_auth_headers_for_requests(domain_name)
+        simulation_id = get_simulation_id_from_config_uuid(domain_name, configuration_id,
+                                                           auth_headers)
+        profile_data = get_historical_device_profiles(domain_name, auth_headers,
+                                                      simulation_id, area_uuid,
+                                                      start_time, end_time,
+                                                      time_resolution_hours)
+
+        if export_file_path:
+            logging.info("Exporting data to %s", export_file_path)
+            with open(export_file_path, 'w') as json_file:
+                json.dump(profile_data, json_file)
+            logging.info("Done")
+        else:
+            logging.info(f"Requested data: \n {profile_data}")
+
+    except Exception as ex:
+        logging.error("Something went wrong: %s", ex)
+
+
--- a/historical_data_api_client/get_historical_data.py
+++ b/historical_data_api_client/get_historical_data.py
@@ -1,0 +1,49 @@
+"""
+Copyright 2018 Grid Singularity
+This file is part of D3A.
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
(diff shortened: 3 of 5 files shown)


historical_data_api_client/cli.py
```diff
@@ -0,0 +1,110 @@
...
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""
+import json
+import logging
+import os
+from datetime import datetime
```
faizan2590, line 21: May be declaring common imports under `historical_data_api_client/__init__.py`
https://github.com/gridsingularity/d3a-interface/pull/230/files#r664487664

faizan2590 (approved):
LGTM

historical_data_api_client/cli.py
hannesdiedrich: This is not the point of the mentioned discussion. The __init__ is not the place for common imports. I could add a `__all__`  variable in order to more easily access modules inside of the historical_data_api package from outside. But this will probably never be the case.
But maybe you meant something else here?

historical_data_api_client/cli.py
```diff
@@ -0,0 +1,110 @@
...
+from colorlog.colorlog import ColoredFormatter
+
+from d3a_api_client.utils import get_simulation_id_from_config_uuid, get_auth_headers_for_requests
+from historical_data_api_client.get_historical_data import get_historical_device_profiles
+from historical_data_api_client.utils import validate_correct_datetime_format, DATE_TIME_FORMAT
+
+log = getLogger(__name__)
+
+
+@click.group(name='d3a-api-client', cls=DefaultGroup, default='run', default_if_no_args=True,
```
fievelk, line 36: I think the `d3a-api-client` is a copy-paste leftover, right?

historical_data_api_client/cli.py
```diff
@@ -0,0 +1,110 @@
...
+@click.option('-t', '--time-resolution-hours', type=int, default=1,
+              help="time resolution in hours")
+@click.option('-r', '--export-file-path', type=str, default=None,
+              help="path to export json file")
+def run(username, password, domain_name, configuration_id, area_uuid, start_time, end_time,
+        time_resolution_hours, export_file_path):
+    try:
+        if username is None:
+            if "API_CLIENT_USERNAME" not in os.environ:
+                raise Exception("Please provide a username either via cli or env variable")
```
fievelk, line 75: Nit: I would suggest using a more specific exception. I'm not sure whether this would be the best way to go, but we could also re-use one of `click`'s exceptions ([see docs](https://click.palletsprojects.com/en/8.0.x/api/#exceptions)).

fievelk: The same comment applies to the other raised Exceptions

d3a_api_client/utils.py
```diff
@@ -344,3 +341,26 @@ def get_name_from_area_name_uuid_mapping(area_name_uuid_mapping, asset_uuid):
...
+            }
+        }
+    '''
+    ret_val = execute_graphql_request(domain_name, query, headers=auth_headers)
+    if "errors" in ret_val:
+        raise Exception(f"Error when acquiring the simulation_id {ret_val['errors']}")
+    return ret_val["data"]["readConfiguration"]["simulationResults"]
+
+
+def get_auth_headers_for_requests(domain_name: str) -> dict:
```
Hassan754, line 361: I suggest caching this return value per domain name, you can also add ttl
```suggestion
@lru_cache()
def get_auth_headers_for_requests(domain_name: str) -> dict:
```

Hassan754 (commented):
left some comments, LGTM after them and pierpaolo's comments are addressed

historical_data_api_client/cli.py
```diff
@@ -0,0 +1,110 @@
...
+    )
+    root_logger = logging.getLogger()
+    root_logger.setLevel(log_level)
+    root_logger.addHandler(handler)
+
+
+@main.command()
+@click.option('-u', '--username', type=str, help="D3A username", required=None)
+@click.option('-p', '--password', type=str, help="D3A password", required=None)
+@click.option('-c', '--configuration-id', type=str, help="configuration uuid", required=True)
```
fievelk, line 58: I'm a bit in doubt about this: If I'm not wrong, we're using different naming for the "same" object (the UUID of a configuration). Here we use `configuration-id`, while on the other `cli.py` we use:
```python
@click.option('-s', '--simulation-id', ...)
```
I think it would be nice to pick one, even though I understand that the concept of simulation is more specific than the one of configuration. I personally think that `configuration-uuid` is a bit more pertinent, since the d3a-api-client can be used to connect to collaborations, not simulations.

historical_data_api_client/cli.py
hannesdiedrich: I know that e.g. pylint does not like the broad `Exception`. Even over the years I haven't understood the reason behind it. What does the user or maintainer gain from getting this exception:
```bash
ClickException : Please provide a username either via cli or env variable
```
or
```bash
HistoricalDataAPIException : Please provide a username either via cli or env variable
```
in comparison to this Exception:
``` bash
Exception: Please provide a username either via cli or env variable
```
Would be great if you could pitch this to me, i havent found something helpful in the internet.

d3a_api_client/utils.py
```diff
@@ -344,3 +341,26 @@ def get_name_from_area_name_uuid_mapping(area_name_uuid_mapping, asset_uuid):
...
                 return area_name
 
+
+def get_simulation_id_from_config_uuid(domain_name: str, config_uuid: str,
+                                       auth_headers: dict) -> str:
+    """Get simulation_id from configuration by performing graphql query."""
+    query = '''
+        query get_sim_id{
+            readConfiguration(uuid:"''' + config_uuid + '''"){
+                simulationResults
```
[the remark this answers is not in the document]
hannesdiedrich: Please try it out in the graphiql interface. The `simulationResults` contains the simlation_id as a string. This is heritage from long time ago and I understand that it can be confusing. We could rename it in the d3a-web once we get rid of d3a.io.

d3a_api_client/utils.py
hannesdiedrich: I dont know how often this method will be called. Probably only 10 times in a row on one day.
But I can add it.

historical_data_api_client/cli.py
```diff
@@ -0,0 +1,110 @@
...
+        if username is None:
+            if "API_CLIENT_USERNAME" not in os.environ:
+                raise Exception("Please provide a username either via cli or env variable")
+        else:
+            os.environ["API_CLIENT_USERNAME"] = username
+        if password is None:
+            if "API_CLIENT_PASSWORD" not in os.environ:
+                raise Exception("Please provide a password either via cli or env variable")
+        else:
+            os.environ["API_CLIENT_PASSWORD"] = password
```
fievelk, line 82: I would suggest to only fetch env variables at the entrypoint of the code (i.e. here in `cli.py`) and to just pass them around as parameters. This would make testing easier, and it would make the code clearer in my opinion.
At the moment, this code doesn't tell us where the username and password will be used. If this could be avoided, one possible refactoring could be:
```suggestion
        username = username or os.getenv("API_CLIENT_USERNAME")
        password = password or os.getenv("API_CLIENT_PASSWORD")
        if not username:
            raise Exception("Please provide a username either via cli or env variable")
        if not password:
            raise Exception("Please provide a password either via cli or env variable")
            
        # And later pass the variables to the methods that use them
```

historical_data_api_client/cli.py
hannesdiedrich: I am a bit confused by your justification and just to be 100 sure, I like to explain again what the difference between simulation_id and configuration_uuid are (you probably knot this, so sorry if this is too verbose):
**configuration_uuid**: Is the uuid of the configuration of any simulation/collaboration/CN
**simulation_id**: (aka jobId) Is the uuid of the simulation/collaboration/CN run and is used for selecting the results of the wanted run of the d3a simulation. Earlier we allowed multiple runs of each simulation and that is why we have those two and not only one uuid. (it will be a major change to get rid of one of them).

Coming to your suggestion: I agree, that we should use the same nomenclature for bot api-clients although the user has to really provide a configuration_uuid here. I wanted to make it easy for them and let them copy/paste the uuid from the URL of the UI. But you are right, I will rename the user-facing parameter and use config_uuid inside of the code. Thanks!

historical_data_api_client/cli.py
hannesdiedrich: The reason for not assigning a variable with these values is that `retrieve_jwt_key_from_server` is only using the env variables for authentication. This way we dont pass around secret information and the user can also set the username and password in the environment without passing it int he cli (which is normally frowned up on AFIAK, but we added it also in the other api-client in order to convenience the user). 
So in order to except both ways of providing the username and password, we need to go my way.
Are you convinced yet? ;-)

historical_data_api_client/cli.py
```diff
@@ -0,0 +1,110 @@
...
+
+        if start_time is None:
+            start_time = "2018-01-01T00:00"
+        if end_time is None:
+            end_time = datetime.now().strftime(DATE_TIME_FORMAT)
```
fievelk, line 87: Nit: not necessary, but in some case you can save space with something like:
```suggestion
        start_time = start_time or "2018-01-01T00:00"
        end_time = end_time or datetime.now().strftime(DATE_TIME_FORMAT)
```

historical_data_api_client/get_historical_data.py
```diff
@@ -0,0 +1,49 @@
...
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""
+import json
+
+import requests
+
+
+def get_historical_device_profiles(domain_name: str, auth_headers: dict,
```
fievelk, line 23: Nit: when defining types, using `typing.Dict` is recommended over the bare `dict` class, because it also allows to specify the type of keys/values you want to use. For example:
```python
get_historical_device_profiles(domain_name: str, auth_headers: Dict[str, str], ...)
```

historical_data_api_client/cli.py
```diff
@@ -0,0 +1,110 @@
...
+        validate_correct_datetime_format(start_time)
+        validate_correct_datetime_format(end_time)
```
fievelk, line 89: If the validation function raises an exception, it won't be caught here. Is this expected?

historical_data_api_client/utils.py
```diff
@@ -0,0 +1,24 @@
...
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""
+from datetime import datetime
+DATE_TIME_FORMAT = "%Y-%m-%dT%H:%M"
+
+
+def validate_correct_datetime_format(time_str: str):
+    """Validate time string follows DATE_TIME_FORMAT."""
+    datetime.strptime(time_str, DATE_TIME_FORMAT)
```
fievelk, line 24: I'd suggest making this more flexible:
```suggestion
def validate_correct_datetime_format(time_str: str, time_format: str):
    """Validate that time string respects the provided time format."""
    datetime.strptime(time_str, time_format)
```
And then you can pass the `DATE_TIME_FORMAT` from the `cli.py` module. E.g.:
```python
DATE_TIME_FORMAT = "%Y-%m-%dT%H:%M"
# ...
utils.validate_correct_datetime_format(start_time, DATE_TIME_FORMAT)
utils.validate_correct_datetime_format(end_time, DATE_TIME_FORMAT)
```

fievelk (commented):
Some comments. If this module is only going to be used internally by GSy, I think it would be better to move it to a different repo because it would be of no interest for the end-users.

historical_data_api_client/cli.py
faizan2590: >  I could add a __all__ variable in order to more easily access modules inside of the historical_data_api package from outside.
Yes thats what I meant

> But this will probably never be the case.

Skip it then, if you think so

historical_data_api_client/cli.py
fievelk: There's some examplation [here](https://realpython.com/the-most-diabolical-python-antipattern/).

The first example that comes to my mind. Suppose I am an external developer and I want to integrate the `d3a-api-client` in my own project. I might have some code like:
```python
from d3a_api_client import cli

def some_complex_function():
    data = cli.run(...)
    preprocess_data(data)

def main():
    some_complex_function()    
```
Suppose I get an exception while running `main()`. If the exception is related to the `d3a_api_client` code, I want to retry with the connection. If the exception is raised by my own preprocessing function, I want to stop and gracefully exit. 
How can I distinguish between the two cases? If the client gives me a proper exception, that's simple:
```python
def main():
    try:
        some_complex_function()
    except D3AApiClientException as ex:
        # The exception comes from the client! Try reconnecting
    except MyOwnPreprocessingException as ex:
        # The exception comes from my own code! Exit or do other stuff
```
Also another example: suppose I want to check my production logs and filter them to only find the ones related to the d3a-api-client. If I have a specific exception, I can just search for it. Searching for the error message isn't as reliable.

In general, proper exception handling allows having a better understanding of what happens.

historical_data_api_client/cli.py
fievelk: TL;DR: your solution is good and backwards-compatible for this PR. In general (and for new code) I prefer passing parameters around and avoid env variables.

---

I'm partly convinced by what you wrote 😄 I see how the change I proposed would not be backwards-compatible with the previous code (which uses env variables, as you said). So I agree that your implementation is the best way to proceed without changing the existing implementation.

On the other hand, I'm not convinced that using global env variables is the best way to go. Secret information is passed around in any case (either in the `os.environ` dict or as parameter) and that probably doesn't make a difference in terms of security. Actually, env variables are visible from anywhere in the code, while parameters are only visible in the function scope.
Moreover, env variables can clash: what if another package uses the same `API_CLIENT_USERNAME` key? What if some method calls `os.environ.pop("API_CLIENT_USERNAME")` (as an extreme example)?

In any case, you can still use env variables in both cases (so the user shouldn't necessarily pass them in the cli). In my idea, the user would set up the env variables and these would be read in the `cli.py` script. From that moment on, they would be passed around as parameters, making it more explicit.

```python
# cli.py
def main():
    username = os.getenv("USERNAME")
    password = os.getenv("PASSWORD")
    retrieve_jwt_key_from_server(username=username, password=password)
```

hannesdiedrich: Ok, thanks a lot! Your point with the variable clash actually came up when I started implementing, but I shrugged it off ;-) 
Actually, I will probably do like you suggest as I will probably move this to its own repo.(https://github.com/gridsingularity/d3a-api-client/pull/156#pullrequestreview-701868057)
But I like more input for this topic first before I go that way.

fievelk: > Actually, I will probably do like you suggest as I will probably move this to its own repo.(#156 (review))

I would be in favour of this. If you go that way, I think it would be very nice to use [`python-dotenv`](https://github.com/theskumar/python-dotenv) to allow easier management of env variables.

PR #157, opened 2021-07-19
fix_requests_dependency

hannesdiedrich:


Hassan754 (dismissed):

spyrostz (approved):
LGTM, but why not recompiling the dependencies and update everything ? Requests 2.25.1 should be a dependency on d3a-interface already, so this is just dependency duplication.

faizan2590 (approved):
LGTM

hannesdiedrich:
> LGTM, but why not recompiling the dependencies and update everything ? Requests 2.25.1 should be a dependency on d3a-interface already, so this is just dependency duplication.

OK, will do.

requirements/base.txt
```diff
@@ -4,49 +4,37 @@
...
-py==1.9.0                 # via pytest, tox
+pluggy==0.13.1            # via pytest
+py==1.9.0                 # via pytest
 pycparser==2.20           # via cffi
 pynacl==1.4.0             # via paramiko
 pyparsing==2.4.7          # via packaging
 pytest==6.1.2             # via -r requirements/base.in
-redis==3.5.3              # via -r requirements/base.in, rq
 regex==2020.11.13         # via awesome-slugify
-requests==2.25.1          # via -r requirements/base.in
```
spyrostz: I see, so d3a-interface is installed via pip as a separate step and is not included in the requirements.in file. In this case all is well. We might think about harmonizing the way we import d3a-interface at some point (by adding it together with the other requirements in the .in file) however should not be done in this story since it is blocking people. I'll notify the next PR that meddles with requirements.

spyrostz (approved):
LGTM

PR #163, opened 2021-08-16
Tests 97

karthikyadav3:


unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,143 @@
...
+        assert redis_client_auto_register.area_uuid == area_uid
+        assert redis_client_auto_register.is_active == False
+        assert redis_client_auto_register._subscribed_aggregator_response_cb is None
+        assert redis_client_auto_register._subscribe_to_response_channels(pubsub_thread) is None
+        assert redis_client_auto_register._blocking_command_responses == {}
+        assert redis_client_auto_register._transaction_id_buffer == []
+
+    @patch('d3a_api_client.redis_client_base.RedisClientBase._subscribe_to_response_channels')
+    def test_function_subscribe_to_response_channels_sucess(self, mock_funct):
+        f = mock_funct()
```
hannesdiedrich, line 34: @karthikyadav3 You are calling the patched function directly and do not create an instance of `RedisClientBase` in order to call the `__init__` . Here you call the function `_subscribe_to_response_channels` again and check whether it was called. This is not the goal of the unitest

karthikyadav3: Do I need a call like this: 
``` @patch('d3a_api_client.redis_client_base.RedisClientBase')
    def test_function_subscribe_to_response_channels_sucess(self, mock_funct):
        f = mock_funct()
        mock_funct._subscribe_to_response_channel.sassert_called_once()
```

hannesdiedrich: I think so.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,176 @@
...
+    @classmethod
+    def setup_class(cls):
+        cls.obj = RedisClientBase(area_id=area_id, autoregister=False)
+
+    def test_constructor_registry_success(self, pubsub_thread=None):
+        assert self.obj.area_id == area_id
+        assert self.obj.area_uuid == area_uid
+        assert self.obj.is_active == False
+        assert self.obj._subscribed_aggregator_response_cb is None
+        assert self.obj._subscribe_to_response_channels(pubsub_thread) is None
```
faizan2590, line 34: That doesn't seems like a useful assertion. What could be actually be checked here if thread is increased by one after calling this method.

You can call [threading.active_count()](https://docs.python.org/3/library/threading.html#threading.active_count) before and after this line and assert that number of active threads have increase exactly by 1

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,176 @@
...
+
+class TestRedisClientBase:
+    @classmethod
+    def setup_class(cls):
+        cls.obj = RedisClientBase(area_id=area_id, autoregister=False)
+
+    def test_constructor_registry_success(self, pubsub_thread=None):
+        assert self.obj.area_id == area_id
+        assert self.obj.area_uuid == area_uid
+        assert self.obj.is_active == False
```
faizan2590, line 32:
```suggestion
        assert self.obj.is_active is False
```
https://www.flake8rules.com/rules/E712.html

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,176 @@
...
+
+@pytest.fixture
+def redis_client_auto_register():
+    return RedisClientBase(area_id=area_id, autoregister=False)
+
+
+class TestRedisClientBase:
+    @classmethod
+    def setup_class(cls):
+        cls.obj = RedisClientBase(area_id=area_id, autoregister=False)
```
hannesdiedrich, line 27: This is a mixture of unittest and pytest practices. Why dont you use the fixture e.g. in the `test_constructor_registry_success` ?

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,176 @@
...
+    def test_function_subscribe_to_response_channels_success(self):
+        RedisClientBase._subscribe_to_response_channels = MagicMock()
+        test = RedisClientBase(area_id=area_id, autoregister=False)
+        test._subscribe_to_response_channels.assert_called()
+        test.pubsub.psubscribe.assert_called()
+
+    def test_function_aggregator_response_callback_success(self):
+        RedisClientBase._aggregator_response_callback = MagicMock()
+        test = RedisClientBase(area_id=area_id, autoregister=False)
+        test._aggregator_response_callback()
```
hannesdiedrich, line 47: Here again, you are calling the `_aggregator_response_callback` and afterwards you are checking whether it is called. This is not the purpose of a unittest.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,176 @@
...
+        test._aggregator_response_callback.assert_called_once()
```
hannesdiedrich, line 48: IMO you can add this and Line 41 and 42 to the `test_constructor_registry_success` and dont have to initiate a RedisClientBase every time.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,176 @@
...
+
+    @pytest.fixture()
```
hannesdiedrich, line 50: Nit, please group the fixtures togesher, ideally outside of the class. Thanks!

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,176 @@
...
+        unregister_fixture.unregister.assert_called_once()
+
+    def test_function_on_register_call_success(self, on_register_fixture):
+        on_register_fixture._on_register()
+        on_register_fixture._on_register.assert_called_once()
+
+    def test_function__check_transaction_id_cached_out_call_success(self,
+                                                                    check_transaction_id_cached_out_call_success_fixture):
+        check_transaction_id_cached_out_call_success_fixture._check_transaction_id_cached_out()
+        check_transaction_id_cached_out_call_success_fixture._check_transaction_id_cached_out.assert_called_once()
```
hannesdiedrich, line 110: This comment: https://github.com/gridsingularity/d3a-api-client/pull/163/files#r691428221 applies to all of these.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,176 @@
...
+            self.redis_client_base = redis_client_auto_register
+            self.redis_client_base.is_active = True
+            self.redis_client_base.register(is_blocking=False)
+
+    def test_unregister_blocking_isactive_false_throws_exception(self, redis_client_auto_register):
+        with pytest.raises(RedisAPIException,
+                           match='API is already unregistered from the market.') as ex:
+            self.obj = redis_client_auto_register
+            self.obj.is_active = False
+            self.obj.unregister(is_blocking=True)
```
hannesdiedrich, line 146: here you are overwriting the self.obj with the fixture. No need, you can just use the fixture directly
```suggestion
            redis_client_auto_register.is_active = False
            redis_client_auto_register.unregister(is_blocking=True)
```

hannesdiedrich: The same applies to the following test functions.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,176 @@
...
+    def test_unregister_blocking_isactive_true_throws_exception(self, redis_client_auto_register):
+        with pytest.raises(RedisAPIException,
+                           match='API unregister process timed out. Server will continue processing '
+                                 'your request on the background and will notify you as soon as '
+                                 'the unregistration has been completed.') as ex:
+            self.obj = redis_client_auto_register
+            self.obj.is_active = True
+            self.obj.unregister(is_blocking=True)
+
+    def test_select_aggregator_self_area_uuid_none_throws_exception(self, redis_client_auto_register):
```
hannesdiedrich, line 157: Here you are only testing the corner cases when there is no connection to the d3a.
This is good, but you also need to test the ideal path.
As I mentioned in the offline discussion, here again an example:
If (after) the `select_aggregator(is_blocking=False)` function is called, the following action should be done, conditions should be True:
1. `self._transaction_id_buffer` should not be empty
2. `self.redis_db.publish` should have been called with the expected `data` JSON string.

hannesdiedrich (commented):
Left some comments, there are still a lot of part of the code that are note tested yet. Mainly, you need to take care of https://github.com/gridsingularity/d3a-api-client/pull/163/files#r691438158 (for all functions).
Let me know if you need help.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,176 @@
+import uuid
+from unittest import mock
+
+import pytest
+from redis import StrictRedis
+import d3a_api_client
+from d3a_api_client.redis_client_base import RedisClientBase, RedisAPIException
+from unittest.mock import MagicMock
+from unittest.mock import Mock
+from unittest.mock import patch
```
fievelk, line 10: Nit: these imports can be refactored to fit in one line:
```suggestion
from unittest.mock import MagicMock, Mock, patch
```
As a general comment: if you're using PyCharm, you can use the `Optimize Imports` feature to sort imports in the right order. You can find more information on our confluence wiki :)

unit_tests/test_redis_client_base.py
fievelk: I agree with Hannes' comment. If you want to set the fixture scope to be "class" scope you can use [fixture scopes](https://docs.pytest.org/en/latest/how-to/fixtures.html#fixture-scopes) like this:
```python
@pytest.fixture(scope="class")
def your_function():
   ...
```

fievelk: Moreover, I would recommend you to use more specific and explanatory variable names. For example, you could replace `obj` with `redis_client`.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,176 @@
...
+def redis_client_auto_register():
+    return RedisClientBase(area_id=area_id, autoregister=False)
+
+
+class TestRedisClientBase:
+    @classmethod
+    def setup_class(cls):
+        cls.obj = RedisClientBase(area_id=area_id, autoregister=False)
+
+    def test_constructor_registry_success(self, pubsub_thread=None):
```
fievelk, line 29: Nit: I think the naming of the test method could be slightly improved. One useful schema you could use is the following:
> `[MethodUnderTest]_[Scenario]_[ExpectedResult]`

In your case this could actually be something even simpler, like: `test_init()`, `test_init_succeeds()` or anything else that you might find good.

unit_tests/test_redis_client_base.py
fievelk: I agree with Faizan. Also, it is in general recommended not to test private methods/variables directly. We mostly care about the output and the side-effects of our public methods.
Suppose we unit-test a public method and we assert its output. We don't care about how many private methods are called inside this public method, as long as the output is what we expect. This saves us a lot of time and it makes the code more flexible.

I highly encourage you to have a look at this amazing talk by Sandi Metz: https://www.youtube.com/watch?v=URSWYvyc42M&t=947s

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,176 @@
...
+        assert self.obj.area_id == area_id
+        assert self.obj.area_uuid == area_uid
+        assert self.obj.is_active == False
+        assert self.obj._subscribed_aggregator_response_cb is None
+        assert self.obj._subscribe_to_response_channels(pubsub_thread) is None
+        assert self.obj._blocking_command_responses == {}
+        assert self.obj._transaction_id_buffer == []
+
+    def test_function_subscribe_to_response_channels_success(self):
```
fievelk, line 38: Nit: I suggest the shorter version
```suggestion
    def test_subscribe_to_response_channels_succeeds(self):
```
I use `succeeds` because methods should in general be verbs, not nouns.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,176 @@
...
+        RedisClientBase._subscribe_to_response_channels = MagicMock()
```
fievelk, line 39: Nit: in theory (following my comment above about testing private methods) we wouldn't need to assert that this method was called. In this specific case you might need it for other reasons, but I encourage you to give it a look and see if it is necessary.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,176 @@
...
+        test = RedisClientBase(area_id=area_id, autoregister=False)
```
fievelk, line 40:
```suggestion
        redis_client = RedisClientBase(area_id=area_id, autoregister=False)
```
Besides this, is there a reason why are you not using `self.obj` anymore here?

unit_tests/test_redis_client_base.py
fievelk: Thinking about it, I think that `_succeeds` is not even necessary. I would just call it `test_subscribe_to_response_channels()`, but that might be subjective.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,176 @@
...
+        test._aggregator_response_callback()
+        test._aggregator_response_callback.assert_called_once()
+
+    @pytest.fixture()
+    def register_fixture(self):
+        initial = RedisClientBase.register
+        RedisClientBase.register = MagicMock()
+        test = RedisClientBase(area_id=area_id, autoregister=False)
+        yield test
+        RedisClientBase.register = initial
```
fievelk, line 56: I think that these tests can be omitted. You are basically doing this:
- replace `RedisClientBase.register()` with a mock;
- call `RedisClientBase.register()`;
- assert that `RedisClientBase.register()` was called.
But this will of course pass, and it doesn't tell us anything about the state of the actual implementation (because we're basically testing the mock!). As Hannes said, this is not exactly a unit-test. I think the same applies to the other fixtures below.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,176 @@
...
+                                 'your request on the background and will notify you as soon as the '
+                                 'registration has been completed.') as ex:
+            redis_client_base = RedisClientBase(area_id=area_id, autoregister=True)
+            redis_client_base.register(is_blocking=True)
+
+    def test_register_blocking_false_published_success(self):
+        # arrange
+        redis_client_base = RedisClientBase(area_id=area_id, autoregister=False)
+        redis_client_base.register(is_blocking=False)
+        assert redis_client_base._blocking_command_responses["register"]["name"] == area_id
```
fievelk, line 132: What are we testing here? When we test a method, we mostly want to test two things (sometimes both):
1. the visible side-effects of that method (i.e. some change on the public variables);
2. the returned output of that method.

So in this case: what happens when `.register()` is called with `is_blocking=False`? In this test, you're checking the value of `redis_client_base._blocking_command_responses`. However, this is a private variable (so we probably don't need to test it). Moreover, we're not testing the other side-effect: we want to be sure that `self.redis_db.publish()` is called with the correct arguments.

I would suggest you to give it a try. The same applies to many other methods in the context of this PR. Please get in touch if you'd like to have more information and share some opinions :)

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,176 @@
...
+
+    def test_register_blocking_false_throw_exception(self, redis_client_auto_register):
+        with pytest.raises(RedisAPIException,
+                           match='API is already registered to the market.') as ex:
+            self.redis_client_base = redis_client_auto_register
```
fievelk, line 137: Why are you using an instance variable now? I'm a bit confused by the fact that you alternate the following styles in your tests:
- instance variable `self.obj`
- instance variable `self.redis_client_base` (instead of directly using the fixture)
- local variable `redis_client`

I think that local variables should be enough, but in any case it would be good to have just one approach if possible.

fievelk (commented):
In general I'm very pleased by this PR, since it helps us to greatly improve our code. So thanks!!!

Some general comments that I believe might be helpful:
- please have a look at this video: https://www.youtube.com/watch?v=URSWYvyc42M&t=948s . This will basically tell you everything I could possibly recommend about testing.
- Please use pylint on your files and fix the warnings as much as possible. Ping me on slack if you need more help with this.

Nits: 
- I would suggest removing `_function_` from the test methods' names (because it's not necessary IMO).
- Please use double quotes `"` in place of single quotes `'` for strings.
- I think we usually name methods using verbs (e.g. `test_X_succeeds` instead of `test_X_success`) but the others may have a different opinion.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,179 @@
...
+from redis import StrictRedis
+import d3a_api_client
+from d3a_api_client.redis_client_base import RedisClientBase, RedisAPIException
+from unittest.mock import MagicMock
+
+area_id = str(uuid.uuid4())
+transactionID = str(uuid.uuid4())
+d3a_api_client.redis_client_base.StrictRedis = MagicMock(spec=StrictRedis)
+d3a_api_client.redis_client_base.StrictRedis.publish = MagicMock(spec=StrictRedis)
+d3a_api_client.redis_client_base.StrictRedis.pubsub.psubscribe = MagicMock()
```
fievelk, line 13: From what I understand, you're mocking these functions because you don't want to execute the real calls while you run the individual unit tests. This is the correct approach (IMO) but it could be improved. Overriding these functions with mocks at the module level is not optimal and could lead to undesired side-effects.

A common way to do this is to use unittest's [`mock.patch()`](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch) decorator.
You can see an example on the d3a repo: https://github.com/gridsingularity/d3a/blob/3812a5f9574943b630b05999b32022bf784b631b/tests/test_strategy_home_meter.py#L88

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+@pytest.fixture
+def redis_client_auto_register():
+    return RedisClientBase(area_id=area_id, autoregister=False)
+
+
+class TestRedisClientBase:
+
+    def test_function__subscribe_to_response_channels(self, redis_client_auto_register):
```
fievelk, line 23:
```suggestion
    def test_subscribe_to_response_channels(self, redis_client_auto_register):
```

unit_tests/test_redis_client_base.py
fievelk: Another comment on this: why are you mocking directly the `RedisClientBase` class, instead of the `redis_client_auto_register` fixture?

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,179 @@
...
+        redis_client_auto_register.pubsub.psubscribe.assert_called_with(
+            **{f"{area_id}/response/register_participant": redis_client_auto_register._on_register,
+               f"{area_id}/response/unregister_participant": redis_client_auto_register._on_unregister,
+               f"{area_id}/*": redis_client_auto_register._on_event_or_response,
+               "aggregator_response": redis_client_auto_register._aggregator_response_callback})
+
+    def test_function_aggregator_response_callback(self, redis_client_auto_register):
+        message = {'data': {"transaction_id": f"{transactionID}"}}
+        message["data"] = json.dumps(message['data'])
+        redis_client = redis_client_auto_register
```
fievelk, line 35: Nit: this is not needed, you can just use `redis_client_auto_register`.

unit_tests/test_redis_client_base.py
fievelk: The same comment (to remove `_function_`) applies to the entire PR (as I had written here: https://github.com/gridsingularity/d3a-api-client/pull/163#pullrequestreview-734068112)

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,179 @@
...
+        redis_client._transaction_id_buffer.append(transactionID)
+        redis_client._aggregator_response_callback(message)
+        assert redis_client._transaction_id_buffer == []
+
+    def test_function_check_buffer_message_matching_command(self, redis_client_auto_register):
+        message = dict()
+        message["data"] = {"transaction_id": f"{transactionID}"}
```
fievelk, line 42:
```suggestion
        message = {"data": {"transaction_id": transactionID}}
```

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,179 @@
...
+        redis_client_auto_register._blocking_command_responses = {"register": {"transaction_id": f"{transactionID}"}}
+        assert redis_client_auto_register._check_buffer_message_matching_command_and_id(message['data']) is None
```
fievelk, line 44: Why are you creating a nested dictionary, if you just need the content of the value of `"data"`?

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def test_function_aggregator_response_callback(self, redis_client_auto_register):
+        message = {'data': {"transaction_id": f"{transactionID}"}}
+        message["data"] = json.dumps(message['data'])
+        redis_client = redis_client_auto_register
+        redis_client._transaction_id_buffer.append(transactionID)
+        redis_client._aggregator_response_callback(message)
+        assert redis_client._transaction_id_buffer == []
+
+    def test_function_check_buffer_message_matching_command(self, redis_client_auto_register):
```
fievelk, line 40:
```suggestion
    def test_check_buffer_message_matching_command_and_id(self, redis_client_auto_register):
```
The name of the tested method wasn't complete

fievelk:
I left some comments but I decided to pause the review for the moment. Please read the suggestions I had left in https://github.com/gridsingularity/d3a-api-client/pull/163#pullrequestreview-734068112 .
Most importantly, please use pylint on the file and fix the warnings you receive. This will greatly help with the review process. I will resume my review once this is done (unless other reviewers give their approval before).

unit_tests/test_redis_client_base.py
```diff
@@ -22,114 +19,62 @@ def redis_client_auto_register():
...
+            redis_client_auto_register._blocking_command_responses = {
+                "register": {"transaction_id": f"{transactionID}"}}
+            redis_client._check_buffer_message_matching_command_and_id(message['data'])
+
+    def test_function_register(self, redis_client_auto_register):
+        data = {"name": area_id, "transaction_id": area_id}
+        redis_client = redis_client_auto_register
+        redis_client.redis_db.publish = MagicMock()
+        redis_client.register(is_blocking=False)
+        redis_client.redis_db.publish.assert_called_once_with(f'{area_id}/register_participant', json.dumps(data))
```
hannesdiedrich, line 69: What are you testing here? 
This is the same as https://github.com/gridsingularity/d3a-api-client/pull/163#discussion_r691428221

karthikyadav3: I am testing whether the redis_db. publish method is called with correct arguments.

unit_tests/test_redis_client_base.py
karthikyadav3: Sure I will change it

unit_tests/test_redis_client_base.py
karthikyadav3: Yes, Sure I will change it

unit_tests/test_redis_client_base.py
hannesdiedrich: Sorry, it was too late yesterday. You are right. ignore me.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def test_register(self, redis_client_auto_register):
+        """checks whether the redis.db.publish is called with correct arguments"""
+        data = {"name": AREA_ID, "transaction_id": AREA_ID}
+        redis_client_auto_register.redis_db.publish = MagicMock()
+        redis_client_auto_register.register(is_blocking=False)
+        redis_client_auto_register.redis_db.publish \
+            .assert_called_once_with(f"{AREA_ID}/register_participant", json.dumps(data))
+
+    def test_register_blocking_true_throws_exception(self, redis_client_auto_register):
```
hannesdiedrich, line 91: Here the cause it not that `is_blocking=True`, but that there is no d3a running and is connected via redis. Please include this at least in the docstring. Thanks!

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,217 @@
...
+        with pytest.raises(RedisAPIException,
+                           match="API registration process timed out. "
+                                 "Server will continue processing "
+                                 "your request on the background "
+                                 "and will notify you as soon as the "
+                                 "registration has been completed."):
+            redis_client_auto_register.register(is_blocking=True)
+
+    def test_register_blocking_false_throw_exception(self, redis_client_auto_register):
+        """checks whether the if is blocking is set to false throws exception"""
```
hannesdiedrich, line 102: Here also, the cause is not that you set `is_blocking=True`, but that the user is already registered. Please adapt the function name and docstring. thanks!

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,217 @@
...
+        self.redis_client = redis_client_auto_register
+        self.redis_client.redis_db.publish = MagicMock()
+        self.redis_client.is_active = True
+        self.redis_client.unregister(is_blocking=False)
+        self.redis_client.redis_db.publish \
+            .assert_called_once_with(f"{AREA_ID}/unregister_participant",
+                                     json.dumps(data))
+
+    def test_unregister_blocking_is_active_false_throws_exception(self, redis_client_auto_register):
+        """checks if is active set to false throws exception"""
```
hannesdiedrich, line 121: https://github.com/gridsingularity/d3a-api-client/pull/163/files#r695721335 applies here too (only that the user is not already registered)

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,217 @@
...
+        with pytest.raises(RedisAPIException,
+                           match="API is already unregistered from the market."):
+            self.redis_client = redis_client_auto_register
+            self.redis_client.is_active = False
+            self.redis_client.unregister(is_blocking=True)
+
+    def test_unregister_blocking_is_active_true_throws_exception(self, redis_client_auto_register):
+        """checks if is active set to true throws exception"""
```
hannesdiedrich, line 129: also here https://github.com/gridsingularity/d3a-api-client/pull/163/files#r695719856

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,217 @@
...
+                                 "Server will continue processing "
+                                 "your request on the background and "
+                                 "will notify you as soon as "
+                                 "the unregistration has been completed."):
+            self.redis_client = redis_client_auto_register
+            self.redis_client.is_active = True
+            self.redis_client.unregister(is_blocking=True)
+
+    def test_on_register(self, redis_client_auto_register):
+        """assigning the correct arguments to blocking command response and
```
hannesdiedrich, line 141: So here you are testing whether there is no exception is raised? 
I think you need at least check if `self.is_active = True` after calling `redis_client_auto_register._on_register(message)`.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def test_on_unregister(self, redis_client_auto_register):
+        """assigning the correct arguments to blocking command response and
+        checking by calling on_unregister function with correct message"""
+        message = {"data": {"device_uuid": TRANSACTION_ID, "transaction_id": TRANSACTION_ID,
+                            "response": "success"}}
+        message["data"] = json.dumps(message["data"])
+        redis_client_auto_register._blocking_command_responses \
+            = {"unregister": {"transaction_id": f"{TRANSACTION_ID}"}}
+        redis_client_auto_register._on_unregister(message)
```
hannesdiedrich, line 157: Here https://github.com/gridsingularity/d3a-api-client/pull/163/files#r695724815 applies to, also that here `self.active` should be False

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,217 @@
...
+                = {"unregister": {"transaction_id": f"{TRANSACTION_ID}"}}
+            redis_client_auto_register._on_unregister(message)
+
+    def test_on_event_or_response(self, redis_client_auto_register):
+        """assigning the correct arguments to blocking command response and
+        checking by calling on_event_or_response function with correct message"""
+        message = {"data": {"device_uuid": TRANSACTION_ID,
+                            "transaction_id": TRANSACTION_ID}}
+        message["data"] = json.dumps(message["data"])
+        redis_client_auto_register._on_event_or_response(message)
```
hannesdiedrich, line 177: What are you really testing here? I think the only way to test `_on_event_or_response` is to check if the ThreadPoolExecutor called `on_event_or_response`, but IMO you can skip it, and thus skip this whole test here.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,217 @@
...
+        self.redis_client.area_uuid = AREA_ID
+        self.redis_client.select_aggregator(aggregator_uuid, is_blocking=False)
+        transaction_id = self.redis_client._transaction_id_buffer[0]
+        data = {"aggregator_uuid": aggregator_uuid,
+                "device_uuid": AREA_ID,
+                "type": "SELECT",
+                "transaction_id": transaction_id}
+        self.redis_client.redis_db.publish.assert_called_once_with("aggregator", json.dumps(data))
+
+    def test_select_aggregator_self_area_uuid_none_throws_exception(self, redis_client_auto_register):
```
hannesdiedrich, line 193: the function name and the docstring do not fit to what is done in there.

hannesdiedrich: Here you are testing the case when you havent registered the device to the aggregator. This is shown by `self.area_uuid=None` but please add this at least to the docstring. Thanks!

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,217 @@
...
+            self.redis_client.area_uuid = str(uuid.uuid4())
+            redis_client_auto_register\
+                .select_aggregator(aggregator_uuid=aggregator_uuid, is_blocking=True)
+
+    def test_unselect_aggregator_throws_exception(self, redis_client_auto_register):
+        """checks if unselect aggregator throws exception"""
+        aggregator_uuid = str(uuid.uuid4())
+        with pytest.raises(NotImplementedError,
+                           match="unselect_aggregator hasn\'t been implemented yet."):
+            redis_client_auto_register.unselect_aggregator(aggregator_uuid=aggregator_uuid)
```
hannesdiedrich, line 217: IMO you can skip this. I is not used and will eventually be implemented.

karthikyadav3: Sure

unit_tests/test_redis_client_base.py
karthikyadav3: Sure

unit_tests/test_redis_client_base.py
fievelk: Also, since we're talking about this:
- please use imperative mood in the docstrings (as described in [PEP 257](https://www.python.org/dev/peps/pep-0257))
  > The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".
- start docstrings with a capital letter
- end docstrings with a dot.

hannesdiedrich (commented):
Left some comments.
Also two general style comments here:
1. Docstrings are in written in imperativ, so instead of `"""checks the buffer message matching the command and id and returns as None"""` it should be `"""Check the buffer message matching the command and id and returns as None."""` (Please also see the capital start and the point at the end (https://www.python.org/dev/peps/pep-0257/#one-line-docstrings)
2. Please omit backslashes line continuation for long lines and use brackets instead https://www.python.org/dev/peps/pep-0008/#maximum-line-length

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,217 @@
...
+        assert redis_client_auto_register \
+                   ._check_buffer_message_matching_command_and_id(message["data"]) is None
+
+    def test_check_buffer_message_matching_command_and_id_throws_exception(self, redis_client_auto_register):
+        """checks buffer message matching command and id does
+           contain transaction id throws exception"""
+        message = {}
+        with pytest.raises(RedisAPIException,
+                           match="The answer message does not contain a valid "
+                                 "\'transaction_id\' member."):
```
fievelk, line 71: Nit: no need to escape the single quotes.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,217 @@
...
+    def test_check_buffer_message_matching_command_and_id_throws_another_exception(self, redis_client_auto_register):
+        """checks buffer message matching command and id throws exception"""
+        message = {"data": {"transaction_id": TRANSACTION_ID}}
+        with pytest.raises(RedisAPIException,
+                           match="There is no matching command response"
+                                 " in _blocking_command_responses."):
+            redis_client_auto_register \
+                ._check_buffer_message_matching_command_and_id(message["data"])
+
+    def test_register(self, redis_client_auto_register):
```
fievelk, line 83: This test is failing at the moment.

karthikyadav3: Only this one or the entire test is failing?

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,217 @@
...
+                                 "the unregistration has been completed."):
+            self.redis_client = redis_client_auto_register
+            self.redis_client.is_active = True
+            self.redis_client.unregister(is_blocking=True)
+
+    def test_on_register(self, redis_client_auto_register):
+        """assigning the correct arguments to blocking command response and
+        checking by calling on_register function with correct message"""
+        message = {"data": {"device_uuid": TRANSACTION_ID, "transaction_id": TRANSACTION_ID}}
+        message["data"] = json.dumps(message["data"])
```
fievelk, line 144: Nit:
```suggestion
        data = {"device_uuid": TRANSACTION_ID, "transaction_id": TRANSACTION_ID}
        message = {"data": json.dumps(data)}
```

fievelk:
General comments:

- Please make sure to address all comments before asking for a new review round on your PR, as this would speed up the review process. I can see that some comments have not been answered (and the issues they point at have not been tackled). One example is this: https://github.com/gridsingularity/d3a-api-client/pull/163/files#r692165785
- Some tests are currently failing (e.g. `test_register` and `test_unregister`). Please make sure to fix them, and feel free to ask if you need any specific help.

- [Nit]: pylint complains about the fact that you declared many methods that don't use the `self` object. I personally hate pylint warnings, so I usually fix these by declaring the test methods as `@staticmethod` and removing the `self` parameter. This is not mandatory, but it is what I would personally do.
- [Nit]: we use a maximum line length of 99 characters. Pylint correctly complains about it, so please make sure to split long lines into smaller ones. You can do that by using hanging indents: see [PEP 8 - Indentation](https://www.python.org/dev/peps/pep-0008/#indentation)
- [Nit]: you don't need to escape single quotes when they are enclosed by double quotes. E.g.: `"this doesn't create any issue"`.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,223 @@
...
+        """Checks whether the transaction id buffer is popping out and returning the empty list."""
+        data = {"transaction_id": TRANSACTION_ID}
+        message = {"data": json.dumps(data)}
+        redis_client_auto_register._transaction_id_buffer.append(TRANSACTION_ID)
+        redis_client_auto_register._aggregator_response_callback(message)
+        assert redis_client_auto_register._transaction_id_buffer == []
+
+    def test_check_buffer_message_matching_command_and_id(self, redis_client_auto_register):
+        """Checks the buffer message matching the command and id and returns as None."""
+        message = {"data": {"transaction_id": TRANSACTION_ID}}
```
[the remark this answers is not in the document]
karthikyadav3: Again I am passing the message argument with data as key to check whether the function returns as None.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,225 @@
...
+    def test_on_register(self, redis_client_auto_register):
+        """Assigning the correct arguments to blocking command response and
+           checking by calling on_register function with correct message
+           and should not throw an exception."""
+        data = {"device_uuid": DEVICE_ID, "transaction_id": TRANSACTION_ID}
+        message = {"data": json.dumps(data)}
+        redis_client_auto_register._blocking_command_responses = {
+            "register": {"transaction_id": TRANSACTION_ID}}
+        redis_client_auto_register._on_register(message)
+
```
hannesdiedrich, line 165: Please also check if is_active is also True ans well as the area_uuid was correctly set
```suggestion
assert redis_client_auto_register.active == True
assert redis_client_auto_register.area_uuid == DEVICE_ID
```

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,225 @@
...
+        """Assigning the correct arguments to blocking command response and
+           checking by calling on_unregister function with correct message
+           and should not throw an exception."""
+        data = {"device_uuid": DEVICE_ID, "transaction_id": TRANSACTION_ID,
+                "response": "success"}
+        message = {"data": json.dumps(data)}
+        redis_client_auto_register._blocking_command_responses = {
+            "unregister": {"transaction_id": TRANSACTION_ID}}
+        redis_client_auto_register._on_unregister(message)
+
```
hannesdiedrich, line 176:
```suggestion
assert redis_client_auto_register.active == False
```

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,225 @@
...
+        with pytest.raises(RedisAPIException,
+                           match=f"Failed to unregister from market {AREA_ID}."
+                                 "Deactivating connection."):
+            data = {"device_uuid": DEVICE_ID, "transaction_id": TRANSACTION_ID,
+                    "response": "unsuccessful"}
+            message = {"data": json.dumps(data)}
+            redis_client_auto_register._blocking_command_responses = {
+                "unregister": {"transaction_id": TRANSACTION_ID}}
+            redis_client_auto_register._on_unregister(message)
+
```
hannesdiedrich, line 188: Here also self.active should be checked.
```suggestion
assert redis_client_auto_register.active == True
```
However, I just saw that there is a bug in the code. `is_active` should only be changed if the command was successful IMO. What do you think?
Screenshot 2021-08-31 at 09 47 54

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,225 @@
...
+           and should not throw an exception."""
+        data = {"device_uuid": DEVICE_ID, "transaction_id": TRANSACTION_ID,
+                "response": "success"}
+        message = {"data": json.dumps(data)}
+        redis_client_auto_register._blocking_command_responses = {
+            "unregister": {"transaction_id": TRANSACTION_ID}}
+        redis_client_auto_register._on_unregister(message)
+
+    def test_on_unregister_throws_exception(self, redis_client_auto_register):
+        """Checks if response is not success should throw exception """
```
hannesdiedrich, line 178:
```suggestion
        """Check if exception is raised when response of on_unregister is not successful."""
```

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,225 @@
...
+
+    @patch("d3a_api_client.redis_client_base.wait_until_timeout_blocking",
+           side_effect=AssertionError)
+    def test_select_aggregator_with_area_uuid_throws_exception(self,
+                                                               mock_wait_until_timeout_blocking,
+                                                               redis_client_auto_register):
+        """Checks if registered user has a area uuid throws exception."""
+        aggregator_uuid = str(uuid.uuid4())
+        with pytest.raises(RedisAPIException,
+                           match="API has timed out."):
```
hannesdiedrich, line 221: This exception is only raised because there is no d3a running.
So I would rename the test to `select_aggregator_throws_exception_if_no_d3a_is_running`
And the docstring accordingly.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,225 @@
...
+        data = {"transaction_id": TRANSACTION_ID}
+        redis_client_auto_register._blocking_command_responses = (
+            {"register": {"transaction_id": TRANSACTION_ID}})
+        assert redis_client_auto_register._check_buffer_message_matching_command_and_id(
+            data) is None
+
+    def test_check_buffer_message_matching_command_and_id_throws_exception(self,
+                                                                           redis_client_auto_register):
+        """Checks buffer message matching command and id does
+           contain transaction id throws exception."""
```
hannesdiedrich, line 68:
```suggestion
    def test_check_buffer_message_matching_command_and_id_throws_exception(self,
                                                                           redis_client_auto_register):
        """Checks if _check_buffer_message_matching_command_and_id throws exception if the message does not
           contain transaction.""
```

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,225 @@
...
+
+    def test_select_aggregator_area_uuid_none_throws_exception(self,
+                                                               redis_client_auto_register):
+        """Checks when the device is not registered to the aggregator throws exception."""
+        aggregator_uuid = str(uuid.uuid4())
+        with pytest.raises(RedisAPIException,
+                           match="The device/market has not ben registered yet,"
+                                 " can not select an aggregator"):
+            redis_client_auto_register.select_aggregator(
+                aggregator_uuid=aggregator_uuid, is_blocking=True)
```
hannesdiedrich, line 211: Nit: Setting `s_blocking=True` explicitly is not needed here, right? S

hannesdiedrich (commented):
Left some minor comments, LGTM after addressed.

unit_tests/test_redis_client_base.py
fievelk: In addition to Hannes' comment, `Checks` should be `Check` (docstrings should use the imperative form).

unit_tests/test_redis_client_base.py
fievelk: Agreed with Hannes. I think this is not a `Nit`, actually: by removing `is_blocking=True` we will also test that the default value of this parameter doesn't change, breaking stuff around the codebase.

fievelk:
Not all comments from the previous reviews have been addressed. As I already wrote, all comments should be either answer or addressed with code changes before asking for a new PR review.

Some examples:
- pylint warnings
- https://github.com/gridsingularity/d3a-api-client/pull/163#discussion_r692165785
- https://github.com/gridsingularity/d3a-api-client/pull/163#discussion_r695732995
- some docstrings are unnecessarily split into shorter pieces. The available length of the lines is 99 characters in our codebase.

It looks like the PR is almost ready, but I'll review it again once the comments above are addressed.

unit_tests/test_redis_client_base.py
karthikyadav3: Sure, I will also check these things.

unit_tests/test_redis_client_base.py
karthikyadav3: Sure, I will adapt to these changes.

unit_tests/test_redis_client_base.py
karthikyadav3: Yes, you are correct.

unit_tests/test_redis_client_base.py
karthikyadav3: Sure, I will adapt to these changes. I will modify to other docstrings as well. Thank you.

unit_tests/test_redis_client_base.py
karthikyadav3: Sure, I will adapt to these changes.

unit_tests/test_redis_client_base.py
karthikyadav3: Sure, I will adapt to these changes.

unit_tests/test_redis_client_base.py
karthikyadav3: Yes, you are correct we don't need it explicitly. But I just used it as how we pass arguments in the main code. But I will remove it.

unit_tests/test_redis_client_base.py
karthikyadav3: Yes, you are right. We can directly use the fixture. I will adapt the suggested changes by using fixtures.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,235 @@
...
+
+import uuid
+import json
+from unittest.mock import MagicMock, patch
+import pytest
+from redis import StrictRedis
+import d3a_api_client
+from d3a_api_client.redis_client_base import RedisClientBase, RedisAPIException
+
+d3a_api_client.redis_client_base.StrictRedis = MagicMock(spec=StrictRedis)
```
fievelk, line 24: I think this was already pointed out in a previous comment. This is not the proper way to mock classes: please use `@patch` within the tests that use `StrictRedis` if you want to create a mock for it.
Mocking it at the module level will affect its functionality for all subsequent tests, and this is not what we want.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,235 @@
...
+        redis_client_auto_register._blocking_command_responses = (
+            {"register": {"transaction_id": TRANSACTION_ID}})
+        assert redis_client_auto_register._check_buffer_message_matching_command_and_id(
+            data) is None
+
+    @staticmethod
+    def test_check_buffer_message_matching_command_and_id_throws_exception(
+            redis_client_auto_register):
+        """Checks if _check_buffer_message_matching_command_and_id throws
+           exception if the message does not contain transaction_id."""
```
fievelk, line 73: Nit:
```suggestion
        """
        Checks if _check_buffer_message_matching_command_and_id throws exception if the message
        does not contain transaction_id.
        """
```

fievelk: I think you don't necessarily need to state the full name of the method in the docstring (if you find a better way to rephrase it). This might help with keeping the docstring shorter.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,235 @@
...
+        transaction_id = redis_client_auto_register._blocking_command_responses[
+            "register"]["transaction_id"]
+        data = {"name": AREA_ID, "transaction_id": transaction_id}
+        redis_client_auto_register.redis_db.publish.assert_called_once_with(
+            f"{AREA_ID}/register_participant", json.dumps(data))
+
+    @staticmethod
+    @patch("d3a_api_client.redis_client_base.wait_until_timeout_blocking",
+           side_effect=AssertionError)
+    def test_register_self_active_false_throws_exception(mock_wait_until_timeout_blocking,
```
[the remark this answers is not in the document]
fievelk: Also: what are you testing here?
You're not setting `active = False` anywhere in the method, so this test is not actually testing what the method name is stating.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,235 @@
...
+        with pytest.raises(RedisAPIException,
+                           match="API registration process timed out. "
+                                 "Server will continue processing "
+                                 "your request on the background "
+                                 "and will notify you as soon as the "
+                                 "registration has been completed."):
+            redis_client_auto_register.register(is_blocking=True)
+
+    @staticmethod
+    def test_register_self_active_true_throw_exception(redis_client_auto_register):
```
fievelk, line 116: Nit: for this test to be complete we could also check that:
- `self.redis_db.publish` is not called;
- `wait_until_timeout_blocking` is not called.

unit_tests/test_redis_client_base.py
fievelk: If you want, you can actually mock it inside the `redis_client_auto_register` fixture, so that it will be applied to all tests.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,235 @@
...
+        redis_client_auto_register.unregister(is_blocking=False)
+        transaction_id = redis_client_auto_register._blocking_command_responses[
+            "unregister"]["transaction_id"]
+        data = {"name": AREA_ID, "transaction_id": transaction_id}
+        redis_client_auto_register.redis_db.publish.assert_called_once_with(
+            f"{AREA_ID}/unregister_participant",
+            json.dumps(data))
+
+    @staticmethod
+    def test_unregister_is_active_false_throws_exception(redis_client_auto_register):
```
fievelk, line 137: Nit: it would be better to refactor test names to have more consistency:
- `test_register_self_active_false_throws_exception`
- `test_register_self_active_true_throw_exception`
- `test_unregister_is_active_false_throws_exception`
- `test_unregister_is_active_true_throws_exception`

As you see, the naming is different. This is a small nit anyway.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,235 @@
...
+    @staticmethod
+    def test_select_aggregator_area_uuid_none_throws_exception(
+            redis_client_auto_register):
+        """Check if the device/market is not registered to the aggregator throws an exception."""
+        aggregator_uuid = str(uuid.uuid4())
+        with pytest.raises(RedisAPIException,
+                           match="The device/market has not ben registered yet,"
+                                 " can not select an aggregator"):
+            redis_client_auto_register.select_aggregator(
+                aggregator_uuid=aggregator_uuid)
```
fievelk, line 221: Nit:
```suggestion
            redis_client_auto_register.select_aggregator(aggregator_uuid=aggregator_uuid)
```

fievelk:
General comments:
- what are the `unit_tests/test.py` and `unit_tests/test_redis_aggregator.py` files for? They're both empty.
- imports should be optimized either using PyCharm optimization feature or following PEP-8 standards
- There are still pylint warnings.

unit_tests/test_redis_client_base.py
karthikyadav3: Sure, I will adapt to these changes. I will mock it inside the fixture.

unit_tests/test_redis_client_base.py
karthikyadav3: Sure, I will change it to one line.

unit_tests/test_redis_client_base.py
karthikyadav3: Sure, Thank you

unit_tests/test_redis_client_base.py
karthikyadav3: I will change these.

unit_tests/test_redis_client_base.py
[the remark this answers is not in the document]
karthikyadav3: Sure, I will call this mock method.

unit_tests/test_redis_client_base.py
karthikyadav3: Sure, I will rephrase the docstring to 99 characters rather than double lines.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,235 @@
...
+    """Tests for the redis client base behaviour without connecting to d3a."""
+
+    @staticmethod
+    @pytest.fixture()
+    def redis_client_auto_register():
+        return RedisClientBase(area_id=AREA_ID, autoregister=False)
+
+    @staticmethod
+    @patch("d3a_api_client.redis_client_base.StrictRedis.pubsub.psubscribe")
+    def test_subscribe_to_response_channels(pubsub_mock, redis_client_auto_register):
```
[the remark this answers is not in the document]
karthikyadav3: I think I don't have to patch it. Because I already mocked the strict redis in the fixture.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,235 @@
...
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""
+
+import uuid
+import json
+from unittest.mock import MagicMock, patch
+import pytest
+from redis import StrictRedis
+import d3a_api_client
+from d3a_api_client.redis_client_base import RedisClientBase, RedisAPIException
```
fievelk, line 22: Imports need to be optimized.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,237 @@
...
+
+# pylint: disable=protected-access
+class TestRedisClientBase:
+    """Tests for the redis client base behaviour without connecting to d3a."""
+
+    @staticmethod
+    @pytest.fixture()
+    def redis_client_auto_register():
+        """Create the fixture for redis client base and mocks strict redis."""
+        d3a_api_client.redis_client_base.StrictRedis = MagicMock(spec=StrictRedis)
```
fievelk, line 39: Please be aware that this is not the proper way to mock module-level objects (`StrictRedis` in this case). As you've done elsewhere in your code, the correct way to do this is to use `patch()`.

In our case there's one more issue: we don't want to just directly `RedisClientBase().redis_db`, because we need to change its value during the instance initialization (not afterwards). That's because `RedisClientBase().pubsub` depends on it during the init stage.

So my suggestion is to patch the `StrictRedis.from_url` method (in the fixture) and return a mock from it. If you do so, you'll be able to delete all these lines from the other methods:
```python
redis_client_auto_register.redis_db.publish = MagicMock()
```

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,237 @@
...
+        with pytest.raises(RedisAPIException,
+                           match=f"Failed to unregister from market {AREA_ID}."
+                                 "Deactivating connection."):
+            data = {"device_uuid": DEVICE_ID, "transaction_id": TRANSACTION_ID,
+                    "response": "unsuccessful"}
+            message = {"data": json.dumps(data)}
+            redis_client_auto_register._blocking_command_responses = {
+                "unregister": {"transaction_id": TRANSACTION_ID}}
+            redis_client_auto_register._on_unregister(message)
+            assert redis_client_auto_register.is_active is True
```
fievelk, line 197: This line is not doing anything. As we've discussed previously, the test will exit its context (`with pytest.raises(...)` as soon as an exception is raised. Therefore, the execution will never reach your final assertion.
Moreover: if you fix it (as we've done with the other tests) you'll actually see that this assertion is wrong and it fails.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,237 @@
...
+            "register": {"transaction_id": TRANSACTION_ID}}
+        redis_client_auto_register._on_register(message)
+        assert redis_client_auto_register.is_active is True
+        assert redis_client_auto_register.area_uuid == DEVICE_ID
+
+    @staticmethod
+    def test_on_unregister(redis_client_auto_register):
+        """Check the on_unregister function with correct message that doesn't throws exception."""
+        data = {"device_uuid": DEVICE_ID, "transaction_id": TRANSACTION_ID,
+                "response": "success"}
```
fievelk, line 178: Nit: no need to break it into two lines
```suggestion
        data = {"device_uuid": DEVICE_ID, "transaction_id": TRANSACTION_ID, "response": "success"}
```

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,237 @@
...
+                                 " in _blocking_command_responses."):
+            redis_client_auto_register._check_buffer_message_matching_command_and_id(data)
+
+    @staticmethod
+    def test_register(redis_client_auto_register):
+        """Check if the redis.db.publish is called with correct arguments."""
+        redis_client_auto_register.redis_db.publish = MagicMock()
+        redis_client_auto_register.register(is_blocking=False)
+        transaction_id = redis_client_auto_register._blocking_command_responses[
+            "register"]["transaction_id"]
```
fievelk, line 97: Since `_blocking_command_responses` is a private property, we should not use it in our tests.
You're using it because you want to know the value of `transaction_id` (since you need it for the assertion later on). 
What you can do is treating `uuid.uuid4()` as what it really is: an outgoing query to an external function. You can just patch it in this test and use any string (e.g. `"some-transaction-uuid"`) as its return value.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,237 @@
...
+            redis_client_auto_register.register(is_blocking=False)
+
+    @staticmethod
+    def test_unregister(redis_client_auto_register):
+        """Check whether the redis.db.publish is called with correct arguments."""
+        redis_client_auto_register.redis_db.publish = MagicMock()
+        redis_client_auto_register.is_active = True
+        redis_client_auto_register.unregister(is_blocking=False)
+        transaction_id = redis_client_auto_register._blocking_command_responses[
+            "unregister"]["transaction_id"]
```
fievelk, line 132: Same as https://github.com/gridsingularity/d3a-api-client/pull/163/files#r706238079

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,237 @@
...
+            redis_client_auto_register):
+        """Check exception is raised if the message does not contain transaction_id."""
+        message = {}
+        with pytest.raises(RedisAPIException,
+                           match="The answer message does not contain a valid "
+                                 "'transaction_id' member."):
+            redis_client_auto_register._check_buffer_message_matching_command_and_id(message)
+
+    @staticmethod
+    def test_check_buffer_message_matching_command_and_id_throws_another_exception(
```
fievelk, line 82: Nit:
```suggestion
    def test_check_buffer_message_matching_command_and_id_without_command_throws_exception(
```

unit_tests/test_redis_client_base.py
karthikyadav3: I think you are correct but I tried the fixture with a patch. Every unit test function needs the Redis connection in this case. Due to this, I have to use the return value of the fixture in all the functions. If I can mock the Strict Redis in the redis_auto_register then I don't have to mock any Redis-related functions again. For this case we can just MagicMock StrictRedis once in the fixture and avoid other magic mocks for redis.

unit_tests/test_redis_client_base.py
karthikyadav3: Yes, you are correct. I will remove the last line.

unit_tests/test_redis_client_base.py
karthikyadav3: Sure, I will adapt these changes.

unit_tests/test_redis_client_base.py
karthikyadav3: Sure, I will rename the function.

unit_tests/test_redis_client_base.py
fievelk: That's exactly what I was suggesting ;)

karthikyadav3: I tried to mock and return the value of from_url but the behavior hasn't changed anything even just mocking only StrictRedis. So I just mocked the StrictRedis in the fixture and removed other lines of mocking redis_db for other functions. Now iit is working fine.

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,235 @@
...
+        """Check if is active set to true throws exception."""
+        with pytest.raises(RedisAPIException,
+                           match="API unregister process timed out. "
+                                 "Server will continue processing "
+                                 "your request on the background and "
+                                 "will notify you as soon as "
+                                 "the unregistration has been completed."):
+            redis_client_auto_register.is_active = True
+            redis_client_auto_register.unregister(is_blocking=True)
+        mock_wait_until_timeout_blocking.assert_called()
```
Hassan754, line 158: Nit, having said the exception was raised, this is indeed called so this line can be removed

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,235 @@
...
+
+    @staticmethod
+    @patch("uuid.uuid4", return_value="some-transaction-uuid")
+    def test_on_register(mock_uuid, redis_client_auto_register):
```
Hassan754, line 162: Nit, In case you're testing the private _on_register method, there are three extra lines that can be tested

1. The mocked call to `_check_buffer_message_matching_command_and_id` method
2. the side effect (logging)
3. the call to the on_register method

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,235 @@
...
+        redis_client_auto_register._on_register(message)
+        assert redis_client_auto_register.is_active is True
+        assert redis_client_auto_register.area_uuid == DEVICE_ID
+
+    @staticmethod
+    @patch("uuid.uuid4", return_value="some-transaction-uuid")
+    def test_on_unregister(mock_uuid, redis_client_auto_register):
+        """Check the on_unregister function with correct message that doesn't throws exception."""
+        redis_client_auto_register.is_active = True
+        redis_client_auto_register.unregister(is_blocking=False)
```
Hassan754, line 176: I would advise not calling the unregister method here, you can mock the `_check_buffer_message_matching_command_and_id` and directly call the _on_unregister  method (Unless we specifically agreed to not test private methods)

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,235 @@
...
+
+    @staticmethod
+    @patch("uuid.uuid4", return_value="some-transaction-uuid")
+    def test_on_unregister_throws_exception(mock_uuid, redis_client_auto_register):
+        """Check if exception is raised when response of on_unregister is not successful."""
+        with pytest.raises(RedisAPIException,
+                           match=f"Failed to unregister from market {AREA_ID}."
+                                 "Deactivating connection."):
+            redis_client_auto_register.is_active = True
+            redis_client_auto_register.unregister(is_blocking=False)
```
Hassan754, line 191: Same as above

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,235 @@
...
+            data = {"name": AREA_ID, "device_id": DEVICE_ID,
+                    "transaction_id": mock_uuid(), "response": "unsuccessful"}
+            message = {"data": json.dumps(data)}
+            redis_client_auto_register._on_unregister(message)
+
+    @staticmethod
+    def test_select_aggregator(redis_client_auto_register):
+        """Check if redis_db.publish is called with correct arguments to select aggregator."""
+        aggregator_uuid = str(uuid.uuid4())
+        redis_client_auto_register.redis_db.publish = MagicMock()
```
Hassan754, line 201: Nit, what is the scope of these mocks? are they reset after the test run ?

Hassan754 (dismissed):
Left some comments
LGTM and thanks a lot!

unit_tests/test_redis_client_base.py
fievelk: That's a good question. As far as I see from the docs, the `patch` decorators/context managers are
> used for patching objects only within the scope of the function they decorate. They automatically handle the unpatching for you, even if exceptions are raised.

In this case, we're manually adding the mock, so I assume that it will stay there for the entire life of the `redis_client_auto_register` fixture object. However, the default scope of a fixture in pytest is just `"function"`, and it will be destroyed once the function returns: https://docs.pytest.org/en/latest/how-to/fixtures.html#fixture-scopes

So it shouldn't make a difference in this case.

Hassan754: Good to know, thanks!

unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,235 @@
...
+from d3a_api_client.redis_client_base import RedisClientBase, RedisAPIException
+
+AREA_ID = str(uuid.uuid4())
+TRANSACTION_ID = str(uuid.uuid4())
+DEVICE_ID = str(uuid.uuid4())
+
+
+# pylint: disable=protected-access
+class TestRedisClientBase:
+    """Tests for the redis client base behaviour without connecting to d3a."""
```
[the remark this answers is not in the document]
fievelk: This is actually not a verb, but rather a noun. Since it's a class docstring, it shouldn't be necessary to use imperative verbs as far as I saw in the [PEP 257](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings).

unit_tests/test_redis_client_base.py
fievelk: It's true in the current case. However, the exception could be raised by any other call/line in the `unregister()` method, without `mock_wait_until_timeout_blocking` to be called. I think the current version makes it a bit clearer where exactly the exception was raised.
I agree that it's not strictly necessary to be so specific here, but I'd keep it here as it doesn't harm. Would this be ok for you?

unit_tests/test_redis_client_base.py
fievelk: I've answered above in this comment: https://github.com/gridsingularity/d3a-api-client/pull/163#discussion_r721442048

unit_tests/test_redis_client_base.py
fievelk: As you said, the reason why we decided not to call the `_check_buffer_message_matching_command_and_id` method is that it is private. Ideally, we shouldn't know anything about it (from the perspective of the test). What we know, instead, is the public interface of the class and that's why we opted to use it here to "set up" the conditions for the test.

unit_tests/test_redis_client_base.py
Hassan754: Sure, it's okay
It's just the the Redis exception will be thrown upon catching an AssertionError, which is the side effect specified in the mock_wait_until_timeout_blocking , this is why I thought it's somehow clear but indeed no harm in keeping it

unit_tests/test_redis_client_base.py
fievelk: 1. I would actually not assert the mocked call, since this is a private method. I would rather create a new test to check whether the exception is raised inside of it. However, that method has already been unit-tested, so if this is not compelling for you I might let it go. If you'd rather have it, I can definitely work on it.
2. Do you think we should test logging calls? We kept it out of scope by choice, but I can add it.
3. Done!

Thank you!

unit_tests/test_redis_client_base.py
hannesdiedrich: @fievelk can you please add this change to the code `RedisClientBase._on_unregister` ? IMO this is a bug. Thanks!

hannesdiedrich (dismissed):
One comment for a change in the client itself. LGTM when resolved

hannesdiedrich (approved):
LGTM

PR #164, opened 2021-08-19
D3ASIM-3549: Added set energy forecast and measurements to the batch …

spyrostz:
…commands. Fixed integration tests by pointing to the branch with the d3a fix of bid_energy command.

~~Integration tests are still pending in order for the d3a implementation to be completed. ~~

Integration tests for the aggregated commands set_energy_forecast and set_energy_measurements are part of this PR.

### DIFF
--- a/d3a_api_client/commands.py
+++ b/d3a_api_client/commands.py
@@ -80,6 +80,12 @@ class ClientCommandBuffer:
     def last_market_dso_stats(self, area_uuid):
         return self._add_to_buffer(area_uuid, Commands.DSO_MARKET_STATS, {"data": {}})
 
+    def set_energy_forecast(self, area_uuid, energy_forecast_kWh):
+        return self._add_to_buffer(area_uuid, Commands.FORECAST, {"energy_forecast": energy_forecast_kWh})
+
+    def set_energy_measurement(self, area_uuid, energy_measurement_kWh):
+        return self._add_to_buffer(area_uuid, Commands.MEASUREMENT, {"energy_measurement": energy_measurement_kWh})
+
     def change_grid_fees_percent(self, area_uuid, fee_percent):
         return self._add_to_buffer(
             area_uuid,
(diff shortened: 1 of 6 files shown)


d3a_api_client/commands.py
```diff
@@ -80,6 +80,12 @@ def device_info(self, area_uuid):
     def last_market_dso_stats(self, area_uuid):
         return self._add_to_buffer(area_uuid, Commands.DSO_MARKET_STATS, {"data": {}})
 
+    def set_energy_forecast(self, area_uuid, energy_forecast_kWh):
+        return self._add_to_buffer(area_uuid, Commands.FORECAST, {"energy_forecast": energy_forecast_kWh})
+
+    def set_energy_measurement(self, area_uuid, energy_measurement_kWh):
+        return self._add_to_buffer(area_uuid, Commands.MEASUREMENT, {"energy_measurement": energy_measurement_kWh})
```
faizan2590, line 87: Shouldn't this directly be reported by device rather than aggregator. I mean an aggregator would off-course be having some AI algo to predict the forecasted energy via `set_energy_forecast ` BUT for truthfulness of actual  `energy_measurement ` should directly be extracted from device(via some device-measurement subscription protocol) rather than aggregator reporting it

faizan2590 (commented):
Left more of a design comment, other than that LGTM.

hannesdiedrich (commented):
Will approve once https://github.com/gridsingularity/d3a/pull/1207#pullrequestreview-734750090 is addressed

d3a_api_client/commands.py
spyrostz: @faizan2590 valid point, but keep in mind that the d3a-api-client is a general library that can be used for various purposes. One of these is a utility/aggregator that is able to control and read data from multiple assets at the same time. In this case, it is the responsibility of the aggregator to send the device measurements and forecasts to us. In such a scenario, it is far more optimal to send one batch command for all the device measurements rather than limit them by sending individual commands for each individual device. Keep in mind also that this is not a theoretical scenario, we already have customers that require this.

hannesdiedrich (commented):
LGTM

PR #166, opened 2021-09-02
Tests 98

karthikyadav3:
TETS-98: Unit tests for redis_aggregator


unit_tests/test_redis_client_base.py
```diff
@@ -0,0 +1,235 @@
+"""
```
hannesdiedrich, line 1: Why is this part of this PR? Isnt' it targeted in https://github.com/gridsingularity/d3a-api-client/pull/163 ?
I would suggest to re-open a  new PR that only contains the new changes. Thanks!

PR #167, opened 2021-09-02
D3ASIM-3567: Added timeslot as an extra command argument for aggregat…

spyrostz:
…or commands.

### DIFF
--- a/d3a_api_client/commands.py
+++ b/d3a_api_client/commands.py
@@ -2,6 +2,7 @@ import logging
 from typing import Dict, List
 
 from tabulate import tabulate
+
 from d3a_api_client.enums import Commands, command_enum_to_command_name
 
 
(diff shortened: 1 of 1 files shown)


d3a_api_client/commands.py
```diff
@@ -16,71 +17,71 @@ def buffer_length(self):
...
+        return self._add_to_buffer(area_uuid, Commands.LIST_BIDS, {"timeslot": timeslot})
 
     def device_info(self, area_uuid):
         return self._add_to_buffer(area_uuid, Commands.DEVICE_INFO, {})
 
     def last_market_dso_stats(self, area_uuid):
         return self._add_to_buffer(area_uuid, Commands.DSO_MARKET_STATS, {"data": {}})
 
-    def set_energy_forecast(self, area_uuid, energy_forecast_kWh: Dict):
+    def set_energy_forecast(self, area_uuid, energy_forecast_kWh: Dict  ):
```
fievelk, line 84: Nit:
```suggestion
    def set_energy_forecast(self, area_uuid, energy_forecast_kWh: Dict):
```

fievelk (commented):
LGTM in general, with some questions: 
1. what's the format of the time slot that needs to be passed? This doesn't seem to be currently documented here and I think it would be good to add it to each docstring;
2. would it be doable to pass a datetime object instead of a string? (Probably not, but I'm asking to be sure)

spyrostz:
> LGTM in general, with some questions:
> 
> 1. what's the format of the time slot that needs to be passed? This doesn't seem to be currently documented here and I think it would be good to add it to each docstring;
> 2. would it be doable to pass a datetime object instead of a string? (Probably not, but I'm asking to be sure)

1. Thanks for pointing this out, the accepted format is one of the formats that function `str_to_pendulum_datetime` from d3a-interface supports. I was under the impression that we had changed this method to support multiple formats, however only two are supported. I will open a separate smaller PR to add all the formats that we use, that you can see in constants_limits.py
2. It could be possible, but you have to extend the default JSON serializer / deserializer. datetimes cannot be serialized and deserialized by the default json package and you have to provide custom strategies for what to do when encountering a datetime object, and (the difficult part) how to convert  a string that looks like a datetime string to a datetime object. Can be done, but will probably be inefficient and ugly.

fievelk (approved):
> > LGTM in general, with some questions:
> > 
> > 1. what's the format of the time slot that needs to be passed? This doesn't seem to be currently documented here and I think it would be good to add it to each docstring;
> > 2. would it be doable to pass a datetime object instead of a string? (Probably not, but I'm asking to be sure)
> 
> 1. Thanks for pointing this out, the accepted format is one of the formats that function `str_to_pendulum_datetime` from d3a-interface supports. I was under the impression that we had changed this method to support multiple formats, however only two are supported. I will open a separate smaller PR to add all the formats that we use, that you can see in constants_limits.py
> 2. It could be possible, but you have to extend the default JSON serializer / deserializer. datetimes cannot be serialized and deserialized by the default json package and you have to provide custom strategies for what to do when encountering a datetime object, and (the difficult part) how to convert  a string that looks like a datetime string to a datetime object. Can be done, but will probably be inefficient and ugly.

Thanks! I think the formats should be documented in the API docs in any case (if they're not yet), since some users might want to implement their own client (without necessarily knowing about `constants_limits.py`).
LGTM

faizan2590 (dismissed):

hannesdiedrich (dismissed):
LGTM

PR #169, opened 2021-09-03
Feature/d3 asim 3554: Script for placing bid/offer with attributes/requirements

faizan2590:


fievelk:
General comments:
- Since setups are supposed to be sort of examples for users/devs, they should probably have more documentation. I would highly suggest to add more comments throughout their code to explain the "what" and "why" of the code you wrote. For example: what is `trade_violation_count` checking?

Nits:
- please fix all pylint warnings (logging, docstrings, quotation marks, etc);
- please favour using the appropriate parameters for `logging` calls. E.g.:
  ```python
  logging.info("current_market_maker_rate: %s", market_info["market_maker_rate"])
  ```
   This will avoid the warnings that pylint is issuing (and it will improve the code).

d3a_api_client/setups/test_stocks_attribute_requirement.py
```diff
@@ -0,0 +1,52 @@
...
+from time import sleep
+
+from d3a_interface.utils import key_in_dict_and_not_none_and_greater_than_zero
+
+from d3a_api_client.redis_aggregator import RedisAggregator
+from d3a_api_client.redis_device import RedisDeviceClient
+from d3a_api_client.redis_market import RedisMarketClient
+
+
+class AutoAggregator(RedisAggregator):
```
fievelk, line 11: I think it would be better to change the name of this class. That's because we already have other classes with the same name, and since you're importing this in the `myco-api-client`, it will create confusion.

fievelk:
One more general comment: I think the name of the modules in the `setups/` folder should be changed: their prefix should be `setup_*` and not `test_*` (which is confusing since it belongs to "real" tests).

spyrostz (approved):
LGTM once comments from @Hassan754 and @fievelk are dealt with

PR #171, opened 2021-09-10
D3ASIM-3572: show exceptions raised in threads

fievelk:
We want to use this utility function both in the d3a-api-client and in the myco-api-client. Therefore, I'm moving it on d3a-interface.

This PR also allows to specify a different target branch for local tox tests.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,7 +5,7 @@ import os
 from setuptools import find_packages, setup
 from d3a_api_client import __version__
 
-TARGET_BRANCH = os.environ.get("TARGET_BRANCH", "master")
+BRANCH = os.environ.get("BRANCH", "master")
 
 
 try:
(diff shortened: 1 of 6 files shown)


Hassan754 (dismissed):
LGTM

setup.py
```diff
@@ -1,22 +1,25 @@
+"""Setup module for the d3a-api-client."""
+
 import os
-from setuptools import setup, find_packages
+
+from setuptools import find_packages, setup
 from d3a_api_client import __version__
 
-target_branch = os.environ.get('BRANCH', 'master')
+TARGET_BRANCH = os.environ.get("TARGET_BRANCH", "master")
```
spyrostz, line 8: Same as the myco PR, this will probably break the CI scripts.

spyrostz (approved):
One comment, LGTM

Hassan754 (approved):

PR #177, opened 2021-10-12
 Fix _check_transaction_id_cached_out check

Hassan754:
The `_check_transaction_id_cached_out` method was not working properly, it was supposed to return `True` when the transaction_id is removed from the buffer, but it was returning the opposite expected value
There were several function calls, 2 of them were prefixed by a `not` statement so they were passing, but a third usage in `execute_batch_commands` was failing because of the unexpected return value

### DIFF
--- a/unit_tests/test_redis_client_base.py
+++ b/unit_tests/test_redis_client_base.py
@@ -240,3 +240,14 @@ class TestRedisClientBase:
             redis_client_auto_register.area_uuid = str(uuid.uuid4())
             redis_client_auto_register.select_aggregator(aggregator_uuid=aggregator_uuid)
         mock_wait_until_timeout_blocking.assert_called()
+
+    def test_is_transaction_response_received(self, redis_client_auto_register):
+        """Check the return value of is_transaction_response_received."""
+        transaction_id = str(uuid.uuid4())
+        redis_client_auto_register._transaction_id_buffer = [transaction_id]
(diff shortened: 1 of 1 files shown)


d3a_api_client/redis_aggregator.py
```diff
@@ -112,10 +115,10 @@ def _events_callback_dict(self, message):
         self._on_event_or_response(payload)
 
     def _check_transaction_id_cached_out(self, transaction_id):
-        return transaction_id in self._transaction_id_buffer
+        return transaction_id not in self._transaction_id_buffer
```
hannesdiedrich, line 118: for the others: actual fix 1/3

d3a_api_client/redis_aggregator.py
```diff
@@ -28,7 +28,7 @@ class RedisAPIException(Exception):
 class RedisAggregator:
 
     def __init__(self, aggregator_name, accept_all_devices=True,
-                 redis_url='redis://localhost:6379'):
+                 redis_url="redis://localhost:6379"):
```
spyrostz, line 31: Outside of the scope of the PR, but since it won't take long, could you please take the time and refactor this to a global constant named LOCAL_REDIS_URL or similar in order to be reusable ? Thanks!

hannesdiedrich (approved):
LGTM

PR #180, opened 2021-11-01
D3ASIM-3649: Add Pylint pre-commit hook

fievelk:
- Add pre-commit hook for Pylint, trailing whitespace and EOF fixer.

PR #183, opened 2021-11-02
D3ASIM-3651: Add flake8 as pre-commit hook

faizan2590:


spyrostz (approved):
I am not sure if this was discussed, but why not copying the whole content of the d3a precommit hook (as far as I see, in addition to flake, we also have multiple pre-commit hooks in d3a)? Is it because we will add pylint anyways, so it is an overkill to also check it via pre-commit ? 

LGTM once this is discussed

fievelk:
> I am not sure if this was discussed, but why not copying the whole content of the d3a precommit hook (as far as I see, in addition to flake, we also have multiple pre-commit hooks in d3a)? Is it because we will add pylint anyways, so it is an overkill to also check it via pre-commit?

@spyrostz I'm adding pylint and two other pre-commit hooks in a different PR (https://github.com/gridsingularity/d3a-api-client/pull/180). Should I add other hooks?

spyrostz:
> > I am not sure if this was discussed, but why not copying the whole content of the d3a precommit hook (as far as I see, in addition to flake, we also have multiple pre-commit hooks in d3a)? Is it because we will add pylint anyways, so it is an overkill to also check it via pre-commit?
> 
> @spyrostz I'm adding pylint and two other pre-commit hooks in a different PR (#180). Should I add other hooks?

IMO it would be great if all hooks were synced, you can see the ones we currently use in d3a here https://github.com/gridsingularity/d3a/blob/master/.pre-commit-config.yaml

fievelk:
> > > I am not sure if this was discussed, but why not copying the whole content of the d3a precommit hook (as far as I see, in addition to flake, we also have multiple pre-commit hooks in d3a)? Is it because we will add pylint anyways, so it is an overkill to also check it via pre-commit?
> > 
> > 
> > @spyrostz I'm adding pylint and two other pre-commit hooks in a different PR (#180). Should I add other hooks?
> 
> IMO it would be great if all hooks were synced across repos (not automatically, manually copying the hooks would do), therefore it would be great if you ported all of them from d3a to the other repos. You can see the ones we currently use in d3a here https://github.com/gridsingularity/d3a/blob/master/.pre-commit-config.yaml

Ok, I'll update the PRs 👍

fievelk (approved):

PR #187, opened 2021-11-12
 Add d3a-api-client cli alias for backward compatibility

Hassan754:
+ Add flake8 configurations in the `tox.ini` module to extend the max-line-length config to 99

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -25,6 +25,10 @@ commands =
 	coverage combine
 	coverage report
 
+[flake8]
+max-line-length = 99
+exclude = .tox,.cache,.pytest_cache
+
 [testenv:integrationtests]
 basepython = python3.8
 deps =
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -25,6 +25,10 @@ commands =
 	coverage combine
 	coverage report
 
+[flake8]
+max-line-length = 99
+exclude = .tox,.cache,.pytest_cache
```
fievelk, line 30: Can you explain why is this needed? Flake is not run in the tox environments, right?

Hassan754: Yep you are right in the fact that we do not run flake8 in the tox tests.
However, this was introduced because the pre-commit hook failed (a line was longer than 77 characters), so I had to add global flake8 configurations (placing them in tox.ini works)

fievelk: Interesting, thanks. I think that adding `--max-line-length 99` to `pre-commit-config.yaml` would have worked as well, but with your solution this should also work from cli if you want to check flake8 manually).

fievelk (approved):

spyrostz (approved):
LGTM

PR #188, opened 2021-11-12
Fix the name of the folder created for tests

fievelk:
Integration tests were trying to delete a folder that was not being created anymore (because of the rebranding).

### DIFF
--- a/integration_tests/build_test_containers.sh
+++ b/integration_tests/build_test_containers.sh
@@ -5,8 +5,8 @@ set -e
 GSY_E_IMAGE_TAG="gsy-e-tests"
 
 if [[ "$(docker images -q ${GSY_E_IMAGE_TAG} 2> /dev/null)" == "" ]]; then
-  echo "Building d3a image ..." && \
-  rm -rf tests/d3a && \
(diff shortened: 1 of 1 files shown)


Hassan754 (approved):
LGTM, so another issue was that the created folders are not being deleted (I wonder how this is handled in the github actions, I assume they'll be deleted anyway)

fievelk:
> LGTM, so another issue was that the created folders are not being deleted (I wonder how this is handled in the github actions, I assume they'll be deleted anyway)

Exactly, that was the main reason why I created this PR. Maybe the github actions wipe their env completely at every run.

spyrostz:
> > LGTM, so another issue was that the created folders are not being deleted (I wonder how this is handled in the github actions, I assume they'll be deleted anyway)
> 
> Exactly, that was the main reason why I created this PR. Maybe the github actions wipe their env completely at every run.

Github actions are dockerized, they create their own container for running the CI job, which gets deleted as soon as the test run completes. Therefore this should not affect Github actions, very handy fix for the local run though!

spyrostz (approved):
LGTM

PR #189, opened 2021-11-15
D3ASIM-3553: Rename device to asset

fievelk:
- Rename `Rest/RedisDeviceClient` to `Rest/RedisAssetClient`
  - Keep classes with old naming for backwards-compatibility
  - Refactor setups to reflect the new naming. Only setups that are used in documentation or tests are updated
- Rename `area_uuid` parameter to `asset_uuid` for methods that are specialized on assets (this change is not backwards-compatible)
- Rename `device_info()` method to `asset_info()` (and its parameter)
  - keep deprecated method for backwards-compatibility


gsy_e_sdk/rest_device.py
```diff
@@ -1,147 +1,17 @@
...
-    logging_decorator, get_aggregator_prefix,
-    domain_name_from_env, websocket_domain_name_from_env,
-    simulation_id_from_env, get_configuration_prefix, log_trade_info)
-from gsy_e_sdk.websocket_device import DeviceWebsocketMessageReceiver
+logger.warning(
+    "The RestDeviceClient class name will be deprecated. Please use RedisAssetClient instead.")
 
-REGISTER_COMMAND_TIMEOUT = 15 * 60
 
+class RestDeviceClient(RestAssetClient):
```
spyrostz, line 13: Minor, however for future purposes, keep in mind that with this approach (moving the content of the module / class to a new one, and marking the existing module / class as deprecated) the git history will be lost (because this code was moved, I assume that we cannot backtrack to past changes in the code). Not really important and I do not think that it is required to change it though.

spyrostz (dismissed):
LGTM

gsy_e_sdk/rest_device.py
fievelk: How would you suggest doing it? We could in theory do the opposite (keep the content in `RestDeviceClient` and use `RestAssetclient(RestDeviceClient)`. But then we would be making future changes to a class that is deprecated, and it's a bit counter-intuitive.

gsy_e_sdk/clients/rest_asset_client.py
```diff
@@ -0,0 +1,171 @@
...
+
+REGISTER_COMMAND_TIMEOUT = 15 * 60
+
+#pylint: disable=too-many-instance-attributes
+class RestAssetClient(APIClientInterface, RestCommunicationMixin):
+    """Client class for assets to be used while working with REST."""
+
+    # pylint: disable=super-init-not-called
+    # pylint: disable=too-many-arguments
+    def __init__(self, area_id, simulation_id=None, domain_name=None, websockets_domain_name=None,
```
hannesdiedrich, line 27: I think it is a good time time rename this variable too:
```suggestion
    def __init__(self, asset_uuid, simulation_id=None, domain_name=None, websockets_domain_name=None,
```

hannesdiedrich (commented):
One possible enhancement, LGTM after addressed.

gsy_e_sdk/rest_device.py
spyrostz: The correct way IMO would be:
1. rename the module (rest_device.py -> rest_asset.py)
2. rename RestDeviceClient->RestAssetClient
3. create a new RestDeviceClient class with a warning log 
4. commit your changes in order for git to track the renames in its history
5. create a new module called rest_device.py , and move the class RestDeviceClient there 
6. Commit

fievelk: But in your steps the `rest_asset.py` stays in the same location of the old module. What I wanted to do was to create a `clients/` package and move the new module inside. I suppose that this would still break the history.

gsy_e_sdk/clients/rest_asset_client.py
fievelk: This would break backwards compatibility for users that used the old parameter, like:
```python
RestDeviceClient(area_id=123, ...)
```
One way to overcome this would be to explicitly init the new `RestAssetClient` in the old `RestDeviceClient` init method. For example:
```python
class RestDeviceClient(RestAssetClient):
    def __init__(
            self, self, area_id, simulation_id=None, domain_name=None, websockets_domain_name=None,
            autoregister=False, start_websocket=True, sim_api_domain_name=None):
        super().__init__(
            area_id=asset_uuid, simulation_id=simulation_id, domain_name=domain_name, 
            websockets_domain_name=websockets_domain_name,
            autoregister=False, start_websocket=True, sim_api_domain_name=sim_api_domain_name)
```
This would make the inheritance process more complex to maintain, because we would need to make sure that every time the parameters are changed in `RestAssetClient` we also change them in the old `RestDeviceClient`.

If you want I can do that, I just want to make sure that this is what we want.

hannesdiedrich: Ah yes, backwards compatibility, I forgot. Having in mind, that we deprecate the old clients at some point anyway I think it is better to present a consistent interface to the user and bite the bullet to have more maintenance to do if we change parameters here (which I do not predict).
TLDR; IMO your proposed change LGTM with one little typo found and corrected here:
```python
class RestDeviceClient(RestAssetClient):
    def __init__(
            self, self, area_id, simulation_id=None, domain_name=None, websockets_domain_name=None,
            autoregister=False, start_websocket=True, sim_api_domain_name=None):
        super().__init__(
            asset_uuid=area_id, simulation_id=simulation_id, domain_name=domain_name, 
            websockets_domain_name=websockets_domain_name,
            autoregister=False, start_websocket=True, sim_api_domain_name=sim_api_domain_name)
```

fievelk: Thanks, I've added the changes in e33a06a 👍 

BTW: the same could be applied to the `RedisDeviceClient` class. However, that class inherits from `RedisClientBase` which is not deprecated. So changing the parameters there would introduce some incoherent changes in my opinion (or we would need to change `RedisClientBase` to support this in a backward-compatible manner). I left it out for the moment, let me know if I should work on it.

gsy_e_sdk/rest_device.py
fievelk: I'll definitely try it out next time! Thanks

Hassan754 (dismissed):
LGTM

hannesdiedrich (approved):
Please let Chris now about these changes so he can either dispatch this to other researchers or at least know when they have problems. Thanks a lot!

Hassan754 (dismissed):

hannesdiedrich (approved):
LGTM

spyrostz (approved):
LGTM

spyrostz (approved):
LGTM

gsy_e_sdk/clients/rest_asset_client.py
```diff
@@ -0,0 +1,172 @@
...
+#pylint: disable=too-many-instance-attributes
+class RestAssetClient(APIClientInterface, RestCommunicationMixin):
+    """Client class for assets to be used while working with REST."""
+
+    # pylint: disable=super-init-not-called
+    # pylint: disable=too-many-arguments
+    def __init__(
+            self, asset_uuid, simulation_id=None, domain_name=None, websockets_domain_name=None,
+             autoregister=False, start_websocket=True, sim_api_domain_name=None):
+        self.simulation_id = simulation_id if simulation_id else simulation_id_from_env()
```
Hassan754, line 30: Nit
```suggestion
        self.simulation_id = simulation_id or simulation_id_from_env()
```

gsy_e_sdk/types.py
```diff
@@ -1,24 +1,20 @@
...
 from gsy_e_sdk.aggregator import Aggregator
 from gsy_e_sdk.redis_aggregator import RedisAggregator
 from gsy_e_sdk.rest_market import RestMarketClient
 from gsy_e_sdk.redis_market import RedisMarketClient
 
 
 def _select_client_type(rest_type, redis_type):
-    if "API_CLIENT_RUN_ON_REDIS" in os.environ and \
-            os.environ["API_CLIENT_RUN_ON_REDIS"] == "true":
+    if "API_CLIENT_RUN_ON_REDIS" in os.environ and os.environ["API_CLIENT_RUN_ON_REDIS"] == "true":
```
Hassan754, line 11: Nit
```suggestion
    if os.environ.get("API_CLIENT_RUN_ON_REDIS") == "true":
```

gsy_e_sdk/redis_device.py
```diff
@@ -1,21 +1,17 @@
-from gsy_e_sdk.constants import LOCAL_REDIS_URL
-from gsy_e_sdk.redis_client_base import RedisClientBase
+import logging
+from gsy_e_sdk.clients.redis_asset_client import RedisAssetClient
 
+logger = logging.getLogger(__name__)
+console_handler = logging.StreamHandler()
+console_handler.setFormatter(logging.Formatter("%(levelname)s: %(message)s"))
+logger.addHandler(console_handler)
```
Hassan754, line 7: shouldn't the cli module already handle these settings?

Hassan754 (approved):
Left one question and some non blocking comments, LGTM

gsy_e_sdk/redis_device.py
fievelk: This handler works when the cli is not used. But you're right in the sense that if the cli is used two warnings are shown (one for each handler). If you want, I can remove this in a new PR.

PR #195, opened 2021-12-08
TESTS-106-suggestions: Remove unneeded mocks and patches

Hassan754:


unit_tests/test_aggregator.py
```diff
@@ -14,11 +14,7 @@
 
 @pytest.fixture(autouse=True)
 def mock_connections(mocker):
-    """Fixture mocks methods and functions which establish external connections"""
-    mocker.patch("gsy_e_sdk.aggregator.blocking_get_request",
-                 return_value=test_agg, autospec=True)
-    mocker.patch("gsy_e_sdk.aggregator.blocking_post_request",
-                 return_value=test_agg, autospec=True)
+    """Mock methods and functions which establish external connections."""
```
Hassan754, line 17: I removed the above patches because these are utils that can be used across the while module, each time with different return values, and I can clearly see that you're indeed patching these below

Hassan754: + Regarding the docstring, we have a special [guideline](https://gridsingularity.atlassian.net/wiki/spaces/D3A/pages/2406449155/Naming+and+Style+conventions#Docstrings) on how to write these, for example we decided to always use the imperative form in the docstrings

This is relevant in the rest of PR

unit_tests/test_aggregator.py
```diff
@@ -40,10 +34,8 @@ def mock_prefixes(mocker):
 
 @pytest.fixture(autouse=True)
 def mock_start_websocket_funcs(mocker):
-    """
-    Fixture mocks external classes used in Aggregator.start_websocket_connections
-    """
-    # mocker.patch("gsy_e_sdk.aggregator.AggregatorWebsocketMessageReceiver")
+    """Mock external classes used in Aggregator.start_websocket_connections."""
+    mocker.patch("gsy_e_sdk.aggregator.AggregatorWebsocketMessageReceiver")
```
Hassan754, line 38: I patched the `AggregatorWebsocketMessageReceiver` object here and reverted the patch that was done in `mock_execute_batch_command_methods`

unit_tests/test_aggregator.py
```diff
@@ -52,31 +44,21 @@ def mock_start_websocket_funcs(mocker):
...
                  return_value=test_batch_command_dict)
     mocker.patch("gsy_e_sdk.aggregator.Aggregator._post_request",
                  return_value=["some_transaction_id", True])
-    mocker.patch("gsy_e_sdk.aggregator.AggregatorWebsocketMessageReceiver"
-                 + ".wait_for_command_response",
-                 return_value=test_response)
     mocker.patch("gsy_e_sdk.aggregator.Aggregator."
                  + "_all_uuids_in_selected_device_uuid_list",
                  return_value=True)
-    mocker.patch("gsy_e_sdk.aggregator.log_bid_offer_confirmation")
```
Hassan754, line 71: This seems to be only be used once, so I patched it in this specific use case, same for the `log_deleted_bid_offer_confirmation`

unit_tests/test_aggregator.py
```diff
@@ -99,137 +79,125 @@ def fixture_aggregator_from_env(mocker):
...
 
     @staticmethod
-    def test_start_websocket_connection__dispatcher(aggregator_explicit):
-        with patch("gsy_e_sdk.aggregator.AggregatorWebsocketMessageReceiver") \
-                as mocked_class:
-            aggregator_explicit.start_websocket_connection()
-            mocked_class.assert_called()
+    def test_start_websocket_connection_dispatcher(aggregator_explicit):
+        aggregator_explicit.start_websocket_connection()
+        assert aggregator_explicit.dispatcher is not None
```
Hassan754, line 194: Checking whether it's not None is sufficient

unit_tests/test_aggregator.py
```diff
@@ -99,137 +79,125 @@ def fixture_aggregator_from_env(mocker):
...
 
     @staticmethod
-    def test_start_websocket_connection__websocket_creation(aggregator_explicit):
-        websocket_uri_test = f"{test_simulation['websockets_domain_name']}" \
-                             + f"/{test_simulation['uuid']}/aggregator/" \
-                             + f"{test_agg['uuid']}/"
+    def test_start_websocket_connection_websocket_creation(aggregator_explicit):
+        websocket_uri_test = (f"{test_simulation['websockets_domain_name']}"
+                              + f"/{test_simulation['uuid']}/aggregator/"
+                              + f"{test_agg['uuid']}/")
```
Hassan754, line 200: We tend to use parenthesis instead of backslashes for line escaping, except for the assertions

unit_tests/test_aggregator.py
```diff
@@ -294,84 +262,78 @@ def test_calculate_grid_fee__return(aggregator_explicit):
...
     @staticmethod
-    def test_execute_batch_commands__execute_batch_called(aggregator_explicit):
-        with patch("gsy_e_sdk.aggregator.ClientCommandBuffer.execute_batch") as mocked_method:
-            aggregator_explicit.start_websocket_connection()
-            aggregator_explicit.execute_batch_commands()
-            mocked_method.assert_called_once()
+    def test_execute_batch_commands_execute_batch_called(aggregator_explicit):
+        aggregator_explicit.start_websocket_connection()
+        aggregator_explicit.execute_batch_commands()
+        aggregator_explicit._client_command_buffer.execute_batch.assert_called_once()
```
Hassan754, line 277: We can directly access the `_client_command_buffer` interface even if it's private, it probably is more straightforward than patching the object

unit_tests/test_aggregator.py
```diff
@@ -294,84 +262,78 @@ def test_calculate_grid_fee__return(aggregator_explicit):
...
-            aggregator_explicit.execute_batch_commands()
-            mocked_method.assert_called()
+    def test_execute_batch_commands_validate_all_uuids_in_list(aggregator_explicit):
+        aggregator_explicit.start_websocket_connection()
+        aggregator_explicit.execute_batch_commands()
+        aggregator_explicit._all_uuids_in_selected_device_uuid_list.assert_called()
 
     @staticmethod
-    def test_execute_batch_commands__post_request_called_with_args(aggregator_explicit):
-        args_ = (f"{'/aggregator_prefix/'}batch-commands",
```
Hassan754, line 322: Why do we need to append the _ at the end of the variable name?

unit_tests/test_aggregator.py
```diff
@@ -294,84 +262,78 @@ def test_calculate_grid_fee__return(aggregator_explicit):
...
-                 {"aggregator_uuid": test_agg["uuid"], "batch_commands": test_batch_command_dict}
-                 )
+    def test_execute_batch_commands_post_request_called_with_args(aggregator_explicit):
+        args = (f"{'/aggregator_prefix/'}batch-commands",
+                {"aggregator_uuid": test_agg["uuid"], "batch_commands": test_batch_command_dict}
+                )
 
-        with patch("gsy_e_sdk.aggregator.Aggregator._post_request",
```
Hassan754, line 326: We're already patching the `_post_request` at the top of the module, so we can alter the return value here instead of repatching it

unit_tests/test_aggregator.py
```diff
@@ -404,12 +366,9 @@ def test_delete_aggregator(aggregator_explicit):
 
 
 def test_add_to_batch_commands(aggregator_explicit):
-    # pylint: disable=protected-access
     buffer_instance = aggregator_explicit.add_to_batch_commands
     assert buffer_instance == aggregator_explicit._client_command_buffer
 
 
 def test_commands_buffer_length(aggregator_explicit):
-    patch("gsy_e_sdk.aggregator.ClientCommandBuffer.buffer_length",
```
Hassan754, line 413: The exact same patch is performed at the top of the module

Hassan754 (commented):
Left some comments, In general It would be great if you can take this PR 1 extra step by removing mocks that are only needed in 1 test (I probably covered these but I could have missed some)

unit_tests/test_aggregator.py
ERafaelMartinez: Regarding your first comment, I patched it for all tests because the constructor implements `_connect_to_simulation() `method, which itself executes `_create_aggregator()` and `list_aggregators()` methods. In these methods, the functions being patched you mention are called and therefore, each time I instance the `Aggregator` class, these previously patched functions would be called and therefore try to establish an external connection. 

As you noted I override them in some tests below, where I wanted to test explicitly if the particular function was called. 

I'm not sure though what you mean by the _while module_.

ERafaelMartinez: Yesterday I also came up with the idea to patch private methods before patching the external functions being used inside them. However I'm not sure If this one is a good practice.

unit_tests/test_aggregator.py
ERafaelMartinez: You mean that you reverted it because it would then be redundant to have the class and a method of it patched as well, right?

unit_tests/test_aggregator.py
ERafaelMartinez: Is the first modification after the function-docstring the one you meant earlier about reverting the patch in this fixture?

unit_tests/test_aggregator.py
```diff
@@ -52,31 +44,21 @@ def mock_start_websocket_funcs(mocker):
...
+    """Mock methods and external functions and other methods used in execute_batch_command."""
     mocker.patch("gsy_e_sdk.aggregator.ClientCommandBuffer.buffer_length",
                  new_callable=PropertyMock, return_value=3)
     mocker.patch("gsy_e_sdk.aggregator.ClientCommandBuffer.execute_batch",
                  return_value=test_batch_command_dict)
     mocker.patch("gsy_e_sdk.aggregator.Aggregator._post_request",
                  return_value=["some_transaction_id", True])
-    mocker.patch("gsy_e_sdk.aggregator.AggregatorWebsocketMessageReceiver"
-                 + ".wait_for_command_response",
-                 return_value=test_response)
```
ERafaelMartinez, line 67: Is the first modification after the function-docstring the one you meant earlier about reverting the patch in this fixture?

unit_tests/test_aggregator.py
Hassan754: I see, in my opinion you should create 1 method that tests the `_connect_to_simulation` (you can patch the mentioned functions here), and in the rest of the tests you can mock this entire `_connect_to_simulation` function (at this stage we already know that it's working properly)

unit_tests/test_aggregator.py
ERafaelMartinez: This is a similar case as the one in the first comment. I decided to patch this way, because in other tests cases it might be used when calling `execute_batch_commands()`. When I want to test explicitly if these functions were called is when I invoke the context manager.

unit_tests/test_aggregator.py
Hassan754: Exactly

unit_tests/test_aggregator.py
Hassan754: @ERafaelMartinez yes

unit_tests/test_aggregator.py
ERafaelMartinez: I understand that the `auto_use=True` clause takes this fixture to every test case. Originally I wanted to avoid this by calling it only on the test_execute_batch_commands__*() test cases, however, as these fixtures do not return anything I had to deal with the warning about not using a passed variable. 

In order to avoid passing the fixture to the test case of interest (and therefore avoiding the warning) I set `auto_use=True`.

I would really appreciate any suggestions regarding this I mentioned.

unit_tests/test_aggregator.py
```diff
@@ -99,137 +79,125 @@ def fixture_aggregator_from_env(mocker):
...
-#####################################
-# DEF TESTS FOR CLASSES CONSTRUCTOR #
-
 class TestConstructor:
-    """Includes methods which test constructor."""
+    """Test the constructor use cases of the aggregator."""
 
     @staticmethod
-    def test_constructor__simulation_id_explicit(aggregator_explicit):
+    def test_constructor_simulation_id_explicit(aggregator_explicit):
```
ERafaelMartinez, line 86: I'd like to inquire here about the main reason why you suggest not using this double underscore.

IMO it helps to identify in an easier manner which use case is being tested. I'm not sure however if i'm violating any PEP rule or smthn.

unit_tests/test_aggregator.py
Hassan754: You can try the `@pytest.mark.usefixtures('my_fixture')` decorator, and in any case, and please always five priority to functional code quality over pylint's warnings, pylint mainly pushes us towards better code but sometimes it could be non supporting to other libraries (this could be clear in tests)

unit_tests/test_aggregator.py
Hassan754: @ERafaelMartinez We don't have a written rule on this in our guidelines but all of our test methods so far use a single underscore between words, if we're to change this it would be awesome if you can raise this topic in the backend channel so others can give their opinion

Hassan754: Actually it seems to be already written https://gridsingularity.atlassian.net/wiki/spaces/D3A/pages/2406449155/Naming+and+Style+conventions?focusedCommentId=3147104267#Naming-conventions

unit_tests/test_aggregator.py
ERafaelMartinez: This would work because the `execute_batch` method has already been patched in `mock_execute_batch_command_methods` fixture, right?

Hassan754: 100% no need to patch it again here

unit_tests/test_aggregator.py
ERafaelMartinez: I did that because I thought it might conflict with a python reserved word. I understand from your suggestion it does not :P

unit_tests/test_aggregator.py
```diff
@@ -404,12 +366,9 @@ def test_delete_aggregator(aggregator_explicit):
 
 
 def test_add_to_batch_commands(aggregator_explicit):
-    # pylint: disable=protected-access
```
ERafaelMartinez, line 407: Is it proper to move this to the top of he document even if it is only used on this test case?

Hassan754: Actually I moved it to the top because we're now accessing several private members throughout the module _client_command_buffer, _all_uuids_in_selected_device_uuid_list, _post_request

unit_tests/test_aggregator.py
ERafaelMartinez: Similar situation as in this comment [here](https://github.com/gridsingularity/gsy-e-sdk/pull/195/files/697cab28a6b68be7ace854535033cb16c47a4be6#r764905505).

unit_tests/test_aggregator.py
ERafaelMartinez: I'll have a look to this you mention.

unit_tests/test_aggregator.py
Hassan754: Well this is debatable, sometimes even private methods deserve their own testing functions.
Anyway, it's up to you, In case you wanted to refrain from testing this private method you can completely ignore the _connect_to_simulation and add an assertion in the aggregator fixture to check whether `aggregator_uuid` is defined after the initialization is complete, still you have to patch the method here

ERafaelMartinez: I ended up patching these functions without an explicit return value just to avoid the execution.  Then when it was needed, I added it explicitly later on the module.

PR #197, opened 2021-12-17
TESTS-99

ERafaelMartinez:
Added unit tests for `RedisAggregator` class.

### DIFF
--- a/unit_tests/test_redis_aggregator.py
+++ b/unit_tests/test_redis_aggregator.py
@@ -1,4 +1,4 @@
-# pylint: disable=missing-function-docstring, protected-access, no-member
+# pylint: disable=missing-function-docstring, protected-access, no-member, too-many-public-methods
 import json
 import uuid
 from unittest.mock import patch, PropertyMock, MagicMock, call
@@ -67,8 +67,8 @@ def fixture_aggregator(mocker):
     return RedisAggregator(aggregator_name=TEST_AGGREGATOR_NAME)
 
 
-class TestRedisAggregatorConstructor:
-    """Include tests for the constructor of the RedisAggregator class."""
+class TestRedisAggregator:
+    """Test RedisAggregator public methods."""
 
     @staticmethod
     @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
@@ -165,12 +165,8 @@ class TestRedisAggregatorConstructor:
             mocked_class.assert_called()
             assert agg.lock is lock_mock
 
-
-@pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
-class TestRedisAggregatorDeleteAggregator:
-    """Test cases for RedisAggregator's delete_aggregator method."""
-
     @staticmethod
+    @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
     def test_delete_aggregator_redis_db_publishes_data(aggregator):
         data = {"name": TEST_AGGREGATOR_NAME,
                 "aggregator_uuid": TEST_AGGREGATOR_UUID,
@@ -184,6 +180,7 @@ class TestRedisAggregatorDeleteAggregator:
             "aggregator", json.dumps(data))
 
     @staticmethod
+    @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
     def test_delete_aggregator_api_time_out_raise_exception(aggregator):
         with patch("gsy_e_sdk.redis_aggregator.wait_until_timeout_blocking",
                    side_effect=AssertionError):
@@ -192,6 +189,7 @@ class TestRedisAggregatorDeleteAggregator:
                 aggregator.delete_aggregator()
 
     @staticmethod
+    @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
     @pytest.mark.parametrize("test_is_blocking, expected_ret_val",
                              [(True, TEST_TRANSACTION_ID), (False, None)])
     def test_delete_aggregator_returns_expected(aggregator, test_is_blocking,
@@ -199,13 +197,9 @@ class TestRedisAggregatorDeleteAggregator:
         aggregator.aggregator_uuid = TEST_AGGREGATOR_UUID
         assert aggregator.delete_aggregator(is_blocking=test_is_blocking) is expected_ret_val
 
-
-@pytest.mark.usefixtures("mock_client_command_buffer_attributes")
-class TestRedisAggregatorExecuteBatchCommands:
-    """Test cases for RedisAggregator's execute_batch_commands method."""
-
     @staticmethod
-    @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
+    @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking",
+                             "mock_client_command_buffer_attributes")
     def test_execute_batch_commands_commands_buffer_length_zero_return_none():
         with patch("gsy_e_sdk.redis_aggregator.RedisAggregator.commands_buffer_length",
                    new_callable=PropertyMock,
@@ -214,6 +208,7 @@ class TestRedisAggregatorExecuteBatchCommands:
             assert agg.execute_batch_commands() is None
 
     @staticmethod
+    @pytest.mark.usefixtures("mock_client_command_buffer_attributes")
     def test_execute_batch_commands_not_all_uuids_in_selected_device_uuid_list_raise_exception(
             aggregator):
         aggregator.device_uuid_list = [TEST_DEVICE_UUID_1]
@@ -222,7 +217,8 @@ class TestRedisAggregatorExecuteBatchCommands:
             aggregator.execute_batch_commands(is_blocking=False)
 
     @staticmethod
-    @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
+    @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking",
+                             "mock_client_command_buffer_attributes")
     def test_execute_batch_commands_batched_command_published(aggregator):
         aggregator.device_uuid_list = [TEST_DEVICE_UUID_1, TEST_DEVICE_UUID_2]
         aggregator.aggregator_uuid = TEST_AGGREGATOR_UUID
@@ -237,6 +233,7 @@ class TestRedisAggregatorExecuteBatchCommands:
                                                        json.dumps(batched_command_input))
 
     @staticmethod
+    @pytest.mark.usefixtures("mock_client_command_buffer_attributes")
     def test_execute_batch_commands_api_time_out_raise_exception(aggregator):
         with patch("gsy_e_sdk.redis_aggregator.wait_until_timeout_blocking",
                    side_effect=AssertionError):
@@ -246,7 +243,8 @@ class TestRedisAggregatorExecuteBatchCommands:
                 aggregator.execute_batch_commands()
 
     @staticmethod
-    @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
+    @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking",
+                             "mock_client_command_buffer_attributes")
     @pytest.mark.parametrize("is_blocking, trans_id_resp_buffer, expected_ret_val",
                              [(True, {TEST_TRANSACTION_ID: TEST_RESPONSE}, TEST_RESPONSE),
                               (False, {TEST_TRANSACTION_ID: TEST_RESPONSE}, None),
@@ -259,12 +257,8 @@ class TestRedisAggregatorExecuteBatchCommands:
 
         assert aggregator.execute_batch_commands(is_blocking=is_blocking) is expected_ret_val
 
-
-@pytest.mark.usefixtures("mock_grid_fee_calculation")
-class TestRedisAggregatorCalculateGridFee:
-    """Test cases for RedisAggregator's calculate_grid_fee method."""
-
     @staticmethod
+    @pytest.mark.usefixtures("mock_grid_fee_calculation")
     def test_calculate_grid_fee_called_with_args(aggregator):
         aggregator.calculate_grid_fee("start_market",
                                       "target_market",
@@ -275,6 +269,7 @@ class TestRedisAggregatorCalculateGridFee:
             "current_market_fee")
 
     @staticmethod
+    @pytest.mark.usefixtures("mock_grid_fee_calculation")
     def test_calculate_grid_fee_returns_expected(aggregator):
         aggregator.grid_fee_calculation.calculate_grid_fee.return_value = "TEST_OK"
         ret_val = aggregator.calculate_grid_fee("start_market",
@@ -282,26 +277,26 @@ class TestRedisAggregatorCalculateGridFee:
                                                 "current_market_fee")
         assert ret_val == "TEST_OK"
 
+    @staticmethod
+    def test_get_uuid_from_area_name_returns_expected(aggregator):
+        test_area_mapping = {"TestArea": [str(uuid.uuid4())]}
 
-def test_get_uuid_from_area_name_returns_expected(aggregator):
-    test_area_mapping = {"TestArea": [str(uuid.uuid4())]}
+        aggregator.area_name_uuid_mapping = test_area_mapping
+        area_name = "TestArea"
+        ret_val = test_area_mapping[area_name][0]
 
-    aggregator.area_name_uuid_mapping = test_area_mapping
-    area_name = "TestArea"
-    ret_val = test_area_mapping[area_name][0]
+        assert aggregator.get_uuid_from_area_name(area_name) == ret_val
 
-    assert aggregator.get_uuid_from_area_name(area_name) == ret_val
+    @staticmethod
+    @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
+    def test_add_to_batch_commands_returns_client_command_buffer():
+        client_command_buffer_mock = MagicMock()
+        with patch("gsy_e_sdk.redis_aggregator.ClientCommandBuffer",
+                   return_value=client_command_buffer_mock):
+            aggregator = RedisAggregator(aggregator_name=TEST_AGGREGATOR_NAME)
+            assert aggregator.add_to_batch_commands is client_command_buffer_mock
 
-
-@pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
-def test_add_to_batch_commands_returns_client_command_buffer():
-    client_command_buffer_mock = MagicMock()
-    with patch("gsy_e_sdk.redis_aggregator.ClientCommandBuffer",
-               return_value=client_command_buffer_mock):
-        aggregator = RedisAggregator(aggregator_name=TEST_AGGREGATOR_NAME)
-        assert aggregator.add_to_batch_commands is client_command_buffer_mock
-
-
-@pytest.mark.usefixtures("mock_client_command_buffer_attributes")
-def test_commands_buffer_length_returns_expected_buffer_length(aggregator):
-    assert aggregator.commands_buffer_length == 3
+    @staticmethod
+    @pytest.mark.usefixtures("mock_client_command_buffer_attributes")
+    def test_commands_buffer_length_returns_expected_buffer_length(aggregator):
+        assert aggregator.commands_buffer_length == 3
--- a/unit_tests/test_redis_aggregator.py
+++ b/unit_tests/test_redis_aggregator.py
@@ -31,36 +31,11 @@ TEST_RESPONSE = {"status": "ready",
                          "buyer": "test_buyer"},
                  }
 
-TEST_MESSAGE_DICT = {
-    "data": {
-        "transaction_id": TEST_TRANSACTION_ID,
-        "status": "SELECTED",
-        "event": "market",
-        "aggregator_uuid": None,
-        "responses": {"TEST_ASSET_UUID": [TEST_RESPONSE]},
-        "device_uuid": None,
-        "grid_tree": {},
-        "trade_list": [{"trade_price": 1,
-                        "traded_energy": 1,
-                        "buyer": "anonymous",
-                        "seller": "test_seller"}],
-
-        "slot_completion": "0%",
-        "content": {
-            "slot_completion": "0%",
-            "start_time": "00:00",
-            "duration_min": 300,
-            "market_slot": 1, }
-    }
-}
-
 
 @pytest.fixture(autouse=True)
 def fixture_mock_connections(mocker):
     mocker.patch("gsy_e_sdk.redis_aggregator.StrictRedis")
     mocker.patch("gsy_e_sdk.redis_aggregator.ThreadPoolExecutor")
-    mocker.patch("gsy_e_sdk.redis_aggregator.Dict",
-                 return_value=TEST_MESSAGE_DICT)
 
 
 @pytest.fixture(name="mock_grid_fee_calculation")
@@ -100,9 +75,8 @@ class TestRedisAggregatorConstructor:
     def test_constructor_grid_fee_calculation_instantiated():
         grid_fee_calculation_mock = MagicMock()
         with patch("gsy_e_sdk.redis_aggregator.GridFeeCalculation",
-                   return_value=grid_fee_calculation_mock) as mocked_class:
+                   return_value=grid_fee_calculation_mock):
             agg = RedisAggregator(aggregator_name=TEST_AGGREGATOR_NAME)
-            mocked_class.assert_called()
             assert agg.grid_fee_calculation is grid_fee_calculation_mock
 
     @staticmethod
@@ -125,25 +99,24 @@ class TestRedisAggregatorConstructor:
     def test_constructor_client_command_buffer_instantiated():
         client_command_buffer_mock = MagicMock()
         with patch("gsy_e_sdk.redis_aggregator.ClientCommandBuffer",
-                   return_value=client_command_buffer_mock) as mocked_class:
+                   return_value=client_command_buffer_mock):
             agg = RedisAggregator(aggregator_name=TEST_AGGREGATOR_NAME)
-            mocked_class.assert_called()
         assert agg._client_command_buffer is client_command_buffer_mock
 
     @staticmethod
     @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
     def test_constructor_threadpoolexecutor_instantiated():
-        ThreadPoolExecutor_mock = MagicMock
+        thread_pool_executor_mock = MagicMock
         with patch("gsy_e_sdk.redis_aggregator.ThreadPoolExecutor",
-                   return_value=ThreadPoolExecutor_mock) as mocked_class:
+                   return_value=thread_pool_executor_mock) as mocked_class:
             agg = RedisAggregator(aggregator_name=TEST_AGGREGATOR_NAME)
             mocked_class.assert_called()
-        assert agg.executor is ThreadPoolExecutor_mock
+        assert agg.executor is thread_pool_executor_mock
 
     @staticmethod
     @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
     def test_constructor_connect_and_subscribe_side_effects_1():
-        """Test  side effects due to the methods called with _connect_and_subscribe().
+        """Test side effects due to the methods called with _connect_and_subscribe().
         -> Side effects of _subscribe_to_aggregator_response_and_start_redis_thread
         """
 
@@ -156,7 +129,7 @@ class TestRedisAggregatorConstructor:
     @staticmethod
     @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
     def test_constructor_connect_and_subscribe_side_effects_2():
-        """Test  side effects due to the methods called with _connect_and_subscribe().
+        """Test side effects due to the methods called with _connect_and_subscribe().
         -> Side effects of _connect_to_simulation -> _create_aggregator private method.
         """
         aggregator = RedisAggregator(aggregator_name=TEST_AGGREGATOR_NAME)
@@ -170,7 +143,7 @@ class TestRedisAggregatorConstructor:
     @staticmethod
     @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
     def test_constructor_connect_and_subscribe_side_effects_3():
-        """Test  side effects due to the methods called with _connect_and_subscribe().
+        """Test side effects due to the methods called with _connect_and_subscribe().
         -> Side effects of _subscribe_to_response_channels private method
         """
         aggregator = RedisAggregator(aggregator_name=TEST_AGGREGATOR_NAME)
@@ -185,11 +158,12 @@ class TestRedisAggregatorConstructor:
     @staticmethod
     @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
     def test_constructor_lock_instantiated():
-        with patch("gsy_e_sdk.redis_aggregator.Lock"
-                   ) as mocked_class:
+        lock_mock = MagicMock()
+        with patch("gsy_e_sdk.redis_aggregator.Lock",
+                   return_value=lock_mock) as mocked_class:
             agg = RedisAggregator(aggregator_name=TEST_AGGREGATOR_NAME)
             mocked_class.assert_called()
-            assert agg.lock is not None
+            assert agg.lock is lock_mock
 
 
 @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
@@ -263,7 +237,6 @@ class TestRedisAggregatorExecuteBatchCommands:
                                                        json.dumps(batched_command_input))
 
     @staticmethod
-    @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
     def test_execute_batch_commands_api_time_out_raise_exception(aggregator):
         with patch("gsy_e_sdk.redis_aggregator.wait_until_timeout_blocking",
                    side_effect=AssertionError):
--- a/unit_tests/test_redis_aggregator.py
+++ b/unit_tests/test_redis_aggregator.py
@@ -1,0 +1,238 @@
+# pylint: disable=missing-function-docstring, protected-access
+import json
+import uuid
+from unittest.mock import patch, PropertyMock, MagicMock
+import pytest
+
+from gsy_e_sdk.redis_aggregator import RedisAggregator, RedisAggregatorAPIException
+
+TEST_AGGREGATOR_NAME = "TestAgg"
+
+TEST_AREA_NAME = "TestArea"
+TEST_AREA_MAPPING = {TEST_AREA_NAME: [str(uuid.uuid4())]}
+
+TEST_TRANSACTION_ID = str(uuid.uuid4())
+TEST_AGGREGATOR_UUID = str(uuid.uuid4())
+
+TEST_DEVICE_UUID_1 = str(uuid.uuid4())
+TEST_DEVICE_UUID_2 = str(uuid.uuid4())
+TEST_BATCH_COMMAND = "TEST_BATCH_COMMAND"
+
+TEST_BATCH_COMMAND_DICT = {
+    TEST_DEVICE_UUID_1: [TEST_BATCH_COMMAND],
+    TEST_DEVICE_UUID_2: [TEST_BATCH_COMMAND]
+}
+
+TEST_TRANSACTION_ID_RESPONSE = "TEST_TRANSACTION_ID_RESPONSE"
+TEST_TRANSACTION_ID_BUFFER = {
+    TEST_TRANSACTION_ID: {"transaction_id": TEST_TRANSACTION_ID,
+                          "response": TEST_TRANSACTION_ID_RESPONSE}
+}
+
+TEST_RESPONSE = "TEST_RESPONSE"
+TEST_TRANSACTION_ID_RESPONSE_BUFFER = {TEST_TRANSACTION_ID: TEST_RESPONSE}
+
+
+@pytest.fixture(autouse=True)
+def fixture_mock_connections(mocker):
+    mocker.patch("gsy_e_sdk.redis_aggregator.StrictRedis")
+    mocker.patch("gsy_e_sdk.redis_aggregator.ThreadPoolExecutor")
+
+
+@pytest.fixture(name="mock_grid_fee_calculation")
+def fixture_mock_grid_fee_calculation(mocker):
+    """Mock GridFeeCalculation class."""
+    mocker.patch("gsy_e_sdk.redis_aggregator.GridFeeCalculation")
+
+
+@pytest.fixture(name="mock_transaction_id_and_timeout_blocking")
+def fixture_mock_transaction_id_and_timeout_blocking(mocker):
+    mocker.patch("gsy_e_sdk.redis_aggregator.uuid.uuid4",
+                 return_value=TEST_TRANSACTION_ID)
+    mocker.patch("gsy_e_sdk.redis_aggregator.wait_until_timeout_blocking")
+
+
+@pytest.fixture(name="mock_client_command_buffer_attributes")
+def fixture_mock_client_command_buffer(mocker):
+    mocker.patch("gsy_e_sdk.redis_aggregator.ClientCommandBuffer.buffer_length",
+                 new_callable=PropertyMock, return_value=3)
+    mocker.patch("gsy_e_sdk.redis_aggregator.ClientCommandBuffer.execute_batch",
+                 return_value=TEST_BATCH_COMMAND_DICT)
+
+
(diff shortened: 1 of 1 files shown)


unit_tests/test_redis_aggregator.py
```diff
@@ -0,0 +1,238 @@
...
+    @staticmethod
+    def test_constructor_threadpoolexecutor_instantiated(aggregator):
+        assert aggregator.executor is not None
+
+    @staticmethod
+    def test_constructor_connect_and_subscribe_called():
+        with patch("gsy_e_sdk.redis_aggregator."
+                   "RedisAggregator._connect_and_subscribe") as mocked_method:
+            RedisAggregator(aggregator_name=TEST_AGGREGATOR_NAME)
+            mocked_method.assert_called()
```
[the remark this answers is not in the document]
Hassan754: One more thing, even though the `_subscribe_to_response_channels1 is a private method and we tend to avoid testing these methods, in this particular case I find it quite important to test it to make sure that the change in channels name is watched  (In theory I would assume that the class should keep track of its channels in a special variable that is tested)

unit_tests/test_redis_aggregator.py
```diff
@@ -0,0 +1,238 @@
...
+    def test_execute_batch_commands_not_all_uuids_in_selected_device_uuid_list_raise_exception(
+            aggregator):
+        aggregator.device_uuid_list = [TEST_DEVICE_UUID_1]
+
+        with pytest.raises(Exception):
+            aggregator.execute_batch_commands(is_blocking=False)
+
+    @staticmethod
+    def test_execute_batch_commands_batched_command_published(aggregator):
+        uuid.uuid4 = MagicMock(return_value=TEST_TRANSACTION_ID)
```
Hassan754, line 159: Can we reuse the `fixture_mock_transaction_id_and_timeout_blocking` here?

Hassan754 (commented):
Left some comments, it looks great
it would be awesome if you also start using the imperative form in your commit messages

fievelk:
Please note that this PR isn't currently working because it's not providing `pytest-mock` in the requirements.

unit_tests/test_redis_aggregator.py
```diff
@@ -0,0 +1,238 @@
...
+def fixture_mock_client_command_buffer(mocker):
+    mocker.patch("gsy_e_sdk.redis_aggregator.ClientCommandBuffer.buffer_length",
+                 new_callable=PropertyMock, return_value=3)
+    mocker.patch("gsy_e_sdk.redis_aggregator.ClientCommandBuffer.execute_batch",
+                 return_value=TEST_BATCH_COMMAND_DICT)
+
+
+@pytest.fixture(name="aggregator")
+def fixture_aggregator(mocker):
+    mocker.patch("gsy_e_sdk.redis_aggregator.RedisAggregator._create_aggregator")
```
fievelk, line 65: I have the feeling that this is not needed. It is a private method, which shouldn't be mocked in general, (that's why it caught my attention). Also, we're already mocking `StrictRedis`, so `_create_aggregator` shouldn't do much I suppose. What you probably wanted to avoid is the waiting time on `wait_until_timeout_blocking()`, but we could probably handle this issue in a different way (and without a global fixture).

Moreover: I'm having the feeling that we could benefit from using a fake redis server for our unit tests. We do it for S3 (with fakes3) so I guess there wouldn't be anything wrong.

unit_tests/test_redis_aggregator.py
```diff
@@ -0,0 +1,238 @@
...
+    """Includes tests for the constructor of the RedisAggregator class."""
+
+    @staticmethod
+    @pytest.mark.usefixtures("mock_grid_fee_calculation")
+    def test_constructor_grid_fee_calculation_instantiated(aggregator):
+        assert aggregator.grid_fee_calculation is not None
+
+    @staticmethod
+    def test_constructor_redis_db_instantiated(aggregator):
+        assert aggregator.redis_db is not None
```
fievelk, line 79: If we really want to test constructors, we should probably check the details of what's happening (not only checking for `is None`). 

Here, for example, we want to make sure that `StrictRedis.from_url` is called with the `redis_url` parameter, and that the result of this operation is stored inside of `self.redis_db`.

fievelk: The same concept applies to `test_constructor_grid_fee_calculation_instantiated` above.

unit_tests/test_redis_aggregator.py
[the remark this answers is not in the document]
fievelk: I agree with Hassan that here we're not really testing the effects of this method. `_connect_and_subscribe` is a private method, so I wouldn't mock it if that can be avoided. We can ignore also the internal calls to `_subscribe_to_aggregator_response_and_start_redis_thread()` and so on, because they are also private methods which would bind us to the implementation.

What we want to actually test are the side-effects of this method, pretending that these intermediate private method calls don't exist (as if everything was implemented inside of the method). What are these side-effects? For example:
1. `self.pubsub.psubscribe` is called with some specific arguments
2. `self.pubsub.run_in_thread` is called with some specific arguments
3. if `self.aggregator_uuid` is specified, then `self.aggregator_uuid` will be equal to some specific UUID
And so on.

Sidenote: the more nested methods we have, the more difficult it is to follow the path to their side-effects. This is usually a warning for us: probably we could refactor the class and split it into several classes, which would then have their own public (and testable) methods.

unit_tests/test_redis_aggregator.py
```diff
@@ -0,0 +1,238 @@
...
+
+
+@pytest.mark.usefixtures("mock_client_command_buffer_attributes")
+class TestRedisAggregatorExecuteBatchCommands:
+    """Test cases for RedisAggregator's execute_batch_commands method."""
+
+    @staticmethod
+    def test_execute_batch_commands_commands_buffer_length_zero_return_none(aggregator):
+        with patch("gsy_e_sdk.redis_aggregator.ClientCommandBuffer.buffer_length",
+                   new_callable=PropertyMock, return_value=0):
```
fievelk, line 146: Here I think we don't need to know about the `ClientCommandBuffer`. This is an implementation detail that we decided to "hide" behind the `self.commands_buffer_length` property. So we just need to know what `self.commands_buffer_length` returns.

fievelk: This, of course, only works if we also test the `commands_buffer_length` property (and it looks like you've done that).

unit_tests/test_redis_aggregator.py
```diff
@@ -0,0 +1,238 @@
...
+        with patch("gsy_e_sdk.redis_aggregator.wait_until_timeout_blocking",
+                   side_effect=AssertionError):
+            aggregator.device_uuid_list = [TEST_DEVICE_UUID_1, TEST_DEVICE_UUID_2]
+            aggregator.aggregator_uuid = TEST_AGGREGATOR_UUID
+            with pytest.raises(RedisAggregatorAPIException):
+                aggregator.execute_batch_commands()
+
+    @staticmethod
+    @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
+    @pytest.mark.parametrize("is_blocking_, trans_id_resp_buffer, expected_ret_val",
```
fievelk, line 185: Nit: why `is_blocking_` and not `is_blocking`? PEP 8 mentions using trailing underscores only in this case:
> single_trailing_underscore_: used by convention to avoid conflicts with Python keyword

I understand that you might have done it to avoid having the same parameter name, but this is not a convention we're currently using as a team and it clashes with the PEP 8 guidelines (as it would make me think that `is_blocking` is a reserved python keyword, while it isn't).
I would personally suggest to just use `is_blocking`.

unit_tests/test_redis_aggregator.py
```diff
@@ -0,0 +1,238 @@
...
+
+
+def test_get_uuid_from_area_name_returns_expected(aggregator):
+    aggregator.area_name_uuid_mapping = TEST_AREA_MAPPING
+    area_name = TEST_AREA_NAME
+    ret_val = TEST_AREA_MAPPING[TEST_AREA_NAME][0]
+
+    patch("gsy_e_sdk.redis_aggregator.get_uuid_from_area_name_in_tree_dict",
+          return_value=ret_val)
+    assert aggregator.get_uuid_from_area_name(area_name) == ret_val
```
fievelk, line 228: Here we're dealing with an outgoing query message. Since the execution of this method doesn't go very far (it's just a simple function in utils.py) I would probably avoid stubbing the return value: this is something you usually do to avoid long computations.
We could just parameterize the possible parameters we pass to `get_uuid_from_area_name()` and check the side-effects.

unit_tests/test_redis_aggregator.py
```diff
@@ -0,0 +1,238 @@
...
+
+
+def test_add_to_batch_commands_returns_client_command_buffer(aggregator):
+    buffer_instance = aggregator.add_to_batch_commands
+    assert buffer_instance == aggregator._client_command_buffer
```
fievelk, line 233: Once again, my attention was caught by the private `_client_command_buffer`. Since it's private, I think that we probably shouldn't know about it. Here we want to know that `self.add_to_batch_commands` will return the same instance of `ClientCommand` that was instantiated in the `__init__()` method.

unit_tests/test_redis_aggregator.py
```diff
@@ -0,0 +1,238 @@
...
+
+
+@pytest.fixture(name="aggregator")
+def fixture_aggregator(mocker):
+    mocker.patch("gsy_e_sdk.redis_aggregator.RedisAggregator._create_aggregator")
+    return RedisAggregator(aggregator_name=TEST_AGGREGATOR_NAME)
+
+
+class TestRedisAggregatorConstructor:
+    """Includes tests for the constructor of the RedisAggregator class."""
```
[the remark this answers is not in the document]
ERafaelMartinez: Are you suggesting here that it would be ok if I grouped several assertions in a single test for the constructor of the class? Or how this should be in order to not override the "one assertion per test" rule.
Or could you please clarify how you would do it?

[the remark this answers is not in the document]
Hassan754: Well I would personally aggregate these assertions in as single test case, I'm not a fan of the one assertion per test methodology as long as all the assertions are related/similar/simple and relevant in the context of the test case performed

unit_tests/test_redis_aggregator.py
```diff
@@ -0,0 +1,238 @@
+# pylint: disable=missing-function-docstring, protected-access
+import json
+import uuid
+from unittest.mock import patch, PropertyMock, MagicMock
+import pytest
+
+from gsy_e_sdk.redis_aggregator import RedisAggregator, RedisAggregatorAPIException
+
+TEST_AGGREGATOR_NAME = "TestAgg"
```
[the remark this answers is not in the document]
ERafaelMartinez: I believe it is a not None value at the moment of creation of the aggregator. And as it is used a couple of times I decided to define it in sake of cleanness.

unit_tests/test_redis_aggregator.py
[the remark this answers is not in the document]
Hassan754: To be honest, I think that I actually am against this one assertion per test case rule, I find it unreasonable and it would exponentially prolong our test modules, imo if a test is related to one specific event (in our case initialization of the aggregator) it's totally okay to write several assertions for this particular event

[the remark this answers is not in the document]
fievelk: The one assertion per test makes sense IMO, but it's not necessary all the time. In simpler cases, I think this is not needed.

Longer version: the rationale behind the one-assertion "rule" is that each unit test should fail for only one reason, because this will save us from spending time debugging the reason why the test failed. From a different perspective: each test should only be set up for one single scenario (the one we're testing). If we test two different situations, this is not good because we can't immediately know which part of our setup (mocks etc) caused the test to fail. Using one assertion makes things easier because it forces us to keep the setup atomic/simple for our unit test.
Also, the one-assertion rule works best for methods that are kept short and that respect the single-responsibility principle.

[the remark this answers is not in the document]
Hassan754: When we have several scenarios (perhaps conditional statements in the code), we could split the one single test method into several scenario based tests where we append the event/action/scenario description in the name of test.
This doesn't necessarily mean that we should only have one assertion for these tests, because each node (conditional statement) in our codebase can lead to multiple changes elsewhere (so we should be able to track all of these changes)
In general, I agree that the one assertion per test CAN make sense for methods that follow the single responsibility principle, especially ones that change or return states (we can assert the values returned/set)

unit_tests/test_redis_aggregator.py
ERafaelMartinez: To address your comment, I have added one more assertion about the object being called. 
However, I'm not sure on how to assert that a certain attribute (set in the constructor) stores the mocked object in this case. 
I have tried the following for example:
```
    def test_constructor_redis_db_instantiated():
        with patch("gsy_e_sdk.redis_aggregator.StrictRedis"
                   ) as mocked_class:
            agg = RedisAggregator(aggregator_name=TEST_AGGREGATOR_NAME)
            mocked_class.from_url.assert_called_with(LOCAL_REDIS_URL)
            assert agg.redis_db is mocked_class
```

This kind of test fails, because AFAIS, when the aggregator is created, a _StrictRedis-mocked_ object is instantiated. When I look for the assertion to see wether the attribute has this mock stored in there, another mock object is created and used to compare. As these two are different, then the test fails. 

Do you have any suggestion on how I could assert that a mocked object is stored in a class attribute?

unit_tests/test_redis_aggregator.py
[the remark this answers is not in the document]
ERafaelMartinez: I see your point and finally understand it. 
I have been working towards a cleaner implementation of the tests. I stoped mocking these private methods and making assertions about the side effects due to them. I will push those changes later when I end up addressing all of the present comments.

unit_tests/test_redis_aggregator.py
fievelk: What we're trying to do here is basically stubbing the return value of `StrictRedis.from_url()`. So we don't need to mock the entire `StrictRedis` object. I took the liberty to try this out, and this seems to work:
```python
@staticmethod
def test_constructor_redis_db_instantiated(aggregator):
    redis_db_mock = MagicMock()
    with patch("gsy_e_sdk.redis_aggregator.StrictRedis.from_url", return_value=redis_db_mock
           ) as from_url_mock:
        agg = RedisAggregator(aggregator_name=TEST_AGGREGATOR_NAME, redis_url="some-url")
        from_url_mock.assert_called_with("some-url")
        assert agg.redis_db is redis_db_mock
```
As you can see, I've instantiated the aggregator passing the `redis_url` argument, because that value will be used to call `StrictRedis.from_url()`. 
The other thing I've done is explicitly creating a `MagicMock` instance, so that we could re-use it in the end to make sure that it's the same instance used to populate the `self.redis_db` variable.

Let me know if this makes sense or if you'd rather change something.
P.S.: Pylint complains about the `aggregator` parameter but I didn't change it in the example to save some time.

unit_tests/test_redis_aggregator.py
ERafaelMartinez: I'm not sure I understand what you mean here. This test was to test that when the _command_buffer_ has no elements (length zero) , the `execute_batch_commands` method returns `None` due to its own logic. 
The mock in this test is in order to fix this length zero as a return value, and therefore "force" the execute_batch_command to return `None`. 

I might not have understood what you said though.

unit_tests/test_redis_aggregator.py
ERafaelMartinez: Indeed

ERafaelMartinez: Yes it is possible. However, at this stage of the _sub-tests_, the `timeout_blocking` mock is not yet needed and therefore would be unnecessarily mocked here. 
It is also not perjudicial, as it would not affect nothing just having it there.
I can use the fixture here so it looks cleaner, but I'll leave that call to you.

unit_tests/test_redis_aggregator.py
```diff
@@ -0,0 +1,238 @@
...
+    @staticmethod
+    def test_calculate_grid_fee_returns_expected(aggregator):
+        aggregator.grid_fee_calculation.calculate_grid_fee.return_value = "TEST_OK"
+        ret_val = aggregator.calculate_grid_fee("start_market",
+                                                "target_market",
+                                                "current_market_fee")
+        assert ret_val == "TEST_OK"
+
+
+def test_get_uuid_from_area_name_returns_expected(aggregator):
```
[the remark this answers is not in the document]
ERafaelMartinez: This is so, because there were no other tests regarding the same method in order to group them all together in one single class.
I thought I could maybe include these loose tests in a class called

```
class TestRedisAggregator<name_which_describes_loose_tests>:
    def @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking"):
        # ...
```

but I did not find any suitable name. Also, in the past PR there were no comments about an analogous case so I assumed there was no problem. I'm open to suggestions.

unit_tests/test_redis_aggregator.py
ERafaelMartinez: Ok, that makes sense to me. Will replicate that in the other tests. Thanks!

unit_tests/test_redis_aggregator.py
ERafaelMartinez: I addressed your comment by implementing something similar to [this](https://github.com/gridsingularity/gsy-e-sdk/pull/197#discussion_r773360128).

unit_tests/test_redis_aggregator.py
Hassan754: @ERafaelMartinez probably Pierpaolo meant that you should mock the `commands_buffer_length` property instead of mocking the `buffer_length` of the `ClientCommandBuffer`, because in fact the `execute_batch_commands` checks for the length of `commands_buffer_length`, which in its own turn checks the `buffer_length`

unit_tests/test_redis_aggregator.py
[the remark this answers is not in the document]
Hassan754: This also caught my eyes, I prefer grouping all tests in a class (well I personally think that all of these tests could be grouped in a single class) however I thought that us it's okay because we know that all of the redis aggregator tests are in the same module `test_redis_aggregator`
So I'm okay with both approaches

unit_tests/test_redis_aggregator.py
```diff
@@ -0,0 +1,329 @@
...
+TEST_BATCH_COMMAND_DICT = {
+    TEST_DEVICE_UUID_1: ["TEST_BATCH_COMMAND"],
+    TEST_DEVICE_UUID_2: ["TEST_BATCH_COMMAND"]
+}
+
+TEST_TRANSACTION_ID_RESPONSE = "TEST_TRANSACTION_ID_RESPONSE"
+TEST_TRANSACTION_ID_BUFFER = {
+    TEST_TRANSACTION_ID: {"transaction_id": TEST_TRANSACTION_ID,
+                          "response": TEST_TRANSACTION_ID_RESPONSE}
+}
```
Hassan754, line 27: I don't see this is used anywhere

unit_tests/test_redis_aggregator.py
```diff
@@ -0,0 +1,329 @@
...
+                           "price": 1,
+                           "seller": "test_seller"},
+                 "bid": {"energy": 1,
+                         "price": 1,
+                         "buyer": "test_buyer"},
+                 }
+
+TEST_TRANSACTION_ID_RESPONSE_BUFFER = {TEST_TRANSACTION_ID: TEST_RESPONSE}
+
+TEST_BATCH_RESPONSES = {"TEST_ASSET_UUID": [TEST_RESPONSE]}
```
Hassan754, line 41: We can get rid of this one too, the reason I push for decreasing the number of these variables because we would probably end up debugging where all of them are used/defined and this may lead to some issues

It would be great if you can remove all variables that are only used once, and remove these global variables that are only used within 1 test case

unit_tests/test_redis_aggregator.py
```diff
@@ -62,44 +96,98 @@ def fixture_mock_client_command_buffer(mocker):
...
+        assert agg.executor is ThreadPoolExecutor_mock
 
     @staticmethod
-    def test_constructor_connect_and_subscribe_called():
-        with patch("gsy_e_sdk.redis_aggregator."
-                   "RedisAggregator._connect_and_subscribe") as mocked_method:
-            RedisAggregator(aggregator_name=TEST_AGGREGATOR_NAME)
-            mocked_method.assert_called()
+    @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
+    def test_constructor_connect_and_subscribe_side_effects_1():
```
Hassan754, line 155: Are you planning on further splitting this method? I vote for doing this

Hassan754 (commented):
Left some comments, will review again once addressed
One side note, it would be great if you can avoid pushing 1 major commit with description "Addressed comments"
First of all we want to use the imperative form, secondly it will be easier for us to review each set of related changes in a single commit, for instance: "Test the side effects of connect_and_subscribe"
Of course not all proposals deserve their own commit, you'll have to find the balance here

ERafaelMartinez:
Thank you for you comments. I have addressed them in the latest commits.

unit_tests/test_redis_aggregator.py
ERafaelMartinez: Addressed in [2b2611e](https://github.com/gridsingularity/gsy-e-sdk/pull/197/commits/2b2611e3b6fb37d7a1a217f1f8866b65660cbb2b). Thanks.

unit_tests/test_redis_aggregator.py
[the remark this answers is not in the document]
ERafaelMartinez: So what would be our agreement here @fievelk ?

unit_tests/test_redis_aggregator.py
```diff
@@ -0,0 +1,334 @@
...
+        """Test  side effects due to the methods called with _connect_and_subscribe().
+        -> Side effects of _subscribe_to_aggregator_response_and_start_redis_thread
+        """
+
+        aggregator = RedisAggregator(aggregator_name=TEST_AGGREGATOR_NAME)
+        channel_dict_1 = {"aggregator_response": aggregator._aggregator_response_callback}
+
+        aggregator.pubsub.run_in_thread.assert_called_with(daemon=True)
```
hannesdiedrich, line 153: Can you please explain? How can the `aggregator.pubsub.run_in_thread` be a mock? I do not see the patch or mock command.

hannesdiedrich: The same applies to the `aggregator.redis_db.publish` in the next test method.

unit_tests/test_redis_aggregator.py
```diff
@@ -0,0 +1,334 @@
...
+                          f"/batch_commands": aggregator._batch_response,
+                          }
+
+        aggregator.pubsub.psubscribe.assert_called_with(**channel_dict_2)
+
+    @staticmethod
+    @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
+    def test_constructor_lock_instantiated():
+        with patch("gsy_e_sdk.redis_aggregator.Lock"
+                   ) as mocked_class:
```
hannesdiedrich, line 189: Nit
```suggestion
        with patch("gsy_e_sdk.redis_aggregator.Lock") as mocked_class:
```

unit_tests/test_redis_aggregator.py
hannesdiedrich: I see it in multiple methods here and wonder how this works. Thanks a lot in advance for explaining!

unit_tests/test_redis_aggregator.py
```diff
@@ -0,0 +1,334 @@
...
+
+        aggregator.execute_batch_commands(is_blocking=False)
+        aggregator.redis_db.publish.assert_called_with(batched_channel_input,
+                                                       json.dumps(batched_command_input))
+
+    @staticmethod
+    @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
+    def test_execute_batch_commands_api_time_out_raise_exception(aggregator):
+        with patch("gsy_e_sdk.redis_aggregator.wait_until_timeout_blocking",
+                   side_effect=AssertionError):
```
hannesdiedrich, line 269: No need to patch twice right?
```suggestion
    def test_execute_batch_commands_api_time_out_raise_exception(aggregator):
        with patch("gsy_e_sdk.redis_aggregator.wait_until_timeout_blocking",
                   side_effect=AssertionError):
```

unit_tests/test_redis_aggregator.py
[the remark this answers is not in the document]
hannesdiedrich: I left a similar question in [here](https://github.com/gridsingularity/gsy-e-sdk/pull/194#discussion_r775394173)
I think that grouping could be useful, but then it is also complicated, because: Have a look at the `test_constructor_grid_fee_calculation_instantiated`, IMO it could also reside here in the `TestRedisAggregatorCalculateGridFee`. 
I do not have a strong opinion here, as we do not have to search for tests very often so having one test class for all test methods works, too.

unit_tests/test_redis_aggregator.py
```diff
@@ -0,0 +1,334 @@
...
+        aggregator.calculate_grid_fee("start_market",
+                                      "target_market",
+                                      "current_market_fee")
+        aggregator.grid_fee_calculation.calculate_grid_fee.assert_called_with(
+            "start_market",
+            "target_market",
+            "current_market_fee")
+
+    @staticmethod
+    def test_calculate_grid_fee_returns_expected(aggregator):
```
hannesdiedrich, line 305: Actually, you are testing whether the `aggregator.calculate_grid_fee` calls the `aggregator.grid_fee_calculation.calculate_grid_fee`, you are not testing whether the the grid_fee value is valid, right? SO I vote for renaming this method.

hannesdiedrich (commented):
Some comments, once addressed, LGTM

unit_tests/test_redis_aggregator.py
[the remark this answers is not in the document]
fievelk: @ERafaelMartinez I've started a conversation about this on Slack. I would say that the general agreement is this:
- tests should be defined inside of classes. Even though Pytest examples define functions in the module scope, we've decided to avoid this (you can check the conversation on slack for more information).
- there should be only one test class for each tested class. There can be exceptions like in this PR, where you created one class for each method; however, the first approach should be preferred when possible.

So class-based is fine. We would like to avoid defining test functions in the module scope. If you've decided to create 1 class for each tested module, please do that also for methods that need only one test. If this means creating too many test classes, then it might be another good reason to actually avoid this approach and just create one single test class for each tested class.

unit_tests/test_redis_aggregator.py
```diff
@@ -0,0 +1,334 @@
...
+    }
+}
+
+
+@pytest.fixture(autouse=True)
+def fixture_mock_connections(mocker):
+    mocker.patch("gsy_e_sdk.redis_aggregator.StrictRedis")
+    mocker.patch("gsy_e_sdk.redis_aggregator.ThreadPoolExecutor")
+    mocker.patch("gsy_e_sdk.redis_aggregator.Dict",
+                 return_value=TEST_MESSAGE_DICT)
```
fievelk, line 63: Is this still being used? Tests pass even when I comment this line out.

unit_tests/test_redis_aggregator.py
```diff
@@ -0,0 +1,334 @@
...
+        with patch("gsy_e_sdk.redis_aggregator.ClientCommandBuffer",
+                   return_value=client_command_buffer_mock) as mocked_class:
+            agg = RedisAggregator(aggregator_name=TEST_AGGREGATOR_NAME)
+            mocked_class.assert_called()
+        assert agg._client_command_buffer is client_command_buffer_mock
+
+    @staticmethod
+    @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
+    def test_constructor_threadpoolexecutor_instantiated():
+        ThreadPoolExecutor_mock = MagicMock
```
fievelk, line 136: Nit:
```suggestion
        thread_pool_executor_mock = MagicMock
```

unit_tests/test_redis_aggregator.py
```diff
@@ -62,44 +96,98 @@ def fixture_mock_client_command_buffer(mocker):
...
-    @pytest.mark.usefixtures("mock_grid_fee_calculation")
-    def test_constructor_grid_fee_calculation_instantiated(aggregator):
-        assert aggregator.grid_fee_calculation is not None
+    @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
+    def test_constructor_grid_fee_calculation_instantiated():
+        grid_fee_calculation_mock = MagicMock()
+        with patch("gsy_e_sdk.redis_aggregator.GridFeeCalculation",
+                   return_value=grid_fee_calculation_mock) as mocked_class:
+            agg = RedisAggregator(aggregator_name=TEST_AGGREGATOR_NAME)
+            mocked_class.assert_called()
```
fievelk, line 105: Nit: I have the feeling that this is redundant since the assertion below implies this already.

unit_tests/test_redis_aggregator.py
fievelk: @hannesdiedrich the reason is that `StrictRedis` and `ThreadPoolExecutor` have been mocked in this fixture: https://github.com/gridsingularity/gsy-e-sdk/pull/197/files#diff-701f7f3f361fabc75b8c82fd7b3833b77782af10ec9c24b196a4c6c9524de42eR58-R63
The fixture has `autouse=True`, which means that it is automatically applied to all individual test functions in this module.

I'm usually a bit skeptical of using `autouse=True` because it's a bit too broad for my personal taste (it could be used in functions that don't need it). But I suppose that in this case Eduardo wanted to avoid mistakenly calling the "real" redis implementation. That's why in a previous comment I suggested using [fakeredis](https://github.com/jamesls/fakeredis) instead, to deal with this in all tests.

unit_tests/test_redis_aggregator.py
```diff
@@ -62,44 +96,98 @@ def fixture_mock_client_command_buffer(mocker):
...
+        aggregator.pubsub.psubscribe.assert_has_calls([call(**channel_dict_1),
+                                                       call(**channel_dict_2)])
 
     @staticmethod
-    def test_constructor_lock_instantiated(aggregator):
-        assert aggregator.lock is not None
+    @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
+    def test_constructor_lock_instantiated():
+        with patch("gsy_e_sdk.redis_aggregator.Lock"
+                   ) as mocked_class:
```
fievelk, line 189: Why aren't you mocking the exact `Lock` instance as you've done in the other tests? That would end up in an assertion like
```python
assert aggregator.lock is lock_mock
```

unit_tests/test_redis_aggregator.py
```diff
@@ -0,0 +1,334 @@
...
+        with patch("gsy_e_sdk.redis_aggregator.ThreadPoolExecutor",
+                   return_value=ThreadPoolExecutor_mock) as mocked_class:
+            agg = RedisAggregator(aggregator_name=TEST_AGGREGATOR_NAME)
+            mocked_class.assert_called()
+        assert agg.executor is ThreadPoolExecutor_mock
+
+    @staticmethod
+    @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
+    def test_constructor_connect_and_subscribe_side_effects_1():
+        """Test  side effects due to the methods called with _connect_and_subscribe().
```
fievelk, line 146: Nit
```suggestion
        """Test side effects due to the methods called with _connect_and_subscribe().
```

unit_tests/test_redis_aggregator.py
```diff
@@ -0,0 +1,334 @@
...
+        aggregator = RedisAggregator(aggregator_name=TEST_AGGREGATOR_NAME)
+        channel_dict_1 = {"aggregator_response": aggregator._aggregator_response_callback}
+
+        aggregator.pubsub.run_in_thread.assert_called_with(daemon=True)
+        aggregator.pubsub.psubscribe.assert_has_calls([call(**channel_dict_1)])
+
+    @staticmethod
+    @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
+    def test_constructor_connect_and_subscribe_side_effects_2():
+        """Test  side effects due to the methods called with _connect_and_subscribe().
```
fievelk, line 159: Nit
```suggestion
        """Test side effects due to the methods called with _connect_and_subscribe().
```

unit_tests/test_redis_aggregator.py
```diff
@@ -0,0 +1,334 @@
...
+                "transaction_id": TEST_TRANSACTION_ID}
+
+        aggregator.redis_db.publish.assert_called_with("aggregator", json.dumps(data))
+        assert TEST_TRANSACTION_ID in aggregator._transaction_id_buffer
+        assert aggregator.aggregator_uuid is TEST_TRANSACTION_ID
+
+    @staticmethod
+    @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
+    def test_constructor_connect_and_subscribe_side_effects_3():
+        """Test  side effects due to the methods called with _connect_and_subscribe().
```
fievelk, line 173: Nit:
```suggestion
        """Test side effects due to the methods called with _connect_and_subscribe().
```

fievelk:
I've left some comments. I think we're going in the right direction here, thanks for dealing with all comments 👍

unit_tests/test_redis_aggregator.py
ERafaelMartinez: I've confirmed that this line is apparently not necessary in order for tests to succeed. 
Originally, I thought I should mock the return value of this object as it is used as default _message_ for methods such as `_batch_response()`. I thought I would allow this method to execute (for it is a private method) as it should, but using the dictionary I forced it to use as a message instead of the one which lives in `typing` module. 
As this message Dict seem to hold information about the simulation, and transactions, I assumed it is connected to the redis server and gets updated after some process. However,  I do not know precisely how this is achieved, so my approach was to create a mocked Dict message inferred/reconstructed from the information the different methods request from this message.
In conclusion, I recreated this Dict message and mocked the actual Dict message used in the RedisAggregator in order to avoid a possible connection request with the server. The fact that even thought this lines are commented the tests succeed might imply this Dict does not make any kind of request to the server and therefore is not necessary to mock it.

ERafaelMartinez: I'll address your comment by deleting these lines. However I would appreciate some insights in order to comprehend how these Dict messages work...

unit_tests/test_redis_aggregator.py
ERafaelMartinez: You are right!

unit_tests/test_redis_aggregator.py
ERafaelMartinez: I'm not really sure how the compiler let that styling pass. I should have been warned about that...

unit_tests/test_redis_aggregator.py
ERafaelMartinez: What Pierpaolo explains is precisely what happens in these tests. 
In the fixture he mentions, the StrictRedis class and ThreadPoolExecutor are mocked to avoid those external connections. 
I set `autouse=True` for them to be used everytime an aggregator instance gets created (which applies to each test case in this case). And therefore, for each instance of the aggregator, this attributes would be mocked and therefore no connections would be stablished. 

I had not understand your previous suggestion about the fakeredis. Regarding that...
If we follow this approach, should I set a patch for StrictRedis whose return value is the fakeredis instance? Or how should that be handled?

unit_tests/test_redis_aggregator.py
ERafaelMartinez: I was asking myself exactly the same. There is not reason. Probably I just missed it.

unit_tests/test_redis_aggregator.py
ERafaelMartinez: Actually, the assertion regarding wether it was called or not is within the test above. 

I see now that you point this out, that this test might not be needed, because if the GridFeeCalculation instance works properly (and therefore its method calculate_grid_fee does as well), then this test should always pass. So this is somehow redundant. 
I believe now that the right thing to do here is get rid of this test.

unit_tests/test_redis_aggregator.py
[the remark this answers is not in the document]
ERafaelMartinez: As there would be to many classes due to this one-test-only methods, I'll address this discussion here by refactoring all test in a single class named _TestRedisAggregator_.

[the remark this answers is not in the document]
ERafaelMartinez: Thanks for the discussion.

unit_tests/test_redis_aggregator.py
fievelk: I was hoping that fakeredis would work like fakes3 (where you just point to a different URL to use the fake server). Unfortunately, it looks like you actually need to import it directly and use it to patch the "real" redis classes (similar to what you've done here). So, in my opinion, it's not as nice as I originally thought.
If you have the time, you could check the possible alternatives and tell us if you find something that you think is worth it. But I wouldn't consider this to be part of this PR.

unit_tests/test_redis_aggregator.py
fievelk: I think there might be a misunderstanding in this case (please correct me if I didn't get your message!). In the line you mentioned:
```python
def _batch_response(self, message: Dict) -> None:
```
the `Dict` object is not the default value of `message` (in that case it would have been `message = Dict`, with the `=` sign in place of `:`).
Using `:` after a parameter we're defining the type of the parameter. So in this case we know that the `message` parameter is expected to be a dictionary. These are called "type hints" ([see docs](https://docs.python.org/3.8/library/typing.html)). Python is not strictly typed, so these types won't generate errors at runtime; but there are tools (like mypy) that automatically check if the types are respected. PyCharm should also have similar features built-in.

unit_tests/test_redis_aggregator.py
```diff
@@ -125,25 +99,24 @@ def test_constructor_pubsub_instantiated(aggregator):
...
     @staticmethod
     @pytest.mark.usefixtures("mock_transaction_id_and_timeout_blocking")
     def test_constructor_threadpoolexecutor_instantiated():
-        ThreadPoolExecutor_mock = MagicMock
+        thread_pool_executor_mock = MagicMock
         with patch("gsy_e_sdk.redis_aggregator.ThreadPoolExecutor",
-                   return_value=ThreadPoolExecutor_mock) as mocked_class:
+                   return_value=thread_pool_executor_mock) as mocked_class:
             agg = RedisAggregator(aggregator_name=TEST_AGGREGATOR_NAME)
             mocked_class.assert_called()
```
fievelk, line 113: Nit: please not that there the same as https://github.com/gridsingularity/gsy-e-sdk/pull/197#discussion_r775449592 applies. This can also be applied to other parts of the PR.

fievelk:
This can't be merged until the CI errors are fixed (they're related to pytest-mock not being installed).

unit_tests/test_redis_aggregator.py
ERafaelMartinez: I'll have a look and see if some nice refactor can be achieved by using fakeredis.

unit_tests/test_redis_aggregator.py
ERafaelMartinez: You're right, I misread the `:` by `=` . Thank you for clearing my confusion!

unit_tests/test_redis_aggregator.py
ERafaelMartinez: I see this. I've been thinking however: Would you say this also applies to the `assert_called_with(_some_args_)` expression?

ERafaelMartinez:
I just pushed some changes addressing your latest comments @fievelk.
Regarding the not-passing tox tests, I'd say we could wait this PR to be merged after TESTS-106 which has the modifications done to the build files and the integration tests.

unit_tests/test_redis_aggregator.py
fievelk: I think this depends on the type of message. 
- If we're talking about an outgoing query, then we don't really care about what arguments we pass. That's because we can just ignore the outgoing query message, without mocking them or setting expectations. As we've seen in the past, we just use their outputted results.
- If we're talking about an outgoing command, then we probably care about the parameters. That's because we will not check the side-effects of outgoing commands, but we will just create a mock and expect to be sending the right content to the other class/method.
- If we're talking about private commands (sent to self), then we usually don't mock/stub/patch them. We just check their side-effects inside of our instance.

Also, please note the difference between "biding to the implementation" and "testing the interface". If an interface (public method) accepts certain parameters, then it's correct to check that we're passing those parameters. We mostly talk about "implementation" when we're referring to methods that are not part of the public interface. It might not be always the case, but it might help.

fievelk (approved):

Hassan754 (approved):

Hassan754:
It would be great if we can merge this PR as soon as possible, it's already too late taking into consideration that it fixes a CI tests bug

PR #202, opened 2021-12-29
feature/TESTS-107

ERafaelMartinez:
Add unit tests for RestAssetClient class.


unit_tests/test_rest_asset_client.py
```diff
@@ -0,0 +1,297 @@
+# pylint: disable=missing-function-docstring, no-member, too-many-public-methods
+import uuid
+from unittest.mock import patch, MagicMock
+# import os
```
Hassan754, line 4: to be removed

unit_tests/test_rest_asset_client.py
```diff
@@ -0,0 +1,297 @@
...
+
+@pytest.mark.usefixtures("mock_environment_use_functions")
+class TestRestAssetClient:
+    """Test methods for RestAssetClient class."""
+
+    @staticmethod
+    @pytest.mark.parametrize("set_value, expected_ret_val",
+                             [(TEST_SIMULATION_ID, TEST_SIMULATION_ID),
+                              (None, TEST_SIMULATION_ID)])
+    def test_constructor_simulation_id_setup(set_value, expected_ret_val):
```
Hassan754, line 64: Nit, a docstring here could be useful (please feel free to rephrase it in your style, mine is just an example)
```suggestion
    def test_constructor_simulation_id_setup(set_value, expected_ret_val):
    """Test whether the simulation id will be set regardless of whether it was explicitly passed or not (should read the env variable)"""
```

unit_tests/test_rest_asset_client.py
```diff
@@ -0,0 +1,297 @@
...
+    def test_constructor_domain_name_setup(set_value, expected_ret_val):
+        client = RestAssetClient(asset_uuid=TEST_ASSET_UUID,
+                                 domain_name=set_value)
+        assert client.domain_name == expected_ret_val
+
+    @staticmethod
+    @pytest.mark.parametrize("set_value, expected_ret_val",
+                             [(TEST_WEBSOCKETS_DOMAIN_NAME, TEST_WEBSOCKETS_DOMAIN_NAME),
+                              (None, TEST_WEBSOCKETS_DOMAIN_NAME)])
+    def test_constructor_websockets_domain_name_setup(set_value, expected_ret_val):
```
Hassan754, line 82: I feel like the first 3 tests can be grouped in one single test, they're all testing whether the client will read the environment variables in case these configurations are not explicitly set (for instance in the first test you could've also tested the websocket_domain_name and it would've been set as well)
By the way, It would be better if you can pass values different than the ones in the environment variables, this will make tests relevant (making sure that the values you passed are prioritized over the environment ones)

Hassan754 (approved):
LGTM, however I see that there are some redundant mocks and patches in the module, I think that this conflicts with the actual usage of mocker library, it would be great if you can find a solution for this

unit_tests/test_rest_asset_client.py
```diff
@@ -0,0 +1,294 @@
...
+    @pytest.mark.usefixtures("mock_transaction_id")
+    def test_register_request_posted_waits_for_command_response_and_return_expected(client):
+        with patch("gsy_framework.client_connections.utils.post_request",
+                   return_value=True):
+            client.dispatcher = MagicMock()
+            client.dispatcher.wait_for_command_response.return_value = TEST_COMMAND_RESPONSE
+
+            assert client.register() == TEST_COMMAND_RESPONSE
+            client.dispatcher.wait_for_command_response.assert_called_with(
+                "register", TEST_TRANSACTION_ID, timeout=15 * 60)
```
hannesdiedrich, line 163: Not sure, but we do not want to test the setting of TEST_COMMAND_RESPONSE, so please use the variable. Thanks!
```suggestion
                "register", TEST_TRANSACTION_ID, timeout= TEST_COMMAND_RESPONSE)
```

unit_tests/test_rest_asset_client.py
```diff
@@ -0,0 +1,294 @@
...
+
+
+@pytest.fixture(name="mock_transaction_id")
+def mock_transaction_id_and_timeout_blocking(mocker):
+    mocker.patch("gsy_framework.client_connections.utils.uuid.uuid4",
+                 return_value=TEST_TRANSACTION_ID)
+
+
+@pytest.fixture(name="client")
+def fixture_rest_asset_client(mock_environment_use_functions):  # pylint: disable=unused-argument
```
hannesdiedrich, line 52: why not setting random values for simulation_id, domain_name and websocket_domain and omitting passing the mock_environment_use_functions here?

unit_tests/test_rest_asset_client.py
```diff
@@ -0,0 +1,294 @@
...
+    mocker.patch("gsy_e_sdk.clients.rest_asset_client.domain_name_from_env",
+                 return_value=TEST_DOMAIN_NAME)
+    mocker.patch("gsy_e_sdk.clients.rest_asset_client.websocket_domain_name_from_env",
+                 return_value=TEST_WEBSOCKETS_DOMAIN_NAME)
+
+
+@pytest.fixture(name="mock_transaction_id")
+def mock_transaction_id_and_timeout_blocking(mocker):
+    mocker.patch("gsy_framework.client_connections.utils.uuid.uuid4",
+                 return_value=TEST_TRANSACTION_ID)
```
hannesdiedrich, line 48: Can you please explain why this is needed? Why mocking a standard library feature?

hannesdiedrich (commented):
Left some comments, LGTM once addressed.

Hassan754 (approved):

unit_tests/test_rest_asset_client.py
ERafaelMartinez: Here I mimicked the arguments with which `dispatcher.wait_for_command_response `method is called. I agree the `15 * 60` is really specific. If you agree, I can change this assertion for just to assert wether it has been called or not, regardless of the arguments.

I'm afraid your suggestion would not work because TEST_COMMAND_RESPONSE is not an argument used when calling that method. You may have thought of `REGISTER_COMMAND_TIMEOUT` though...

ERafaelMartinez:
```suggestion
                "register", TEST_TRANSACTION_ID, timeout=REGISTER_COMMAND_TIMEOUT)
```
It could work like this and the variable being imported from `rest_asset_client.py` as it is defined there...

unit_tests/test_rest_asset_client.py
ERafaelMartinez: The values for those attributes still being random. It is just that they are set up directly from the mocked "read_from_env" functions. 
However I see maybe what catch your eyes, being that it would not be necessary to apply the mock to this fixture if the parameters were explicitly set up.

unit_tests/test_rest_asset_client.py
ERafaelMartinez: In this case I mock it to make assertions using the id, forcing it to return a specific id previously generated. 

For example, as a side effect on register (due to `_post_request() `private method), a transaction id is generated and added to a dictionary used to post a request.

As the `post_request` message must be sent for the `rest_asset_client` to work properly, I made an assertion this being called with some explicit arguments. One of which is a dictionary containing a particular `transaction_id`. If I were to let `uuid4()` to generate a random id, there would be no way (AFAIK) for me to know this and assert if the post was made with the expected transaction_id or not.

However, if I assume `uuid4()` works properly (which it should) I can safely mock it to return a convenient uuid which I know beforehand, to assert that `post_request` was called with a dictionary containing this id.

Hassan754 (approved):

unit_tests/test_rest_asset_client.py
hannesdiedrich: I see. Thanks a lot for the information!

hannesdiedrich (approved):
LGTM

PR #207, opened 2022-01-13
feature/D3ASIM-3695

ERafaelMartinez:
Integration test for testing the connection to the settlement markets. This PR contains the feature file and the aggregator used in the test.
For this test to work, the file in [this other PR](https://github.com/gridsingularity/gsy-e/pull/1358) in gsy-e is also needed.

### DIFF
--- a/integration_tests/settlement_market.feature
+++ b/integration_tests/settlement_market.feature
@@ -1,0 +1,9 @@
+Feature: Settlement Market Tests
+
+Scenario: Assets can post bids and offers
+  Given redis container is started
+  And gsy-e is started in paused mode using setup strategy_tests.external_devices_settlement_market (-t 60s -s 60m -d 4h --slot-length-realtime 2s)
+  When the gsy-e-sdk is connecting to gsy-e with test_aggregator_settlement
+  And gsy-e is resumed
+  And the gsy-e-sdk is connected to the gsy-e until finished
+  And the gsy-e-sdk does not report errors
--- a/integration_tests/test_aggregator_settlement.py
+++ b/integration_tests/test_aggregator_settlement.py
@@ -1,0 +1,174 @@
+# pylint: disable=too-many-instance-attributes, broad-except
+import logging
+import traceback
+import random
+
+from gsy_framework.constants_limits import DATE_TIME_FORMAT
+from pendulum import from_format
+
+from gsy_e_sdk.clients.redis_asset_client import RedisAssetClient
+from integration_tests.test_aggregator_base import TestAggregatorBase
+
+
+class SettlementAggregator(TestAggregatorBase):
+    """Aggregator class to test the connections to settlement
+     markets:
+     Done by asserting whether bids/offers are successfully
+     posted on these markets."""
+
+    def __init__(self, *args, **kwargs):
+
+        self.base_forecast_load = 10
+        self.base_measurement_load = self._calculate_random_deviated_energy(
+            self.base_forecast_load, 5)
+        self.base_forecast_pv = 10
+        self.base_measurement_pv = self._calculate_random_deviated_energy(
+            self.base_forecast_pv, 5)
+
+        self._has_tested_settlement_posts = False
+        super().__init__(*args, **kwargs)
+
+        # self._set_first_measurement_forecast()
+
+    def _setup(self):
+        load_client = RedisAssetClient("forecast-measurement-load", pubsub_thread=self.pubsub)
+        load_client.select_aggregator(self.aggregator_uuid)
+        self.load_uuid = load_client.area_uuid
+
+        pv_client = RedisAssetClient("forecast-measurement-pv", pubsub_thread=self.pubsub)
+        pv_client.select_aggregator(self.aggregator_uuid)
+        self.pv_uuid = pv_client.area_uuid
+
+    def _set_first_measurement_forecast(self):
+        self.add_to_batch_commands.set_energy_forecast(
+            asset_uuid=self.load_uuid,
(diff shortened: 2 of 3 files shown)


integration_tests/test_aggregator_settlement.py
```diff
@@ -0,0 +1,174 @@
...
+            asset_uuid=self.pv_uuid,
+            energy_forecast_kWh={"2022-01-12T01:00": self.base_forecast_pv}) \
+            .set_energy_measurement(
+            asset_uuid=self.load_uuid,
+            energy_measurement_kWh={"2022-01-12T00:00": self.base_measurement_load}) \
+            .set_energy_measurement(
+            asset_uuid=self.pv_uuid,
+            energy_measurement_kWh={"2022-01-12T00:00": self.base_measurement_pv})
+
+        self.send_batch_commands()
```
hannesdiedrich, line 56: Please remove as it is not used (also the commented call of this method). Thanks!

integration_tests/settlement_market.feature
```diff
@@ -0,0 +1,9 @@
+Feature: Settlement Market Tests
+
+Scenario: Assets can post bids and offers
+  Given redis container is started
+  And gsy-e is started in paused mode using setup strategy_tests.external_devices_settlement_market (-t 60s -s 60m -d 4h --slot-length-realtime 2s)
+  When the gsy-e-sdk is connecting to gsy-e with test_aggregator_settlement
```
hannesdiedrich, line 6: Where did you implement this step?

integration_tests/test_aggregator_settlement.py
```diff
@@ -0,0 +1,174 @@
...
+
+                # manage bids and offers in settlement markets
+
+                if self._can_place_in_settlement_market(asset_info):
+
+                    for market_slot, energy in asset_info["unsettled_deviation_kWh"].items():
+
+                        if energy and energy > 0:
+                            self.add_to_batch_commands.bid_energy(asset_uuid=area_uuid,
+                                                                  energy=energy / 2, price=30,
```
hannesdiedrich, line 141: why are you only sending half of the energy here? I think it would be a good idea bid all the energy in order to test the correct exposing of the unsettled energy.

integration_tests/test_aggregator_settlement.py
```diff
@@ -0,0 +1,174 @@
...
+    def on_market_cycle(self, market_info):
+        logging.info("market_info: %s", market_info)
+
+        try:
+            for area_uuid, area_dict in self.latest_grid_tree_flat.items():
+                if not area_dict.get("asset_info"):
+                    continue
+
+                # manage forecast and measurement information
+
```
hannesdiedrich, line 105: Please go easy on the extra blank lines. "Use blank lines in functions, sparingly, to indicate logical sections." 
Reference in PEP8: https://www.python.org/dev/peps/pep-0008/#blank-lines

hannesdiedrich (commented):
I wonder how you could test it locally. One of the test steps is not implemented. Once your test is run and passes, I will review again. 
Let me know if you need help with that and we can meet offline.

integration_tests/settlement_market.feature
ERafaelMartinez: I forgot to commit this file initially. Done here [f9271ad](https://github.com/gridsingularity/gsy-e-sdk/pull/207/commits/f9271ad5ac9928e039b368fba0e240b78a3a03a6).

integration_tests/test_aggregator_settlement.py
ERafaelMartinez: There was no reason other than I base on other test. Comment addressed here [7b5d0d7](https://github.com/gridsingularity/gsy-e-sdk/pull/207/commits/7b5d0d7ed599e94a7ebdf622b816d7d5e7453c05). I also added a test for what you mention.

hannesdiedrich (approved):
Thanks! LGTM

integration_tests/test_aggregator_settlement.py
```diff
@@ -0,0 +1,159 @@
+# pylint: disable=too-many-instance-attributes, broad-except
```
faizan2590, line 1: FYI, backend team had decided to not disable pylint at the whole module level, but only at places where it needs to be.

No need to adapt now, just to keep in mind for future PRS.

integration_tests/test_aggregator_settlement.py
```diff
@@ -0,0 +1,159 @@
...
+        return energy + sign * delta
+
+    @staticmethod
+    def _can_place_forecast_measurements(area_dict):
+        return "area_name" in area_dict and ("forecast-measurement" in area_dict["area_name"])
+
+    @staticmethod
+    def _next_market_slot(market_info):
+        return from_format(market_info["market_slot"], DATE_TIME_FORMAT) \
+            .add(minutes=60).format(DATE_TIME_FORMAT)
```
faizan2590, line 55: Use `()` instead of `\` as per [PEP 8](https://realpython.com/python-pep8/#indentation-following-line-breaks) recommendation

faizan2590 (approved):
Some minor enhancement proposal, other than that LGTM

integration_tests/test_aggregator_settlement.py
fievelk: It depends on the specific warning we're suppressing. For example, if we want to suppress the missing docstring warnings for an entire test module, it might make sense to put it on top rather than repeating it 20 times. If we're silencing just 2 or 3 warnings, then I would recommend placing the disable comment on the same line/scope.

PR #208, opened 2022-01-31
GSYE-122: As tester, I would like to move all integration tests to a separate repo.

ERafaelMartinez:
This PR aims to address GSYE-122 ticket. The changes suggested by it apply only to gsy-e-sdk repository.


ERafaelMartinez:
Hey @mclrch , could I get your help here please?
I'm not sure why the authentication is not working the same on this repo as in the others, and therefore the checks failing.

tox.ini
```diff
@@ -32,17 +32,28 @@ exclude = .tox,.cache,.pytest_cache
 [testenv:integrationtests]
 basepython = python3.8
 deps =
-	-r requirements/base.txt
+	-rrequirements/dev.txt
```
fievelk, line 35: Nit:
```suggestion
    -rrequirements/dev.txt
```

Please check also the rest of the PR, as I see that the same problem occurs in other parts.

fievelk: BTW: is there a reason why we're using `-rrequirements` rather than `-r requirements`?

tox.ini
```diff
@@ -32,17 +32,28 @@ exclude = .tox,.cache,.pytest_cache
...
+passenv = LANG TERM LANGUAGE LC_ALL LD_LIBRARY_PATH BRANCH INTEGRATION_TESTS_REPO_URL TARGET_REPO
 setenv =
     API_CLIENT_RUN_ON_REDIS = true
+	INT_TESTS_REPO_SSH_URL = git@github.com:gridsingularity/gsy-backend-integration-tests.git
```
fievelk, line 39: Here we don't necessarily need to use the SSH version, right? I have the feeling that this variable could be renamed to `INTEGRATION_TESTS_REPO_URL` (or whatever you like). Then the CI can fill its value with a SSH URL, while locally we can decide to also use HTTPS. What do you think?

fievelk, line 39: I have the feeling that we're creating unnecessary env variables. We have 3 variables that do more or less the same: `TARGET_REPO`, `INTEGRATION_TESTS_REPO_URL` and `INT_TESTS_REPO_URL_SSH`. I understood that the CI server uses HTTPS, while locally we might use SSH. But can't this can be tackled with just one variable?
Suppose we just use `TARGET_REPO`:
- the CI server can fill it with an HTTPS URL (in `development.yaml`, for example);
- for local runs we can fill it with the SSH URL (in `tox.in`).

Do you think this would work?

tox.ini
ERafaelMartinez: Will do, thank you.
AFAIK that's tox syntax. I thought I'd seen it on the documentation. However now that I looked directly for it I couldn't find it. 
I found [this](https://stackoverflow.com/questions/12258456/how-can-tox-install-the-modules-via-the-requirements-file/37450242#:~:text=What%20helped%20me%20is%20the%20following%20(the%20other%20solution%20didn%27t%20work%20for%20me)%3A) on stackoverflow though.

hannesdiedrich: Was it failing for you without the space?

tox.ini
ERafaelMartinez: I think is also confusing having so many env variables, specially all nested in order to determine the branch from which is being copied.
However, I'm not sure how that would work. At least it would not be handled automatically and would need to edit tox.ini file while working locally.

I think it could look something like this:
```suggestion
passenv = LANG TERM LANGUAGE LC_ALL LD_LIBRARY_PATH BRANCH INTEGRATION_TESTS_REPO
setenv =
    API_CLIENT_RUN_ON_REDIS = true
    # uncomment for local run using local repository
    # INTEGRATION_TESTS_REPO = /Users/SomeUser/PathToRepo/
allowlist_externals =
	git
	/bin/cp
	/bin/rm
commands_pre =
	pip install -e .
    git clone {env:INTEGRATION_TESTS_REPO:git@github.com:gridsingularity/gsy-backend-integration-tests.git} {envtmpdir}/gsy-backend-integration-tests
```

The thing we should be careful with is to not override the `INTEGRATION_TESTS_REPO` variable set by the workflow with one given inside the testenv, so the tox.ini on the remote repository should always have that one line commented.

I also thought initially for something like this:

```suggestion
passenv = LANG TERM LANGUAGE LC_ALL LD_LIBRARY_PATH BRANCH INTEGRATION_TESTS_REPO
setenv =
    API_CLIENT_RUN_ON_REDIS = true
    # uncomment for local run using local repository
    # INTEGRATION_TESTS_REPO = /Users/SomeUser/PathToRepo/
    # uncomment for local run using remote repository
    # INTEGRATION_TESTS_REPO = git@github.com:gridsingularity/gsy-backend-integration-tests.git
allowlist_externals =
	git
	/bin/cp
	/bin/rm
commands_pre =
	pip install -e .
    git clone {env:INTEGRATION_TESTS_REPO:} {envtmpdir}/gsy-backend-integration-tests
```

The command would look much cleaner, but if uncommenting the appropriate line is forgotten, then it would have nothing set as default.

On the first suggestion, one could deal with uncommenting the line only if interested on running the tests with a local branch.

Even more:

```suggestion
passenv = LANG TERM LANGUAGE LC_ALL LD_LIBRARY_PATH BRANCH INTEGRATION_TESTS_REPO
setenv =
    API_CLIENT_RUN_ON_REDIS = true
allowlist_externals =
	git
	/bin/cp
	/bin/rm
commands_pre =
	pip install -e .
    git clone {env:INTEGRATION_TESTS_REPO:git@github.com:gridsingularity/gsy-backend-integration-tests.git} {envtmpdir}/gsy-backend-integration-tests
```

This should work invoking the testenv as
`INTEGRATION_TESTS_REPO=/Users/SomeUser/PathToRepo/ tox -e integrationtests`
And wouldn't need any local changes of the tox.ini file. 

Which approach do you like the most?
Do you have any further ideas?

ERafaelMartinez: I believe the last approach is the closest to what you initially described.

tox.ini
ERafaelMartinez: I did not see any failures, but I noted that without the space they are not installed, or at least not explicitly mentioned on the building logs, just as the answer on stackoverflow I just posted mentions...

tox.ini
ERafaelMartinez: Addressed [here](https://github.com/gridsingularity/gsy-e-sdk/pull/208#:~:text=of%20environment%20variables.-,442d4f9,-Merge%20state). Will replicate on other repos.

ERafaelMartinez:
Added new commits to comply with agreements reached on related PR's.

hannesdiedrich (approved):
LGTM

tox.ini
fievelk: That's interesting! Maybe it's just a tox feature (here [the docs](https://tox.wiki/en/latest/example/basic.html#depending-on-requirements-txt-or-defining-constraints), which don't say much about the missing whitespace).

tox.ini
ERafaelMartinez: But what would then happen if you run `INTEGRATION_TESTS_REPO=/Some/path/ tox -e integrationtests`?
Wouldn't INTEGRATION_TESTS_REPO be overriden inside the testenv with `git@github.com:gridsingularity/gsy-backend-integration-tests.git`, and therefore fail to clone locally?

ERafaelMartinez:
Adapted to comments, appreciate your re-approval. Thanks!

tox.ini
fievelk: You're right, I was assuming that `setenv` would have been overridden by the value passed by the CLI command, but I was wrong :P I suppose there must be another way, but I couldn't figure it out yet. We can leave it as it is!

tox.ini
```diff
@@ -38,16 +38,16 @@ setenv =
...
 allowlist_externals =
     git
-    /bin/cp
+    /bin/ln
     /bin/rm
 commands_pre =
     pip install -e .
     git clone {env:INTEGRATION_TESTS_REPO:git@github.com:gridsingularity/gsy-backend-integration-tests.git} {envtmpdir}/gsy-backend-integration-tests
-    cp -r {envtmpdir}/gsy-backend-integration-tests/gsy_e_sdk_tests/integration_tests/ ./integration_tests/
+    ln -sf {envtmpdir}/gsy-backend-integration-tests/gsy_e_sdk_tests/integration_tests/ ./integration_tests
```
hannesdiedrich, line 46: I do not understand why softlinks are better here. But if it works, LGTM

hannesdiedrich (approved):
LGTM

tox.ini
hannesdiedrich: Please explain the reason here, thanks! (could not find the conversation about it)

spyrostz: That was my suggestion, and it was admittedly minor. It is in order to to avoid the time that it will take to copy the files In this case the time is negligible, however it is better as a practice to not copy files if one can avoid it and use symbolic links instead.

hannesdiedrich: Ah ok, makes sense. Thanks!

spyrostz (approved):

PR #209, opened 2022-02-23
GSYE-204: Rename the area_id constructor argument of the RedisAssetClient to asset_uuid

Hassan754:


gsy_e_sdk/clients/redis_asset_client.py
```diff
@@ -8,3 +9,7 @@ class RedisAssetClient(RedisClientBase):
     the same approach as in the REST case, where we have two different classes for devices and
     markets.
     """
+
+    def __init__(self, asset_uuid, autoregister=True, redis_url=LOCAL_REDIS_URL,
```
Hassan754, line 13: The parent constructor was expecting the `area_id` argument whereas the `RestAssetClient` expects an asset_uuid, so in order to homogenize the interface of both RestAssetClient and RedisAssetClient, I have override the constructor here

hannesdiedrich: Hmm, but the `area_id` is not the same as `area_uuid`. The `area_id` is the name of the area because we can not know the area_uuid in the redis case. If you look into the `RedisClientBase`, the area_uuid is set only when after registering to the simulation.

hannesdiedrich (commented):
Left an important comment. We need to discuss this or just leave it and just know that the area_id is not an area_uuid

gsy_e_sdk/clients/redis_asset_client.py
Hassan754: Oops, I have forgot about this although I stumbled upon it several times in the past, I may have said this before but again, we need to rename this variable as it's misleading.
Regardless, one thing that we need to agree with is whether we want both the `RestAssetClient` and `RedisAssetClient`'s instantiation's interface to be homogenized or not

PR #214, opened 2022-04-04
TESTS-143: I want to reduce tox files redundancy on tox file testenvs.

ERafaelMartinez:

Modularization of the testenvs in tox file was improved for gsy-web repository. The changes on this repo will match up with that done on gsy-web.

## Proposed changes

Improve the modularization of the tox configuration environments.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,10 +4,7 @@ envlist = check_readme, covclean, py38, coverage, integrationtests
 
 [testenv]
 basepython = python3.8
-passenv = LANG TERM LANGUAGE LC_ALL LD_LIBRARY_PATH BRANCH INTEGRATION_TESTS_REPO INTEGRATION_TESTS_BRANCH
-deps =
-    pip-tools
-    coverage
+passenv = LANG TERM LANGUAGE LC_ALL LD_LIBRARY_PATH BRANCH
 commands_pre =
     pip install --upgrade pip
     pip install -rrequirements/base.txt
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -63,25 +62,15 @@ deps = readme_renderer
...
-    pip install -e .
-    pip install -r requirements/dev.txt
-    git clone -b {env:INTEGRATION_TESTS_BRANCH:master} {env:INTEGRATION_TESTS_REPO:git@github.com:gridsingularity/gsy-backend-integration-tests.git} {envtmpdir}/gsy-backend-integration-tests
-    ln -sf {envtmpdir}/gsy-backend-integration-tests/gsy_e_sdk_tests/integration_tests/ ./integration_tests
+    {[testenv:integrationtests]commands_pre}
 commands =
-    pytest unit_tests
-    behave ./integration_tests/ --stop
+    {[testenv:unittests]commands}
+    {[testenv:integrationtests]commands} --stop
```
fievelk, line 74: Are you sure this works? What happens if we add a new line to the `testenv:integrationtests` commands?

fievelk (commented):
Left one question, will review again.

tox.ini
ERafaelMartinez: Yes! I tested it locally before pushing and also the ci checks passed. 
In [this line](https://github.com/gridsingularity/gsy-e-sdk/runs/5810656597?check_suite_focus=true#step:8:399) you can see how it looks like. :)

fievelk: That's nice :) But I'm a bit afraid that this will break if we add a new line in `testenv:integrationtests`, because at that point the `--stop` argument will be appended to the wrong command.

ERafaelMartinez: If the theoretical new command is added before the actual execution of `behave`, this should not break. But I do get your point. Adapted on last commit.

fievelk (approved):

PR #215, opened 2022-04-05
TESTS-141: As gsy-backend-integration-tests, I want to build a docker image with the latest changes of the working branch

ERafaelMartinez:
## Reason for proposed changes

- The migration of the integration tests to the current repository did not consider the need of creating a `gsy-e` image based on a particular branch when executing tests on gsy-myco-sdk and/or gsy-e-sdk, and therefore after the creation of a PR, the right image would not be built and tests would fail.

## Proposed changes

- Adapt PR template message to include the gsy-e TARGET_BRANCH that will be considered to build the docker image used by the tests.
- Adapt github actions to parse variable from PR body and pass it to the testenvs.

**INTEGRATION_TESTS_BRANCH**=TESTS-141

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -8,3 +8,5 @@ Please describe what we want to achieve and why.
 
 <!--- If needed, choose which branch of the gsy-backend-integration-tests repository will be used to run integration tests. Please only edit the name of the branch. -->
 **INTEGRATION_TESTS_BRANCH**=master
+<!--- If needed, choose which gsy-e branch should be used to build docker image to execute integration tests.-->
+**TARGET_BRANCH**=master
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -9,4 +9,4 @@ Please describe what we want to achieve and why.
 <!--- If needed, choose which branch of the gsy-backend-integration-tests repository will be used to run integration tests. Please only edit the name of the branch. -->
 **INTEGRATION_TESTS_BRANCH**=master
 <!--- If needed, choose which gsy-e branch should be used to build docker image to execute integration tests.-->
-**TARGET_BRANCH**=master
+**GSY_E_TARGET_BRANCH**=master
--- a/.github/workflows/development.yaml
+++ b/.github/workflows/development.yaml
@@ -29,8 +29,8 @@
         env:
           prBody: ${{ steps.PR.outputs.pr_body }}
         run: |
-          echo "::set-output name=PARSED_TARGET_BRANCH::$(echo -e $prBody | sed -n 's/.*\*\*TARGET_BRANCH\*\*=\([^ ]*\).*/\1/p')"
-        id: parse_target_branch
+          echo "::set-output name=PARSED_GSY_E_TARGET_BRANCH::$(echo -e $prBody | sed -n 's/.*\*\*GSY_E_TARGET_BRANCH\*\*=\([^ ]*\).*/\1/p')"
+        id: parse_gsye_target_branch
 
       - name: Validate parsed integration tests branch
         env:
@@ -39,12 +39,12 @@
           echo "::set-output name=INTEGRATION_TESTS_BRANCH::${PARSED_INTEGRATION_TESTS_BRANCH:-'master'}"
         id: validated_integration_tests_branch
 
-      - name: Validate parsed target branch
+      - name: Validate parsed gsy-e target branch
         env:
-          PARSED_TARGET_BRANCH: ${{ steps.parse_target_branch.outputs.PARSED_TARGET_BRANCH }}
+          PARSED_GSY_E_TARGET_BRANCH: ${{ steps.parse_gsye_target_branch.outputs.PARSED_GSY_E_TARGET_BRANCH }}
         run: |
-          echo "::set-output name=TARGET_BRANCH::${PARSED_TARGET_BRANCH:-'master'}"
-        id: validated_target_branch
+          echo "::set-output name=GSY_E_TARGET_BRANCH::${PARSED_GSY_E_TARGET_BRANCH:-'master'}"
+        id: validated_gsye_target_branch
 
       - name: Setup Python
         uses: actions/setup-python@v1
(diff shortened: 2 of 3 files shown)


.github/pull_request_template.md
```diff
@@ -8,3 +8,5 @@ Please describe what we want to achieve and why.
 
 <!--- If needed, choose which branch of the gsy-backend-integration-tests repository will be used to run integration tests. Please only edit the name of the branch. -->
 **INTEGRATION_TESTS_BRANCH**=master
+<!--- If needed, choose which gsy-e branch should be used to build docker image to execute integration tests.-->
+**TARGET_BRANCH**=master
```
hannesdiedrich, line 12: I vote for renaming this to `GSY-E-TARGET-BRANCH`, because I was first confused which target is meant here.

fievelk: I was thinking the same but I wasn't sure whether to raise the question. Since Hannes had the same feeling, I'd also vote for the renaming he suggests.

.github/workflows/development.yaml
```diff
@@ -23,14 +23,28 @@
           prBody: ${{ steps.PR.outputs.pr_body }}
         run: |
           echo "::set-output name=PARSED_INTEGRATION_TESTS_BRANCH::$(echo -e $prBody | sed -n 's/.*\*\*INTEGRATION_TESTS_BRANCH\*\*=\([^ ]*\).*/\1/p')"
-        id: parse_branch
+        id: parse_int_tests_branch
```
fievelk, line 26:
```suggestion
        id: parse_integration_tests_branch
```

.github/pull_request_template.md
ERafaelMartinez: Ok, I'm also in favor of this. 
The only thing we should keep in mind is that this would increase the total amount of variables we use.

This `TARGET_BRANCH` and `TARGET_REPO` variables are also used in context of GBIT. For instance, if you want to execute `gsy-web` tests from GBIT but you want to point it to a particular branch/repo, you would do something like:

`$ TARGET_BRANCH=sample/branch TARGET_REPO=/Some/Path/gsy-web tox -e test-gsy-web`

The same variables are used here because it serves a very similar purpose. However I do agree that the name is not clear enough.

ERafaelMartinez: Will adapt after conversation is resolved.

hannesdiedrich: I see now. As this is a feature only for the local runs, I propose to rename oll of them :
for the gsy-web --> `GSY_WEB_TARGET_REPO`
for the gsy-e --> `GSY_E_TARGET_REPO`
No need to expose these in the PR template IMO.

ERafaelMartinez: In this case, we should also rename then:
for the gsy-e-sdk --> GSY_E_SDK_TARGET_REPO
for the gsy-myco-sdk --> GSY_MYCO_SDK_TARGET_REPO

Are we sure we're ok with this although this increases the number of variables from 1 to 4? An upside would be though, that the name of the variable reveals more clearly the use of it.

Also, they are not exposed in the PR template. The only one which can't get away from this is then GSY_E_TARGET_REPO because it is needed to build the docker image also in CI, not only when local testing. This variable must be exposed.

hannesdiedrich: I think we are spinning in circles here and I blame myself for not being up to date here. Sorry.
This is how I understand it:
The `*_TARGET_REPO` values we only need when we run the integration test of any repo locally, right? If this is not set, the url if the tested repo is used, right?
The `*_TARGET_BRANCH` values are used to set the correct branch for the GSY-E that is used to build a test image to test the connection between GSY-E and SDK and MYCO-SDK, right?
For the other repos we do not need any external branches, we use the `INTEGRATION_TESTS_BRANCH` to select the updated integration tests, right?

ERafaelMartinez: There are two main cases in which this variables are used: when tests ar run from GBIT and (added in this PR) when tests are run either from GSY-E-SDK or GSY-MYCO-SDK.

> The *_TARGET_REPO values we only need when we run the integration test of any repo locally, right? If this is not set, the url if the tested repo is used, right?

Correct. You are talking about the case in which the tests are executed from GBIT.

> The *_TARGET_BRANCH values are used to set the correct branch for the GSY-E that is used to build a test image to test the connection between GSY-E-SDK and GSY-MYCO-SDK, right?

Jein, initially TARGET_BRANCH alone was used together with TARGET_REPO to select the branch of  the TARGET_REPO which should be used when running the tests on GBIT (for this reason what you mentioned is false)

In the context of this PR, TARGET_BRANCH is used exclusively to point to a particular GSYE branch (for this reason what you mentioned is true). 

However, when TARGET_BRANCH is used in GSYE_SDK and in GSY_MYCO_SDK (and not from GBIT as in above case), TARGET_BRANCH exclusively points to a GSYE branch, and TARGET_REPO is exclusively used to point to a local GSYE repo. This is why it make sense to rename it in this case to GSYE_TARGET_BRANCH and GSYE_TARGET_REPO, because for case 2, this is the only use this variables have. This implies the addition of these 2 new variables though, as for the scenario in which the same variables are used (TARGET_BRANCH, TARGET_REPO), we could "reuse" the already defined variables used on the GBIT workflow and therefore avoid managing 2 more variables.

> For the other repos we do not need any external branches, we use the INTEGRATION_TESTS_BRANCH to select the updated integration tests, right?

Yes, to select the updated tests we do as you mention. However, after this PR, we would also need to specify TARGET_BRANCH for GSYE_SDK and in GSY_MYCO_SDK repositories, and therefore expose it in the PR message. (Or GSYE_TARGET_BRANCH if it is agreed to rename it).

hannesdiedrich: Ok, so I think we agree to 
1. rename `TARGET_BRANCH` into `GSY-E-TARGET_BRANCH` on GSY-E-SDK, GSYE-MYCO_SDK and GBIT.
2. rename `TARGET_REPO` into `GSY-E-TARGET_REPO` on GSY-E-SDK, GSYE-MYCO_SDK and GBIT (but not expose it on the PR description template)
Right?

ERafaelMartinez: I think so, would just adapt it slightly to:

1. rename TARGET_BRANCH into GSY-E-TARGET_BRANCH on GSY-E-SDK, GSYE-MYCO_SDK ~and GBIT~.
2. rename TARGET_REPO into GSY-E-TARGET_REPO on GSY-E-SDK, GSYE-MYCO_SDK ~and GBIT~ (but not expose it on the PR description template)

for GBIT to use TARGET_BRANCH and TARGET_REPO regardless of the repo called.

hannesdiedrich (approved):
LGTM

spyrostz (approved):
LGTM

PR #219, opened 2022-05-02
GSYE-274: Make --setup a required CLI option with no default value.

mrsaemir:
As described in [GSYE-274](https://gridsingularity.atlassian.net/browse/GSYE-274), we need to force the user to enter a valid setup name.

## Proposed changes
- Make --setup option required.


fievelk (approved):
LGTM. 

Nit: Not needed, but if you're going to add other commits I think we could also change [this line](https://github.com/gridsingularity/gsy-e-sdk/blob/ed3bd6655a336a657a0ef59f40678dad6ede1487/gsy_e_sdk/cli.py#L124) to use `gsy_e_sdk`, since the `d3a_api_client` name is deprecated.

spyrostz (approved):
LGTM

PR #221, opened 2022-05-19
Asset api bid offer posting

andreabertolini1995:

The need to implement simple example setups that only contain the bare minimum logic, in order to provide users with very straightforward examples and, at the same time, to create a sort of template that can be used to share bugs with the backend team.

## Proposed changes

Proposed two basic scripts that show how a consumption, generation and storage asset would post a bid and/or offer to the Spot market via Redis and Rest connection, without following any trading strategy.The scripts have been adapted to the best practices of backed code by checking structure and logic using both Black and Pylint.


gsy_e_sdk/setups/Asset_API_scripts/rest_bid_offer_posting.py
```diff
@@ -0,0 +1,158 @@
...
+    for key, value in indict.items():
+        if key == "name":
+            name = value
+        if key == "type":
+            area_type = value
+        if key == "registered" and value:
+            outdict[area_type].append(name)
+        if "children" in key:
+            for children in indict[key]:
+                _get_assets_name(children, outdict)
```
fievelk, line 112: As far as I understand, this method takes the name of the area and, if it's registered, adds it to `outdict` (which is a dict of all the registered assets).

I know this implementation already existed, but I think it would be good to refactor it because the way this works now is a bit obscure. Would this be the same?
```suggestion
    def _get_assets_name(indict: dict, outdict: dict):
        if indict.get("registered") is True:
            area_type = indict["type"]
            outdict[area_type].append(indict["name"])

        for child in indict.get("children", []):
            _get_assets_name(child, outdict)
```
I gave for granted that we always have `area_type` inside.

Do you think this is equivalent to the other version? Tagging @hannesdiedrich to also ask for his opinion.

(I would also suggest to rename `indict` to `node` or `area` and `outdict` to `registered_assets`)

gsy_e_sdk/setups/Asset_API_scripts/rest_bid_offer_posting.py
```diff
@@ -0,0 +1,158 @@
...
+        asset.select_aggregator(aggr.aggregator_uuid)
+    return asset_uuid_map
+
+
+print()
+print("Registering assets ...")
+asset_uuid_mapping = {}
+asset_uuid_mapping = register_asset_list(load_names, asset_args, asset_uuid_mapping)
+asset_uuid_mapping = register_asset_list(pv_names, asset_args, asset_uuid_mapping)
+asset_uuid_mapping = register_asset_list(storage_names, asset_args, asset_uuid_mapping)
```
fievelk, line 148: It looks like here we're registering all assets (also PVs and storages), so this module is behaving differently than the redis one. I suppose we want to make them behave similarly, right?

gsy_e_sdk/setups/Asset_API_scripts/rest_bid_offer_posting.py
```diff
@@ -0,0 +1,147 @@
...
+            self.execute_batch_commands()
+
+    def on_event_or_response(self, message):
+        pass
+
+    def on_finish(self, finish_info):
+        self.is_finished = True
+
+
+def get_assets_name(node: dict) -> dict:
```
hannesdiedrich, line 81: NIT: Here in the wrapper I propose to rename `node` into `registry` or `representation`. What do you think?

gsy_e_sdk/setups/Asset_API_scripts/rest_bid_offer_posting.py
```diff
@@ -0,0 +1,147 @@
...
+        asset.select_aggregator(aggregator.aggregator_uuid)
+    return asset_uuid_map
+
+
+print()
+print("Registering assets ...")
+asset_uuid_mapping = {}
+asset_uuid_mapping = register_asset_list(LOAD_NAMES, asset_args, asset_uuid_mapping)
+asset_uuid_mapping = register_asset_list(PV_NAMES, asset_args, asset_uuid_mapping)
+asset_uuid_mapping = register_asset_list(STORAGE_NAMES, asset_args, asset_uuid_mapping)
```
hannesdiedrich, line 137: How about:
```suggestion
asset_uuid_mapping = register_asset_list(LOAD_NAMES+PV_NAMES+STORAGE_NAMES, asset_args, asset_uuid_mapping)
```

gsy_e_sdk/setups/Asset_API_scripts/rest_bid_offer_posting.py
```diff
@@ -0,0 +1,147 @@
...
+    wrapper for _get_assets_name
+    """
+    if node == {}:
+        return {}
+    reg_assets = {"Area": [], "Load": [], "PV": [], "Storage": []}
+    _get_assets_name(node, reg_assets)
+    return reg_assets
+
+
+def _get_assets_name(node: dict, reg_assets: dict):
```
hannesdiedrich, line 93: We have a soft rule to use the type hints from the `typing` module (for Lists, Dictionaries, Sets:
```python
from typing import List, Dict
```
```suggestion
def _get_assets_name(node: Dict, reg_assets: Dict):
```

gsy_e_sdk/setups/Asset_API_scripts/rest_bid_offer_posting.py
```diff
@@ -0,0 +1,147 @@
...
+asset_args = {"autoregister": False, "start_websocket": False}
+if AUTOMATIC:
+    registry = aggregator.get_configuration_registry()
+    registered_assets = get_assets_name(registry)
+    LOAD_NAMES = registered_assets["Load"]
+    PV_NAMES = registered_assets["PV"]
+    STORAGE_NAMES = registered_assets["Storage"]
+
+
+def register_asset_list(asset_names, asset_params, asset_uuid_map):
```
hannesdiedrich, line 118: Please also add type hints here:
```suggestion
def register_asset_list(asset_names: List, asset_params: Dict, asset_uuid_map: Dict) -> Dict:
```

gsy_e_sdk/setups/Asset_API_scripts/rest_bid_offer_posting.py
```diff
@@ -0,0 +1,147 @@
...
+    for child in node.get("children", []):
+        _get_assets_name(child, reg_assets)
+
+
+aggregator = Oracle(aggregator_name=ORACLE_NAME)
+simulation_id = os.environ["API_CLIENT_SIMULATION_ID"]
+domain_name = os.environ["API_CLIENT_DOMAIN_NAME"]
+websockets_domain_name = os.environ["API_CLIENT_WEBSOCKET_DOMAIN_NAME"]
+asset_args = {"autoregister": False, "start_websocket": False}
+if AUTOMATIC:
```
hannesdiedrich, line 110: If `AUTOMATIC` is set to False, the code further down will fail, AFAIS. Can you please explain why you need this flag? Please also add an else block if it is needed. Thanks!

gsy_e_sdk/setups/Asset_API_scripts/rest_bid_offer_posting.py
```diff
@@ -0,0 +1,147 @@
...
+from gsy_e_sdk.clients.rest_asset_client import RestAssetClient
+from gsy_e_sdk.utils import get_area_uuid_from_area_name_and_collaboration_id
+
+current_dir = os.path.dirname(__file__)
+ORACLE_NAME = "oracle"
+
+LOAD_NAMES = ["Load 1 L13", "Load 2 L21", "Load 3 L17"]
+PV_NAMES = ["PV 1 (4kW)", "PV 3 (5kW)"]
+STORAGE_NAMES = ["Tesla Powerwall 3"]
+AUTOMATIC = True
```
hannesdiedrich, line 19: How about naming it into the following or similar? This would make it more self explanatory. 
```suggestion
CONNECT_TO_ALL_ASSETS = True
```

gsy_e_sdk/setups/Asset_API_scripts/rest_bid_offer_posting.py
```diff
@@ -0,0 +1,147 @@
+# flake8: noqa
```
hannesdiedrich, line 1: Why are you disabling flake8?

hannesdiedrich (commented):
A few minor comments, looks good in general.

gsy_e_sdk/setups/Asset_API_scripts/rest_bid_offer_posting.py
hannesdiedrich: What do you think about my proposal? IMO AUTOMATIC could mean anything. If you want to keep the naming, please add an inline comment to explain what it means, thanks!

gsy_e_sdk/setups/Asset_API_scripts/rest_bid_offer_posting.py
```diff
@@ -0,0 +1,147 @@
...
+Template file for a trading strategy through the gsy-e-sdk api client using Rest.
+"""
+
+import os
+from time import sleep
+from gsy_e_sdk.aggregator import Aggregator
+from gsy_e_sdk.clients.rest_asset_client import RestAssetClient
+from gsy_e_sdk.utils import get_area_uuid_from_area_name_and_collaboration_id
+
+current_dir = os.path.dirname(__file__)
```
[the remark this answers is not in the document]
hannesdiedrich: Please address.

gsy_e_sdk/setups/Asset_API_scripts/rest_bid_offer_posting.py
hannesdiedrich: Please answer.

gsy_e_sdk/setups/Asset_API_scripts/rest_bid_offer_posting.py
hannesdiedrich: Please reply or adapt, thanks!

gsy_e_sdk/setups/Asset_API_scripts/rest_bid_offer_posting.py
hannesdiedrich: Here I am also curious about your opinion.

gsy_e_sdk/setups/Asset_API_scripts/rest_bid_offer_posting.py
andreabertolini1995: Flake8 was disabled to disable the format test, now this line can be deleted.

hannesdiedrich (approved):
LGTM

gsy_e_sdk/utils.py
```diff
@@ -337,3 +337,27 @@ def get_name_from_area_name_uuid_mapping(area_name_uuid_mapping, asset_uuid):
...
     return None
+
+
+def get_assets_name(registry: Dict) -> Dict:
+    """
+    Parse the grid tree and return all registered assets / markets
+    Wrapper for _get_assets_name
+    """
+    if registry == {}:
+        return {}
```
fievelk, line 348: The registry returned by `get_configuration_registry()` can also be `None` in case of errors. I'd suggest replacing this with
```suggestion
    if not registry:
        return {}
```

fievelk (approved):
One comment that would be good to address. I will still give my LGTM since it shouldn't be too problematic most of the time.

gsy_e_sdk/setups/Asset_API_scripts/rest_bid_offer_posting.py
```diff
@@ -0,0 +1,122 @@
...
+simulation_id = os.environ["API_CLIENT_SIMULATION_ID"]
+domain_name = os.environ["API_CLIENT_DOMAIN_NAME"]
+websockets_domain_name = os.environ["API_CLIENT_WEBSOCKET_DOMAIN_NAME"]
+asset_args = {"autoregister": False, "start_websocket": False}
+if CONNECT_TO_ALL_ASSETS:
+    registry = aggregator.get_configuration_registry()
+    registered_assets = get_assets_name(registry)
+    LOAD_NAMES = registered_assets["Load"]
+    PV_NAMES = registered_assets["PV"]
+    STORAGE_NAMES = registered_assets["Storage"]
```
fievelk, line 93: If `get_assets_name` returns `{}`, these lines will fail with a `KeyError`. I suppose that if `registered_assets` is empty/None, we might want to print a message and probably stop the execution, right? OR at least take some kind of measure to address the problem.

gsy_e_sdk/setups/Asset_API_scripts/rest_bid_offer_posting.py
```diff
@@ -88,9 +88,12 @@ def on_finish(self, finish_info):
...
     registered_assets = get_assets_name(registry)
-    LOAD_NAMES = registered_assets["Load"]
-    PV_NAMES = registered_assets["PV"]
-    STORAGE_NAMES = registered_assets["Storage"]
+    LOAD_NAMES = registered_assets.get("Load", [])
+    PV_NAMES = registered_assets.get("PV", [])
+    STORAGE_NAMES = registered_assets.get("Storage", [])
+    assert len(LOAD_NAMES + PV_NAMES + STORAGE_NAMES) != 0, (
+        "No assets are registered to the aggregator"
+    )
```
hannesdiedrich, line 96: This is not needed for the `register_asset_list` but only for the convenience of the user, right?

hannesdiedrich (approved):
LGTM

gsy_e_sdk/setups/Asset_API_scripts/rest_bid_offer_posting.py
ERafaelMartinez: Right! In particular, to deal with @fievelk 's concern expressed [here](https://github.com/gridsingularity/gsy-e-sdk/pull/221#discussion_r899879316).

fievelk (approved):

gsy_e_sdk/setups/Asset_API_scripts/rest_bid_offer_posting.py
ERafaelMartinez: Oh, yes, you're right. The main concert is dealt with the `.get()` method. I introduced the assertion because I thought it would be no use to connect to a simulation with no assets on such lists. But, as you say, an Exception would be best. Do you have any in mind?

PR #222, opened 2022-05-19
Refactoring asset api basic strategies script

andreabertolini1995:

## Proposed changes

Proposed two scripts that show how a consumption, generation and stoaage asset would post bids and offers following a basic trading strategy via Redis and Rest connection. The core the scripts has been taken from the Asset_API_template script, adopting best practices used for backend code (used both Black and Pylint to check the code) and generally improved script's structure and logic.

### DIFF
--- a/gsy_e_sdk/setups/Asset_API_scripts/redis_basic_strategies.py
+++ b/gsy_e_sdk/setups/Asset_API_scripts/redis_basic_strategies.py
@@ -1,25 +1,25 @@
 # flake8: noqa
 """
-Template file for a trading strategy through the gsy-e-sdk api client
+Template file for a trading strategy through the gsy-e-sdk api client using Redis.
 """
 
 import os
 from time import sleep
 from gsy_framework.utils import key_in_dict_and_not_none_and_greater_than_zero
-from gsy_e_sdk.types import aggregator_client_type
+from gsy_e_sdk.redis_aggregator import RedisAggregator
 from gsy_e_sdk.clients.redis_asset_client import RedisAssetClient
 
 current_dir = os.path.dirname(__file__)
 ORACLE_NAME = "oracle"
 
-load_names = ["Load 1 L13", "Load 2 L21", "Load 3 L17"]
-pv_names = ["PV 1 (4kW)", "PV 3 (5kW)"]
-storage_names = ["Tesla Powerwall 3"]
+LOAD_NAMES = ["Load 1 L13", "Load 2 L21", "Load 3 L17"]
+PV_NAMES = ["PV 1 (4kW)", "PV 3 (5kW)"]
+STORAGE_NAMES = ["Tesla Powerwall 3"]
 
 TICKS = 10  # leave as is
 
 
-class Oracle(aggregator_client_type):
+class Oracle(RedisAggregator):
     """Class that defines the behaviour of an "oracle" aggregator."""
 
     def __init__(self, *args, **kwargs):
@@ -223,9 +223,8 @@ class Oracle(aggregator_client_type):
         self.is_finished = True
 
 
-aggr = Oracle(aggregator_name=ORACLE_NAME)
-AssetClient = RedisAssetClient
-asset_args = {"autoregister": True, "pubsub_thread": aggr.pubsub}
+aggregator = Oracle(aggregator_name=ORACLE_NAME)
+asset_args = {"autoregister": True, "pubsub_thread": aggregator.pubsub}
 
 
 def register_asset_list(asset_names, asset_params, asset_uuid_map):
@@ -233,21 +232,24 @@ def register_asset_list(asset_names, asset_params, asset_uuid_map):
     for asset_name in asset_names:
         print("Registered asset:", asset_name)
         asset_params["area_id"] = asset_name
-        asset = AssetClient(**asset_params)
+        asset = RedisAssetClient(**asset_params)
         asset_uuid_map[asset.area_uuid] = asset.area_id
-        asset.select_aggregator(aggr.aggregator_uuid)
+        asset.select_aggregator(aggregator.aggregator_uuid)
     return asset_uuid_map
 
 
 print()
 print("Registering assets ...")
 asset_uuid_mapping = {}
-asset_uuid_mapping = register_asset_list(load_names, asset_args, asset_uuid_mapping)
+asset_uuid_mapping = register_asset_list(LOAD_NAMES, asset_args, asset_uuid_mapping)
+asset_uuid_mapping = register_asset_list(PV_NAMES, asset_args, asset_uuid_mapping)
+asset_uuid_mapping = register_asset_list(STORAGE_NAMES, asset_args, asset_uuid_mapping)
+
 print()
 print("Summary of assets registered:")
 print()
 print(asset_uuid_mapping)
 
 # loop to allow persistence
-while not aggr.is_finished:
+while not aggregator.is_finished:
     sleep(0.5)
--- a/gsy_e_sdk/setups/Asset_API_scripts/redis_basic_strategies.py
+++ b/gsy_e_sdk/setups/Asset_API_scripts/redis_basic_strategies.py
@@ -5,7 +5,6 @@ Template file for a trading strategy through the gsy-e-sdk api client using Redi
 
 import os
 from time import sleep
-from gsy_framework.utils import key_in_dict_and_not_none_and_greater_than_zero
 from gsy_e_sdk.redis_aggregator import RedisAggregator
 from gsy_e_sdk.clients.redis_asset_client import RedisAssetClient
 
@@ -33,13 +32,11 @@ class Oracle(RedisAggregator):
             return
         self.build_strategies(market_info)
         self.post_bid_offer()
-        self.execute_batch_commands()
 
     def on_tick(self, tick_info):
         """Place a bid or an offer each 10% of the market slot progression."""
         i = int(float(tick_info["slot_completion"].strip("%")) / TICKS)
         self.post_bid_offer(i)
-        self.execute_batch_commands()
 
     def build_strategies(self, market_info):
         """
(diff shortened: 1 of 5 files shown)


gsy_e_sdk/setups/Asset_API_scripts/redis_basic_strategies.py
```diff
@@ -0,0 +1,250 @@
...
+                                            * self.asset_strategy[area_uuid][
+                                                "fee_to_market_maker"
+                                            ]
+                                            - fit_rate
+                                    )
+                                    * (i / TICKS),
+                                ),
+                                3,
+                            )
+                        )
```
hannesdiedrich, line 119: Of course, this is personal taste and if you think that the users prefer this way of indention, please ignore me. However, I think it is not readable and propose the following where you separate the calculation of the rate from the rounding and max():
```suggestion
    sell_rate = market_maker_rate + self.asset_strategy[area_uuid]["fee_to_market_maker"] - (
            market_maker_rate + 2 * self.asset_strategy[area_uuid]["fee_to_market_maker"] - fit_rate) * (i / TICKS)

    gen_strategy.append(round(max(0, sell_rate)), 3))
```
Can you please explain why you round the values?

hannesdiedrich: Sorry, the indention in the diff is not pretty.
Have a look here:
Screenshot 2022-06-03 at 14 20 28

gsy_e_sdk/setups/Asset_API_scripts/redis_basic_strategies.py
```diff
@@ -0,0 +1,250 @@
...
+                                + self.asset_strategy[area_uuid]["fee_to_market_maker"],
+                                3,
+                            )
+                        )
+                self.asset_strategy[area_uuid]["buy_rates"] = load_strategy
+
+            # Generation strategy
+            if "available_energy_kWh" in area_dict["asset_info"]:
+                gen_strategy = []
+                for i in range(0, TICKS):
```
hannesdiedrich, line 97: single character variables are discouraged (https://peps.python.org/pep-0008/#names-to-avoid)
```suggestion
                for tick in range(0, TICKS):
```

gsy_e_sdk/setups/Asset_API_scripts/rest_basic_strategies.py
```diff
@@ -0,0 +1,290 @@
...
+                            )
+                            * (i / TICKS),
+                            3,
+                        )
+                    )
+
+                self.asset_strategy[area_uuid]["buy_rates"] = batt_buy_strategy
+                self.asset_strategy[area_uuid]["sell_rates"] = batt_sell_strategy
+
+    def post_bid_offer(self, i=0):
```
hannesdiedrich, line 178: here too, please rename `i` into maybe `rate_index` (if I understand correctly)

gsy_e_sdk/setups/Asset_API_scripts/rest_basic_strategies.py
```diff
@@ -0,0 +1,290 @@
...
+            self.execute_batch_commands()
+
+    def on_event_or_response(self, message):
+        pass
+
+    def on_finish(self, finish_info):
+        self.is_finished = True
+
+
+def get_assets_name(node: dict) -> dict:
```
hannesdiedrich, line 225: As this is used also in other scripts, I suggest to put this into a `helper` module and import it in order to reduce code duplication.

hannesdiedrich (commented):
I left some comments. Please adapt and I will review again.
Also I have a general comment: Please rename the asset api package from `Asset_API_scripts` into `asset_api_scripts` in order to follow [PEP8](https://peps.python.org/pep-0008/#package-and-module-names) Thanks!

gsy_e_sdk/setups/Asset_API_scripts/redis_basic_strategies.py
andreabertolini1995: The indentation I have proposed is the one Black suggests. But I agree, it is not readable and I will change it. 

Regarding the round function, I investigated, and it had been initially added just to avoid extremely long numbers in the terminal. I believe we can remove the round function and avoid this 'esthetic' issue somewhere else in the backend? What do you think?

hannesdiedrich: > the one Black suggests 
Who is "Black"?  

Regarding the rounding: If you are ok with sending the rounded values to the exchange, all good. I thought that this rounding could mask one of the bidding strategies' behaviour.

gsy_e_sdk/setups/Asset_API_scripts/redis_basic_strategies.py
hannesdiedrich: It applies to all variables. The goal is that the reader of the code should understand it without comments or docstrings.

gsy_e_sdk/setups/Asset_API_scripts/redis_basic_strategies.py
andreabertolini1995: Black is a formatting tool: https://pypi.org/project/black/.

hannesdiedrich: I guess this proves that Black is not useful :-)

gsy_e_sdk/setups/Asset_API_scripts/redis_basic_strategies.py
```diff
@@ -0,0 +1,250 @@
...
+from gsy_e_sdk.clients.redis_asset_client import RedisAssetClient
+
+current_dir = os.path.dirname(__file__)
+ORACLE_NAME = "oracle"
+
+LOAD_NAMES = ["Load 1 L13", "Load 2 L21", "Load 3 L17"]
+PV_NAMES = ["PV 1 (4kW)", "PV 3 (5kW)"]
+STORAGE_NAMES = ["Tesla Powerwall 3"]
+
+TICKS = 10  # leave as is
```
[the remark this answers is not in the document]
hannesdiedrich: This hs not been adapted.

gsy_e_sdk/setups/asset_api_scripts/redis_basic_strategies.py
```diff
@@ -0,0 +1,196 @@
...
+                                 self.asset_strategy[area_uuid]["fee_to_market_maker"] -
+                                 (market_maker_rate +
+                                  self.asset_strategy[area_uuid]["fee_to_market_maker"] -
+                                  med_price) * (tick / TICKS)
+                                 )
+                    batt_sell_strategy.append(sell_rate)
+                self.asset_strategy[area_uuid]["buy_rates"] = batt_buy_strategy
+                self.asset_strategy[area_uuid]["sell_rates"] = batt_sell_strategy
+
+    def post_bid_offer(self, rate_index=0):
```
hannesdiedrich, line 122: Sorry, I just see this now. The `rate_index` is always 0. So only the first value of the `self.asset_strategy` is used. Is this wanted? Why are you constructing the `self.asset_strategy` with 10 members in the first place?
Also, your explanation for the `TICKS` constant does not conform to this behaviour. Or did I miss something?

hannesdiedrich: Answer from Andrea:

> Actually, rate_index = 0 only when I call the post_bid_offer function without specifying any argument, as in the on_market_cycle method. However, every time on_tick is triggered, I call the post_bid_offer function specifying the rate_index, overwriting the default value of 0. Thus, all the 10 members of the strategy are actually used.

hannesdiedrich: I see now. Thanks for pointing me to it. 
Now I understand fully what you are trying here. Unfortunately, this implementation only works, if TICK=10. 
If we change the frequency of the dispatching of the `tick_info` in the backend from every 10% to e.g. every 5 %, this script will fail, because e.g. 80/5 = 16 (which wich is an index that will not part of your `asset_strategy` dictionary then. 
Also, if the user decides to post offers more often then every 10% of the slot, he/she can not simple change the value. This is probably why you added `to leave as it is` to the comment, right? 
So, as this is a backend-dependant setting, I would vote for renaming it to: `TICK_DISPATCH_FREQUENCY_PERCENT` or similar.
Sorry to be relentless here, but the point of these refactorings was also to be better debuggable by the backend team.

gsy_e_sdk/setups/asset_api_scripts/redis_basic_strategies.py
```diff
@@ -0,0 +1,198 @@
...
+                area_uuid,
+                self.get_uuid_from_area_name("Market Maker"),
+                "current_market_fee",
+            )
+
+            # Consumption strategy
+            if "energy_requirement_kWh" in area_dict["asset_info"]:
+                load_strategy = []
+                for tick in range(0, TICK_DISPATCH_FREQUENCY_PERCENT):
+                    if tick < TICK_DISPATCH_FREQUENCY_PERCENT - 2:
```
fievelk, line 68: Nit: I'd add a comment to explain what's this `-2` and why it's needed.

gsy_e_sdk/setups/asset_api_scripts/redis_basic_strategies.py
```diff
@@ -0,0 +1,198 @@
...
+                        buy_rate = (fit_rate -
+                                    self.asset_strategy[area_uuid]["fee_to_market_maker"] +
+                                    (market_maker_rate +
+                                     2 * self.asset_strategy[area_uuid]["fee_to_market_maker"] -
+                                     fit_rate) * (tick / TICK_DISPATCH_FREQUENCY_PERCENT)
+                                    )
```
fievelk, line 74: Since this is a bit convoluted, I think this part deserves an explanation. It could be a comment, or you could maybe create a separate method with a docstring to explain what the load strategy does and why.

The same would apply to the generation and the storage strategies.

gsy_e_sdk/setups/asset_api_scripts/redis_basic_strategies.py
```diff
@@ -0,0 +1,198 @@
...
+                continue
+            self.asset_strategy[area_uuid] = {}
+            self.asset_strategy[area_uuid]["asset_name"] = area_dict["area_name"]
+            self.asset_strategy[area_uuid][
+                "fee_to_market_maker"
+            ] = self.calculate_grid_fee(
+                area_uuid,
+                self.get_uuid_from_area_name("Market Maker"),
+                "current_market_fee",
+            )
```
fievelk, line 62: Nit: I'd suggest moving the fee into a variable (since it's being accessed multiple times below) and maybe refactoring the dictionary instantiation:
```suggestion
            fee_to_market_maker = self.calculate_grid_fee(
                area_uuid,
                self.get_uuid_from_area_name("Market Maker"),
                "current_market_fee")
            
            self.asset_strategy[area_uuid] = {
                "asset_name": area_dict["area_name"],
                "fee_to_market_maker": fee_to_market_maker
            }
```

gsy_e_sdk/setups/asset_api_scripts/redis_basic_strategies.py
```diff
@@ -0,0 +1,198 @@
...
+                                 self.asset_strategy[area_uuid]["fee_to_market_maker"] -
+                                 (market_maker_rate +
+                                  self.asset_strategy[area_uuid]["fee_to_market_maker"] -
+                                  med_price) * (tick / TICK_DISPATCH_FREQUENCY_PERCENT)
+                                 )
+                    batt_sell_strategy.append(sell_rate)
+                self.asset_strategy[area_uuid]["buy_rates"] = batt_buy_strategy
+                self.asset_strategy[area_uuid]["sell_rates"] = batt_sell_strategy
+
+    def post_bid_offer(self, rate_index=0):
```
fievelk, line 124: Nit: we've decided to rename "bid/offer" into one name ("order"). So this might become shorter:
```suggestion
    def post_order(self, rate_index=0):
```

fievelk (approved):
Some comments, mostly for adding more documentation about the strategies.

PR #223, opened 2022-05-19
Degrees of freedom asset api refactoring

andreabertolini1995:

## Proposed changes

Proposed a scripts that shows how to post a bid with requirements (on…
…ly via Redis connection). A JSON file has been also added where the list of requirements for each asset can be listed.

### DIFF
--- a/gsy_e_sdk/setups/Asset_API_scripts/degrees_of_freedom.py
+++ b/gsy_e_sdk/setups/Asset_API_scripts/degrees_of_freedom.py
@@ -48,29 +48,27 @@ class Oracle(RedisAggregator):
     def post_bid(self):
         """Post a bid to the exchange."""
         for area_uuid, area_dict in self.latest_grid_tree_flat.items():
-            if "asset_info" not in area_dict or area_dict["asset_info"] is None:
+            asset_info = area_dict.get("asset_info")
+            if not asset_info:
                 continue
-            if (
-                "energy_requirement_kWh" in area_dict["asset_info"]
-                and area_dict["asset_info"]["energy_requirement_kWh"] > 0.0
-            ):
+            required_energy = asset_info.get("energy_requirement_kWh")
+            if required_energy:
                 rate = 10
-                energy = area_dict["asset_info"]["energy_requirement_kWh"]
                 if area_dict["area_name"] in self.degrees_of_freedom:
                     requirements = self.build_requirements_dict(
-                        rate, energy, area_dict["area_name"]
+                        rate, required_energy, area_dict["area_name"]
                     )
                     self.add_to_batch_commands.bid_energy_rate(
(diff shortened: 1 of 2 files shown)


hannesdiedrich (approved):
LGTM

gsy_e_sdk/setups/Asset_API_scripts/degrees_of_freedom.py
```diff
@@ -0,0 +1,138 @@
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.is_finished = False
+        self.degrees_of_freedom = {}
+
+    def on_market_cycle(self, market_info):
+        """Place a bid or an offer whenever a new market is created."""
+        if self.is_finished is True:
+            return
+        self.read_requirements()
```
fievelk, line 33: Nit: Right now, we're reading the requirements file on every market cycle event. I think this could be moved in the `__init__()` method so that we only read the DoF requirements once when the Oracle is instantiated.

gsy_e_sdk/setups/Asset_API_scripts/degrees_of_freedom.py
```diff
@@ -0,0 +1,138 @@
...
+
+
+aggregator = Oracle(aggregator_name=ORACLE_NAME)
+asset_args = {"autoregister": True, "pubsub_thread": aggregator.pubsub}
+
+
+def get_partner_ids(partners):
+    """Get the trading partners' UUID from the assets' names."""
+    asset_ids = {v: k for k, v in asset_uuid_mapping.items()}
+    return [asset_ids[p] for p in partners]
```
fievelk, line 113: This only works because `asset_uuid_mapping` is defined before we run this function (`get_partner_ids()`). However, `asset_uuid_mapping` is not defined inside of this function, so if we refactor and move stuff around, or if we change the variable name, this method will break.
In theory, `asset_uuid_mapping` should be added to the function parameters:
```suggestion
def get_partner_ids(partners, asset_uuid_mapping):
    """Get the trading partners' UUID from the assets' names."""
    asset_ids = {v: k for k, v in asset_uuid_mapping.items()}
    return [asset_ids[p] for p in partners]
```
However, this means that the `asset_uuid_mapping` should also be available inside of the `Oracle` instance (because it needs to be used by `build_requirements_dict()`). But the Oracle instance is used inside of `register_asset_list()` (which generates `asset_uuid_mapping`) so we're in a loop. This would require a bit more refactoring IMO, as it's probably not an immediate fix.

fievelk (commented):
One bigger comment. I would generally consider this as blocking, but - if you confirm this is working - I'll give my LGTM in case you want to merge.

fievelk (approved):

PR #224, opened 2022-05-19
Canary Network asset api refactoring

andreabertolini1995:

## Proposed changes

Proposed a script that would show how to send live data to a running Canary Network on the UI.


gsy_e_sdk/setups/Asset_API_scripts/canary_network_live_data.py
```diff
@@ -0,0 +1,143 @@
...
+        # pylint: disable=unused-variable
+        for area_uuid, area_dict in self.latest_grid_tree_flat.items():
+            asset_info = area_dict.get("asset_info")
+            if not asset_info:
+                continue
+
+            # Consumption assets
+            if "energy_requirement_kWh" in asset_info:
+                asset_name = area_dict["area_name"]
+                globals()[f"{asset_name}"].set_energy_forecast(
```
hannesdiedrich, line 54: Interesting! I am not sure whether this is best practice, but I see the need for it, so keep it as is.
(I ignore the case when `f"{asset_name}"` is not in the global variabel buffer, you probably want to raise an exception then, right?)

hannesdiedrich (commented):
Please also adapt for comments I had in the other PRs. In general, looks good to me.

gsy_e_sdk/setups/Asset_API_scripts/canary_network_live_data.py
andreabertolini1995: Could you maybe elaborate on how you would introduce this exception? Thanks a lot!

hannesdiedrich: Sorry, I formulated not clear. Let me try again:
If you have not added the `f"{asset_name}"` key value pair to the `globals()`, an exception will be raised. This is wanted behaviour, right?

hannesdiedrich (approved):
LGTM

gsy_e_sdk/setups/Asset_API_scripts/canary_network_live_data.py
```diff
@@ -0,0 +1,117 @@
...
```
fievelk, line 55: I'm getting back to this, as using globals is usually discouraged. I assume you want to access the `RestAssetClient` instance for this specific asset. In theory, we should find a way to pass the list of assets to the Oracle (so that it can use it in this method, without accessing `globals()`). One way could be to add it to the `__init__()` (as I mentioned in the other PR) and probably refactor the `register_asset_list()` method.

gsy_e_sdk/utils.py
```diff
@@ -337,3 +337,27 @@ def get_name_from_area_name_uuid_mapping(area_name_uuid_mapping, asset_uuid):
...
     return None
+
+
+def get_assets_name(registry: Dict) -> Dict:
+    """
+    Parse the grid tree and return all registered assets / markets
+    Wrapper for _get_assets_name
+    """
+    if registry == {}:
+        return {}
```
fievelk, line 348: Same as https://github.com/gridsingularity/gsy-e-sdk/pull/221/files#r899873985

gsy_e_sdk/setups/Asset_API_scripts/canary_network_live_data.py
```diff
@@ -0,0 +1,117 @@
...
+simulation_id = os.environ["API_CLIENT_SIMULATION_ID"]
+domain_name = os.environ["API_CLIENT_DOMAIN_NAME"]
+websockets_domain_name = os.environ["API_CLIENT_WEBSOCKET_DOMAIN_NAME"]
+asset_args = {}
+if CONNECT_TO_ALL_ASSETS:
+    registry = aggregator.get_configuration_registry()
+    registered_assets = get_assets_name(registry)
+    LOAD_NAMES = registered_assets["Load"]
+    PV_NAMES = registered_assets["PV"]
+    STORAGE_NAMES = registered_assets["Storage"]
```
fievelk, line 87: Same as https://github.com/gridsingularity/gsy-e-sdk/pull/221/files#r899879316

fievelk (approved):
Left some questions and comments. I'll give my LGTM in case you want to proceed faster, even though it would be good to address some comments if there's time :)

gsy_e_sdk/setups/Asset_API_scripts/canary_network_live_data.py
```diff
@@ -0,0 +1,117 @@
...
+from gsy_e_sdk.utils import get_area_uuid_from_area_name_and_collaboration_id, get_assets_name
+
+ORACLE_NAME = "oracle"
+
+# List of assets's names to be connected with the API
+LOAD_NAMES = ["Load 1", "Load 2"]
+PV_NAMES = ["PV 1", "PV 2"]
+STORAGE_NAMES = []
+
+CONNECT_TO_ALL_ASSETS = False
```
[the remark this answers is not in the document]
andreabertolini1995: Yes exactly. Also, the choice of setting CONNECT_TO_ALL_ASSETS is totally user-dependant, so I don't think there is a particular reason behind, why by default we set it as True or False.

fievelk (dismissed):

hannesdiedrich (approved):
LGTM

fievelk (approved):

PR #229, opened 2022-07-11
fix gsy-e-sdk setup scripts

maurerle:

Currently one gets "no module named d3a_api_client.setups.{setup_module_name}" as this is a reference to the old folder

## Proposed changes

-

### DIFF
--- a/gsy_e_sdk/cli.py
+++ b/gsy_e_sdk/cli.py
@@ -121,7 +121,7 @@ def load_client_script(base_setup_path, setup_module_name):
     """Load client script."""
     try:
         if base_setup_path is None:
-            importlib.import_module(f"d3a_api_client.setups.{setup_module_name}")
+            importlib.import_module(f"gsy_e_sdk.setups.{setup_module_name}")
         else:
             sys.path.append(base_setup_path)
             importlib.import_module(setup_module_name)


fievelk:
Hi @maurerle, thanks for your contribution! 
Could you please provide a reproducible example of this issue? What command are you using to get this error?

I'm asking because the old `d3a_api_client` package name is still kept as an "alias" for the new `gsy_e_sdk` and it should still work. For example, running the following command seems to works correctly, showing the expected warning:
```sh
$ gsy-e-sdk -l DEBUG run --setup test_redis_aggregator
WARNING: The d3a_api_client package name will be deprecated soon. Please use gsy_e_sdk instead.
11:06:29.205 INFO     ( 127) root                          : Trying to create aggregator test_aggr
11:06:29.218 INFO     (  76) root                          : Trying to register to load
...
```

maurerle:
Why should the above mentioned example show this warning? It is not using d3a_api_client anywhere?

I had the error after I removed the `d3a_api_client/__init__.py` it works normally, but why should one still keep this old relict? And for how long should it stay there?

fievelk:
> Why should the above mentioned example show this warning? It is not using d3a_api_client anywhere?

The code is using the old `d3a_api_client`, exactly on the line you removed. That's why this warning (not an error) is raised.

> I had the error after I removed the `d3a_api_client/__init__.py` it works normally, but why should one still keep this old relict? And for how long should it stay there?

The old `d3a_api_client` module is being kept for backwards compatibility. Existing users have implemented their code (i.e. setups) before this repo was renamed to `gsy_e_sdk`. Therefore, if we completely remove support for `d3a_api_client` their setups code will break and this is not advisable. That's why we keep the old module as a simple alias to the new one and show a warning. In the future, we might remove this backward compatibility and completely deprecate the old name, but we don't have a date for it yet.

PR #233, opened 2022-09-12
PH-1131: Fix rest_basic_strategies for API Storage Strategies

ERafaelMartinez:

Fix issues on API template strategies for Storage assets. Previous strategy was not allowing proper rate updates for storages.

## Proposed changes

**INTEGRATION_TESTS_BRANCH**=PH-1131

### DIFF
--- a/gsy_e_sdk/setups/asset_api_scripts/rest_basic_strategies.py
+++ b/gsy_e_sdk/setups/asset_api_scripts/rest_basic_strategies.py
@@ -14,8 +14,8 @@ from gsy_e_sdk.utils import get_area_uuid_from_area_name_and_collaboration_id
 ORACLE_NAME = "oracle"
 
 # List of assets's names to be connected with the API
-LOAD_NAMES = ["Load 1 L13", "Load 2 L21", "Load 3 L17"]
-PV_NAMES = ["PV 1 (4kW)", "PV 3 (5kW)"]
+LOAD_NAMES = ["Load L13", "Load 2 L21", "Load 3 L17"]
+PV_NAMES = ["PV 2 (1kW)", "PV 3 (5kW)"]
 STORAGE_NAMES = ["Tesla Powerwall 3"]
 
 # Frequency of bids/offers posting in a market slot - to leave as it is
@@ -65,7 +65,7 @@ class Oracle(Aggregator):
                 "fee_to_market_maker"
             ] = self.calculate_grid_fee(
                 area_uuid,
-                self.get_uuid_from_area_name("Market Maker"),
+                self.get_uuid_from_area_name("Grid Market"),
                 "current_market_fee",
             )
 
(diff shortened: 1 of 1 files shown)


gsy_e_sdk/setups/asset_api_scripts/rest_basic_strategies.py
```diff
@@ -153,16 +153,20 @@ def post_bid_offer(self, rate_index=0):
...
 
             # Storage assets
-            buy_energy = asset_info.get("energy_to_buy")
-            if buy_energy:
+            buy_energy = asset_info.get("energy_to_buy", 0)
+            bid_energy = asset_info.get("energy_active_in_bids", 0)
+            if buy_energy > 0 or bid_energy > 0:
                 buy_rate = self.asset_strategy[area_uuid]["buy_rates"][rate_index]
-                self.add_to_batch_commands.bid_energy_rate(
+                buy_energy = buy_energy + bid_energy
```
hannesdiedrich, line 160: So you are offering the energy, that you placed bids for already?

hannesdiedrich (commented):
One clarification. Otherwise LGTM

gsy_e_sdk/setups/asset_api_scripts/rest_basic_strategies.py
ERafaelMartinez: Thanks for your review!
To answer your question:
Yes, I'm accounting for the `energy_to_buy` **and** the `energy_active_in_bids` as the total energy requirements.

Then, a new bid will be made considering the whole amount of energy that is required, replacing the old bid.
This allows posting a new (updated) bid that overrides the old one but with an updated rate.

I don't understand why this would be dangerous. Maybe the name I gave the `bid_energy`variable is misleading, as it refers to the `energy_active_in_bids` value received in the area_dict on the corresponding tick. Maybe I got you wrong. Could you please elaborate on that case?

hannesdiedrich: Yes, I got all that and I see my confusion now. Have a look one line below. You want to buy energy here, but call `self.add_to_batch_commands.offer_energy_rate()` . I think it should be `self.add_to_batch_commands.bid_energy_rate()`. 
Thanks!

spyrostz: Agreed with Hannes' comment and also want to suggest to rename the variable `bid_energy` to `open_bids_energy` in order for the intent to be immediately clear. Thanks!

ERafaelMartinez: @hannesdiedrich You're right! Thanks for pointing that out.

I addressed both observations on the latest commit.

ERafaelMartinez:
Thank you for your reviews! I addressed your comments on the latest commit.

hannesdiedrich (approved):
LGTM

PR #244, opened 2024-10-25
GSYE-803: Add set_live_generation to RestAssetClient

hannesdiedrich:

Please describe what we want to achieve and why.

## Proposed changes

-

### DIFF
--- a/gsy_e_sdk/clients/rest_asset_client.py
+++ b/gsy_e_sdk/clients/rest_asset_client.py
@@ -121,6 +121,16 @@ class RestAssetClient(APIClientInterface, RestCommunicationMixin):
 
         return None
 
+    @logging_decorator("set-live-generation")
+    def set_live_generation(self, live_data: Dict, do_not_wait=False):
+        """Send live generation data to gsy-web."""
+        transaction_id, posted = self._post_request(f"{self.endpoint_prefix}/set-live-generation",
+                                                    {"live_data": live_data})
+        if posted and do_not_wait is False:
+            return self.dispatcher.wait_for_command_response("set-live-generation", transaction_id)
+
+        return None
+
     # pylint: disable=invalid-name
     @logging_decorator("set-energy-measurement")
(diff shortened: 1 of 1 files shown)


spyrostz (approved):
LGTM

gsy_e_sdk/clients/rest_asset_client.py
```diff
@@ -121,6 +121,16 @@ def set_energy_forecast(self, energy_forecast_kWh: Dict, do_not_wait=False):
 
         return None
 
+    @logging_decorator("set-live-generation")
+    def set_live_generation(self, live_data: Dict, do_not_wait=False):
+        """Send live generation data to gsy-web."""
+        transaction_id, posted = self._post_request(f"{self.endpoint_prefix}/set-live-generation",
+                                                    {"live_data": live_data})
```
BigTava, line 128: Doesn't the endpoint only accept config_uuid and area_uuid as per 
set_live_generation(request, config_uuid, area_uuid):
What is the relationship between live_data and these?

hannesdiedrich: I did not study informatics, so my following explanation might not use the correct nomenclature, but let me try to explain anyway:
There are two streams for data in a REST API: The parameters and the body. The parameters are put in the URL of the request and any additional data can be put in the body. 
So for this specific API endpoint a valid URL looks like:
```
http://localhost:8000/external-connection/api/cc3251d1-320e-4158-818f-e2e49b2bfc04/d1013b2f-bc07-403b-b61e-5f593cb0e749/set-live-generation/
``` 
That contains the config_uuid and the area_uuid. The actual data is sent in the body, that is represented by the `request.data` in the endpoint implementation here:  
https://github.com/gridsingularity/gsy-web/pull/2198/files#diff-50bbb55953ee93d276958a5a67f62263471aba048061dfedd78ef488f513988aR92

The additional data is a dictionary that looks like:
```
{"live_data":"20240101T00:00"}
```

PR #245, opened 2024-11-25
GSYE-822: Removed SCM group settings from the GraphQL query, since it…

spyrostz:
… is unsupported.

Please describe what we want to achieve and why.

## Proposed changes

-


hannesdiedrich (approved):
LGTM

